Issue #115, opened 2015-12-07
Package slurm-torque-ohpc depends on slurm-perlapi, not slurm-perlapi-ohpc


ajdecon:
Package slurm-torque-ohpc in the OpenHPC_1.0 repo cannot be installed due to an unsatisfied dependency on slurm-perlapi. Note that the OpenHPC_1.0 repo does contain a package "slurm-perlapi-ohpc", which satisfies the dependencies on perl(Slurm) and perl(Slurmdb) according to "yum deplist"; but the package appears to depend explicitly on slurm-perlapi without the ohpc name, which is not available.

Output of "yum deplist slurm-torque-ohpc" follows:

```
$ sudo yum deplist slurm-torque-ohpc
Loaded plugins: fastestmirror
Loading mirror speeds from cached hostfile
 * base: centos-mirror.jchost.net
 * epel: lug.mtu.edu
 * extras: mirrors.adams.net
 * updates: repos.redrockhost.com
package: slurm-torque-ohpc.x86_64 15.08.1-4.1
  dependency: /usr/bin/perl
   provider: perl.x86_64 4:5.16.3-285.el7
  dependency: libc.so.6(GLIBC_2.14)(64bit)
   provider: glibc.x86_64 2.17-78.el7
  dependency: libpthread.so.0()(64bit)
   provider: glibc.x86_64 2.17-78.el7
  dependency: libpthread.so.0(GLIBC_2.2.5)(64bit)
   provider: glibc.x86_64 2.17-78.el7
  dependency: perl(English)
   provider: perl.x86_64 4:5.16.3-285.el7
  dependency: perl(File::Temp)
   provider: perl-File-Temp.noarch 0.23.01-3.el7
  dependency: perl(FindBin)
   provider: perl.x86_64 4:5.16.3-285.el7
  dependency: perl(Getopt::Long) >= 2.24
   provider: perl-Getopt-Long.noarch 2.40-2.el7
  dependency: perl(POSIX)
   provider: perl.x86_64 4:5.16.3-285.el7
  dependency: perl(Slurm)
   provider: slurm-perlapi-ohpc.x86_64 15.08.1-4.1
  dependency: perl(Slurmdb)
   provider: slurm-perlapi-ohpc.x86_64 15.08.1-4.1
  dependency: perl(Switch)
   provider: perl-Switch.noarch 2.16-7.el7
  dependency: perl(autouse)
   provider: perl.x86_64 4:5.16.3-285.el7
[...]

PR #116, opened 2015-12-07
Fix for issue #115

ajdecon:
Fix for #115. This change adds the PROJ_DELIM tag to the slurm-perlapi dependency of slurm-torque-hpc, as specified in other dependencies.

Note that the slurm specfile seems a little inconsistent in where it uses "slurm" vs "%{pname}", and in where the "%{PROJ_DELIM}" tag is included. For example, the bluegene and aix packages are dependent on "slurm" instead of "%{pname}%{PROJ_DELIM}" so I suspect they won't build. However, lacking a convenient bluegene to test on, I thought I'd restrict this fix to the issue I actually encountered. ;)


koomie:
Just cleaning up - Ticket #115 was closed, but the companion PR was not. Note that this was resolved back with the v1.0.1 release via dbc26fd94e30c1c74830dc9c89d6bff128ea8024.

PR #178, opened 2016-02-16
Cube

surak:
[Forschungszentrum Jülich](http://www.fz-juelich.de) would like to provide our performance tools to OpenHPC: [Score-P](http://www.score-p.org/), [Scalasca](http://www.scalasca.org) and [Cube](http://www.scalasca.org/software/cube-4.x/download.html). I would be the maintainer of them here. 

This is a copy of the spec files for TAU, adapted for Cube.


koomie:
Thanks @surak. Landed this PR on a new issue_178 branch.

surak:
Thanks. I'm still figuring out things. I committed this: https://github.com/surak/ohpc/commit/851758357420dbdfcc850b9dcdbc42ed3b9afa48
should it appear here?

koomie:
FYI, the OBS build is triggering off the issue_178 branch so I landed that surak@8517583 patch on the branch, and the rebuild kicked off.

koomie:
Just a quick note to say that a basic install of the current build on centos7.1 went ok for me locally. Once installed, a "module load cube" allowed me to fire up the cube GUI when running X. Now, a couple of quick questions:

* are there some examples pre-installed that can be used with the GUI?
* I didn't see man pages installed; should there be some? If not, I would propose we remove the MANPATH setting in the module

Also, I removed the CUBE_MAKEFILE setting which I assume was carried over using tau as a template.  Finally, I added CLASSPATH which I saw in the example module file sitting under share/.

registry:
No manpages  - correct. 
Example cube file will be added in next version (noted in tickets). 
Currently we install couple of sources as examples of cube library usage. If one compiles them and runs -> they produce a cube file, which can be open with Cube GUI

koomie:
Landed 	https://github.com/openhpc/ohpc/pull/178/commits/17000841623b012102f4de2f6f54bfdbbeaa1604 on the issue_178 branch.

surak:
What about 6e1fdc2?

koomie:
Just doing a little house cleaning. This was superseded with the work that came after the component  submission review and #319.

PR #181, opened 2016-02-23
Cube and scorep

surak:
(I'm not sure if I am doing this correct, or if it would trigger automatically, it's all confusing here)

Cube is going to be separated in multiple builds:

- Cube Library
- Cube Writer
- Cube GUI
- Java reader

Besides, I added a prototype for Score-P.

Those files were added to my branch/fork/whatever (I guess). It does not show on the issue_178, nor other changes I have done recently.

### DIFF
--- a/README
+++ b/README
@@ -1,0 +1,2 @@
+OpenHPC: Community building blocks for HPC systems.
+
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ section of the latest
 ###### [CentOS 7.1]
 * [ohpc-release.x86_64.rpm] (https://github.com/openhpc/ohpc/releases/download/v1.0.1.GA/ohpc-release-1.0-1.x86_64.rpm) (md5sum=8dbaed1bea6245c04cf201131df74bc3)
 * [Install Guide](https://github.com/openhpc/ohpc/releases/download/v1.0.1.GA/Install_guide-CentOS7.1-1.0.1.pdf)
+* [ISO image](http://build.openhpc.community/OpenHPC:/1.0/CentOS_7.1/iso/OpenHPC-1.0-CentOS_7.1.iso) mirror of Yum repository (md5sum=e57db7c5dd712c7b9fa48e4ebd43c517)
 
 #### Questions, Comments, or Bug Reports?
 
--- a/components/perf-tools/cube/SOURCES/rpmlintrc
+++ b/components/perf-tools/cube/SOURCES/rpmlintrc
@@ -1,0 +1,3 @@
+addFilter("devel-file-in-non-devel-package")
+addFilter("suse-filelist-forbidden-backup-file")
+addFilter("buildroot")
--- a/components/perf-tools/cube/SPECS/cube.spec
+++ b/components/perf-tools/cube/SPECS/cube.spec
@@ -17,7 +17,7 @@
 %define PNAME %(echo %{pname} | tr [a-z] [A-Z])
 
 
-Name: %{pname}-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
+Name: %{pname}%{PROJ_DELIM}
 
 Version:   4.3.3
 Release:   1%{?dist}
@@ -98,21 +98,20 @@ popd
 
 
 # OpenHPC module file
-%{__mkdir} -p %{buildroot}%{OHPC_MODULEDEPS}/%{compiler_family}-%{mpi_family}/%{pname}
-%{__cat} << EOF > %{buildroot}/%{OHPC_MODULEDEPS}/%{compiler_family}-%{mpi_family}/%{pname}/%{version}
+%{__mkdir_p} %{buildroot}/%{OHPC_MODULES}/%{pname}
+%{__cat} << EOF > %{buildroot}/%{OHPC_MODULES}/%{pname}/%{version}
 #%Module1.0#####################################################################
 
 proc ModulesHelp { } {
 
     puts stderr " "
-    puts stderr "This module loads the %{pname} library built with the %{compiler_family} compiler"
-    puts stderr "toolchain and the %{mpi_family} MPI stack."
+    puts stderr "This module loads the %{pname} tool"
     puts stderr "\nVersion %{version}\n"
 
 }
-module-whatis "Name: %{pname} built with %{compiler_family} compiler"
+module-whatis "Name: %{pname}"
 module-whatis "Version: %{version}"
-module-whatis "Category: runtime library"
+module-whatis "Category: performance tool"
 module-whatis "Description: %{summary}"
 module-whatis "URL %{url}"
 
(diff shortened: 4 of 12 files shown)


JohnWestlund:
Hi @surak, I'm not up to date on the context of this request but here are some things I notice.

First the reason this isn't showing up as an update to #178 is because you're trying to merge into a different branch: OHPC 1.0.1 vs issue_178. You wouldn't want to merge into 1.0.1 anyway since that's for minor updates to 1.0.1 -- no package additions, just security updates, etc.

If OBS is pointed at the OHPC issue_178 branch to perform builds, it will only kick off once it's been merged into OHPC's repo. I'll let @koomie decide if this PR is ready to get merged in.

I would close PR #178, and use this PR instead.

koomie:
Hi @surak, thanks for this PR. I see that you did create it against the issue_178 branch which was very useful so that I could accept it using the Github UI directly. I'd suggest continuing to use that approach while we iterate on the additions.

A couple of quick comments; I did a little bit of reorganization to move each of the builds to a separate directory in perf-tools/. Had you been using the same input tarball for all of the builds, I would have suggested a single .spec file that generates multiple packages, but in this case, going with separate subdirectories seems to make more sense since they are driven by alternate inputs (and it melds with the OBS build system more directly).

With that said, I enabled builds in OBS to try and match your split configuration. Most of the builds failed initially as it could not download a number of the input tarballs with the url as specified. When I checked the web page for the download url, I noticed a "TP1" designation for a number of the source tarballs, so I tweaked several of those.  Not sure if that's what you had in mind, so definitely worth a look.  Of the new builds, I believe "cubew" is the only one building so far. There was an empty "Requires" line and duplicate "BuildRequires" that I had to change, and a sed line that was erroring off that I commented out for the time being.  Most of the other builds are not starting due to cubelib not being available. It looks like it has a number of dependencies that are not resolved by the distro (e.g. Rcpp, Rscript, etc).  How would you normally resolve those?

surak:
Hello!
So, should I keep using issue_178 branch for everything Forschungszentrum Jülich does?

TP1 comes from "Technology preview". The idea is that those are early alphas of what will become Cube. The codebase is the same, but the components were split in their unique functionalities.

I have a question, though. I started to install OHPC here, in order to reproduce the build system, and, with that, be able to test things properly instead of asking for pull requests with blind eyes. However, I have found no documentation on it. Would you care to point to where can I find the steps to reproduce the environment of the OBS in our own machines (or at least a VM) so I can submit clean, working spec files?

koomie:
Yes, continuing to use the issue_178 branch as the source of PRs for this effort on cube/scorep/scalasca seems reasonable to me.  

Regarding the builds, one thing I might suggest before endeavoring to reproduce the OBS build system (which is probably overkill in this case), is to do some rpmbuilds locally.  At the end of the day, OBS just calls `rpmbuild`, so I often do this by hand locally to iron out some of the initial kinks.  As an example, here are some steps I did to attempt to build scorep with the latest .spec file:

```
# go to latest scorep config in issue_178 branch
$ cd components/perf-tools/scorep/SPECS/

# initialize rpmbuild environment locally
$ mkdir -p ~/rpmbuild/{BUILD,RPMS,SOURCES,SPECS,SRPMS}

# include OHPC_macros in your rpm build sandbox
$ cp ../../../OHPC_macros ~/rpmbuild/SOURCES/

# pull down the desired input tarball (matching url in .spec file)
$ wget -P ~/rpmbuild/SOURCES/ http://www.vi-hps.org/upload/packages/scorep/scorep-1.4.2.tar.gz

# add any necessary BuildRequires locally....

# build locally
$ rpmbuild -bb scorep.spec 
Executing(%prep): /bin/sh -e /var/tmp/rpm-tmp.jozMRv
+ umask 022
+ cd /home_nfs/kwschulz/rpmbuild/BUILD
+ cd /home_nfs/kwschulz/rpmbuild/BUILD
+ rm -rf scorep-1.4.2
+ /usr/bin/gzip -dc /home_nfs/kwschulz/rpmbuild/SOURCES/scorep-1.4.2.tar.gz
+ /usr/bin/tar -xf -
+ STATUS=0
+ '[' 0 -ne 0 ']'
+ cd scorep-1.4.2
+ /usr/bin/chmod -Rf a+rX,u+w,g-w,o-w .
+ sed -i -e 's/^BITS.*/BITS = 64/' src/Profile/Makefile.skel
sed: can't read src/Profile/Makefile.skel: No such file or directory
error: Bad exit status from /var/tmp/rpm-tmp.jozMRv (%prep)


RPM build errors:
    Bad exit status from /var/tmp/rpm-tmp.jozMRv (%prep)
```

Note that the sed error above is the same as we see with the latest OBS build attempt at:

https://build.openhpc.community/package/live_build_log/OpenHPC:Devel/scorep/CentOS_7.2/x86_64

So, you could install CentOS7.2 on a machine (or VM), enable the OpenHPC 1.0.1 repo so you can install any necessary sub-dependencies and do the builds similarly to OBS without having to jump thru the hoops of a full OBS install/config.

registry:
Hi
About dependencies of CubeLib: One of the Cube's tools (cube_dump) is capable to export the profiling data 
into R matrices, if R is installed. But only R is not enough. R itself has to have installed modules Rscript, RInside and Rcpp.
(http://mazamascience.com/WorkingWithData/?p=728, 
https://stat.ethz.ch/R-manual/R-devel/library/utils/html/Rscript.html
http://dirk.eddelbuettel.com/code/rinside.html
and so on)
These are installed usually within R console. I don't know how one formulates such dependency in .spec file.
While building Cube out of tarball configure script looks up for these features and enables if found. 
Otherwise cube_dump just doesn't export data into R , but builds and works otherwise. (Preprocessor guards)
If zlib is found -> Cube supports zlib compression.

Dependenty on CubeW:  zlib-devel . If not found - it doesn't support compression, but builds anyway.

Dependency in jcuber: it is native java library and builds a CubeReader.jar file using javac and jar tools.
I don't know which rpm provides them to setup Requires properly.

Dependency on CubeGUI: it depends on dbus-devel (if not found "vampir" plugin is not build but gui builds anyway.. .
so... weak dependency), qt (>4.6.2) - strong dependenty (is a gui) and cubelib.

registry:
Do "cubelib, jcuber and cubegui " build if one skips "Required" entry  (leave empty) and lets "configure" 
finds things ?

koomie:
I commented out the R requirements (and a failing sed command) and then cubelib was able to build:

https://build.openhpc.community/package/live_build_log/OpenHPC:Devel/cubelib/CentOS_7.2/x86_64

registry:
ok (about R and building cubelib)

What is about javac/jar ?  Won't be supported? Or jcuber is out of scope of OpenHPC? 
To clarify, TAU is the user of jcuber. Before they downloaded our tarball form the scalasca.org web page, built it on their cluster and resulting .jar file included into TAU package.  (because they built it with a relatively old Java compiler to generate most compatible jvm code)

Theoretically their .spec file should depend on jcuber. 

And how one can specify dependency between such .spec file on build server like  cubegui <- cubelib ?

PR #214, opened 2016-04-12
add a build dep on librdmacm-devel to enable RDMA CM support

LaHaine:
I've noticed that the ohpc builds of mvapich2 don't support RDMA CM. It is possible to enable it by adding a build dep on librdmacm-devel.


koomie:
Thanks for the addition and doing the PR against 1.1_Factory. I landed it on 1.1 and the rebuilds are kicking off now. Assuming there are issues arising from tests running in CI, we'll squeeze this in for 1.1.

PR #235, opened 2016-06-01
Fix for Issue # 213

nirmalasrjn:
As suggested by Scott Suchyta, appending memlock settings before # End of file


koomie:
Landed patch onto 1.1.1. branch.  Thanks.

PR #236, opened 2016-06-10
Enable NFS at boot time

nirmalasrjn:


koomie:
Landed onto 1.1.1 branch.  Thanks @nirmalasrjn .

PR #240, opened 2016-06-22
Resolves Issue#239

nirmalasrjn:
To remove hard-coding changed compute_regex variable to point to just
the node prefix.Made appropriate changes to wwsh provision set commands and
stateful provisioning commands where the compute_regex variable has been referenced.

### DIFF
--- a/docs/recipes/install/centos7.2/input.local.template
+++ b/docs/recipes/install/centos7.2/input.local.template
@@ -56,7 +56,8 @@ enable_intel_packages="${enable_intel_packages:-0}"
 num_computes="${num_computes:-4}"
 
 # regex that matches defined compute hostnames
-compute_regex="${compute_regex:-c}"
+compute_regex="${compute_regex:-c*}"
+compute_prefix="${compute_prefix:-c}"
 
 # compute hostnames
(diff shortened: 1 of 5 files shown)


koomie:
Looking at this, I see now why you altered the use of `${compute_regex}` in PR #253.  However, I think we need to be careful here. Your redefinition in this PR means that the variable is no longer defining a legitimate regex. At a minimum, the variable would likely need to be renamed to try and minimize confusion, but there are many times when having a regex that covers all compute names is very useful, so I would be in favor of not going this route.  

Just thinking out loud here, perhaps an alternative solution would be to introduce another input variable that encapsulates the compute node delimiter.  Say something like:


`compute_delim="${compute_delim:-c}"`

or 

`compute_prefix=${compute_prefix:-c}"`

and leave compute_regex as per its original intent.  What do you think about this approach?

nirmalasrjn:
I agree that it is good to have a regex that covers all compute nodes. I will go with your suggestion to add a new variable. I will do that and resubmit the pull request. Thank you

koomie:
Thanks for updating. Landed on 1.1.2 branch.

PR #242, opened 2016-06-24
Update intel-compilers-devel.spec

oere:
Typo in summary of the package.


koomie:
Thanks for the catch. Landed on 1.1.2 branch.

PR #244, opened 2016-06-24
Adds head node as nameserver of compute nodes

nirmalasrjn:
Issue #243


koomie:
Landed onto 1.1.2 branch.  Thanks.

koomie:
I think we need to think about this one some more and will push to the 1.3 milestone. My concern is that the administrator may have already put reasonable settings for DNS in `/etc/resolv.conf`. And, they may prefer to resolve via something other than ${sms_name} which was the motivation for suggesting to copy from `/etc/resolv.conf` in the first place. If we go this route, should probably consider defining a variable for dns resolution and search.

crbaird:
Hi @nirmalasrjn. This PR has stalled a bit. Would you have interest in updating it and adding some DNS variables to inputs.tex and re-basing to the 1.3.5 branch, or can it just be closed?

Thanks

adrianreber:
Closing as the 1.3.x branch has not seen any activity for a long time.

PR #253, opened 2016-06-28
Provides option for stateful provisioning

nirmalasrjn:
Resolves Issue#252.

### DIFF
--- a/docs/recipes/install/common/stateful.tex
+++ b/docs/recipes/install/common/stateful.tex
@@ -21,13 +21,13 @@ which partition to format, and what the filesystem layout will look like.
 [sms](*\#*) export sda1="mountpoint=/boot:dev=sda1:type=ext3:size=500"
 [sms](*\#*) export sda2="dev=sda2:type=swap:size=32768"
 [sms](*\#*) export sda3="mountpoint=/:dev=sda3:type=ext3:size=fill"
-[sms](*\#*) wwsh -y object modify -s bootloader=sda -t node "${compute_regex}[1-${num_computes}]" 
-[sms](*\#*) wwsh -y object modify -s diskpartition=sda -t node "${compute_regex}[1-${num_computes}]" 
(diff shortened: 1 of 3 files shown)


docs/recipes/install/common/stateful.tex
```diff
@@ -2,42 +2,40 @@
...
-\begin{lstlisting}[language=bash,literate={-}{-}1,keywords={},upquote=true]
 # Update node object parameters
 [sms](*\#*) export sda1="mountpoint=/boot:dev=sda1:type=ext3:size=500"
 [sms](*\#*) export sda2="dev=sda2:type=swap:size=32768"
 [sms](*\#*) export sda3="mountpoint=/:dev=sda3:type=ext3:size=fill"
-[sms](*\#*) wwsh -y object modify -s bootloader=sda -t node "${compute_regex}" 
-[sms](*\#*) wwsh -y object modify -s diskpartition=sda -t node "${compute_regex}" 
-[sms](*\#*) wwsh -y object modify -s diskformat=sda1,sda2,sda3 -t node "${compute_regex}" 
-[sms](*\#*) wwsh -y object modify -s filesystems="$sda1,$sda2,$sda3" -t node "${compute_regex}" 
+[sms](*\#*) wwsh -y object modify -s bootloader=sda -t node "${compute_regex}[1-${num_computes}]"
```
koomie: I think the intent was for ${compute_regex} to be a regex that maps to all computes. So, do we really need/want to replace ${compute_regex} with ${compute_regex}[1-${num_computes}] here?

koomie:
Thanks for updating this PR. Landed on the 1.1.2 branch.

koomie:
I'm going to re-register this for the 1.3 milestone so can iterate a bit. I like the intent of this change, but might suggest that we not just include the disk settings verbatim from the docs as they are today.  For the script, it should probably introduce some options for the choice of disk, some basic partitioning sizes, etc.....

koomie:
No progress here, closing this out.

PR #255, opened 2016-07-07
Fix SUSE name

dirkmueller:
SuSE was renamed to SUSE a little bit more than 12 years ago,
its time to adjust.


koomie:
Thank you - sorry for propagating the wrong convention.

Issue #239, opened 2016-06-22
Remove hard-coding of compute hostnames


nirmalasrjn:
Remove hard-coding of compute host names in slurm configuration file and in /etc/clustershell/groups.d/local.cfg.

JohnWestlund:
@koomie @crbaird Deepika has some patches that might help address this

koomie:
Passing in CI.

PR #258, opened 2016-08-03
Issue#257

nirmalasrjn:
Install guide to Include option to enable OpenHPC repository using ISO image and  an input variable that encapsulates compute node delimiter.


nirmalasrjn:
The Issue #239 has taken care of introducing the input variable for compute node delimiter.  Since a  simple utility has been introduced via Issue# 363 to enable OpenHPC repository from downloaded tarball, the part about, enabling OpenHPC repo, using ISO image, has also been taken care of. So this pull request is no longer relevant and is being closed now.

PR #263, opened 2016-08-26
aarch64: first batch of fixes

ericvh:
With these patches, aarch64 on CentOS_7.2 builds 128 successful packages.  They are predominantly configuration fixes with a single patch against openblas to fix some bad code in the arm64 support.  I regressed them against x86_64 CentOS 7.2 and SuSe 12 SP 1 with no adverse effects on package builds.

Build test results against aarch64 available in my home project: https://build.openhpc.community/project/show/home:eric:OpenHPC:1.2:Factory

This is my first time really doing any sort of packaging work with Linux, I've mostly just stuck to code before, so be gentle.  Happy to take suggestions on ways of doing things better.

Signed-off-by: Eric Van Hensbergen <eric.vanhensbergen@arm.com>

### DIFF
--- a/components/admin/lmod/SPECS/lmod.spec
+++ b/components/admin/lmod/SPECS/lmod.spec
@@ -95,6 +95,8 @@ if [ ! -z "\$SLURM_NODELIST" ];then
      return
 fi
 
+export LUA_CPATH="%{LUA_CPATH}"
+export LUA_PATH="%{LUA_PATH}"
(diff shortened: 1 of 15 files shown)


koomie:
Just a head's up that I landed all of these except for c1c0c34. I'm wondering if we can sort out another option for the lua path, but probably need to get interactive access to an aarch64 environment.

Thanks for the updates.

PR #269, opened 2016-08-31
Subnet Manager for InfiniBand Network

nirmalasrjn:


koomie:
Going to schedule this for 1.2, but think we need to tweak per comment on the patch.

crbaird:
bumping smart opensm support to update1

koomie:
Hey @nirmalasrjn, we squeezed a tooltip in referencing opensm and a small bit in the recipe script to optionally install;l the extra bit is simply as follows:

````
# Optionally enable opensm subnet manager                                                                                                  
if [[ ${enable_opensm} -eq 1 ]];then
     zypper -n install opensm
     systemctl enable opensm
     systemctl start opensm
fi
```

I added a default for `enable_opensm` into the template input files.  Hopefully this a reasonable compromise for what you had in mind.

PR #273, opened 2016-09-06
Obs/open hpc 1.1.2 factory

e4t:
These fixes address the following issues:
1. munge: */lib/tmpfiles.d/munge.conf neither gets built nor installed
   on SUSE.
2. Macro definitions like %{!?foo: %define foo bar} should be avoided
   due to the scoping rules for rpm macros. a '%global' should be used
   instead of the '%define'. This affects %PROJ_DELIM, %compiler_family
   and %mpi_family
3. A macro %{RPM_BUILD_ROOT} does not exist. Either it is the shell variable
   ${RPM_BUILD_ROOT} or the macro %{buildroot}: this affected the packages 
   'R', 'spack' and 'gsl'.
4. An rpmlint check in 'cube' for running ldconfig in the %post and %postun
   scripts thru errors and raised the error level above the threshold.
   Since this package uses lmod to set up the environment, ldconfig should
   not be needed.


koomie:
closed branch.

PR #315, opened 2016-10-13
easybuild: update patch after version bump

pawosm-arm:
Currently used easybuild fix fails silently to apply non-x86 patch causing false positive. This fix is less version dependent.


boegel:
fwiw: this patch will be irrelevant/broken for the upcoming EasyBuild v3.0.0 (ETA Nov'16), thanks to backporting of the patch into upstream @ https://github.com/hpcugent/easybuild-easyblocks/pull/1003

PR #316, opened 2016-10-14
warewulf: copy config.sub too

pawosm-arm:


jmstover:
I actually don't see an issue in it. @pawosm-arm , what was the reasoning behind copying the config.guess to the build directory? It... shouldn't need it, but if there's some valid reason I can add it upstream so it's there.

pawosm-arm:
This is possible fix to address build failure on SuSe/AArch64, from the log you can find that config.sub as shipped with warewulf don't quite understand what -suse- means.

jmstover:
Ahh... Yeah, we don't move out much beyond the RHEL based world. ;)

On an SL 6.5 provisioner I have here, the autoconf and RPM config.guess are the same, so there wouldn't be an issue that I can foresee also adding this in upstream as well so the patch isn't needed constantly.

pawosm-arm:
Actually this patch helps with building on SuSe:
https://build.openhpc.community/package/show/home:eric:OpenHPC:1.2:Factory/warewulf-ipmi

PR #336, opened 2016-11-10
Tests: OpenBLAS: AArch64 tweaks

pawosm-arm:


pawosm-arm:
These tests cannot be linked due to missing symbols in OpenBLAS library for AArch64. These symbols are extensions to BLAS and are provided as x86_64 and MIPS64 assembly and not available for AArch64. The tests should not make calls to them on AArch64.

PR #342, opened 2016-11-18
impi: Adjust version parsing to deal with spaces

chuckatkins:
This makes the version parsing replace " Update " with "." to prevent
spaces in generated module file names, which breaks other module
infrastructure.  This is specifically to address a problem with the impi
shipped with Parallel Studio Cluster Edition 2017 Update 1, in which
the version retrieved for automatic module file creation  was
"2017 Update 1". This now becomes "2017.1" thus resulting in a more
sensible module file name that doesn't break the remaining module
infrastructure..

### DIFF
--- a/components/mpi-families/impi-devel/SPECS/intel-mpi.spec
+++ b/components/mpi-families/impi-devel/SPECS/intel-mpi.spec
@@ -64,7 +64,7 @@ if [ -d ${topDir} ];then
 
     for dir in ${versions}; do
 	if [ -e ${topDir}/${dir}/linux/mpi/intel64/bin/mpiicc ];then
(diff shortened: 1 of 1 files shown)


chuckatkins:
Fixes #343

chuckatkins:
Fixes #343

chuckatkins:
If I had known it was to get pulled into the release branch I would have based it much earlier in the graph.  It looks like the merged into obs/OpenHPC_1.2.1_Factory pulled in a bunch of unrelated stuff.  Would it be helpful if I rebased the branch back to the v1.2.GA tag instead?  the commit is trivial so there shouldn't be any conflicts.  That will let it cleanly merge into both master and release.

Or, since you already did the merge should I just leave it alone instead?

koomie:
It's ok. I probably should have cherry-picked it to avoid pulling in unrelated stuff from master branch but it's landed on 1.2.1 now.

Issue #254, opened 2016-07-04
Intel module file is not created by intel-compilers-devel-ohpc


LaHaine:
The package intel-compilers-devel-ohpc-2016-19.1.x86_64 tries to generate a module file for the Intel compiler but fails to do that. This happens when I install the package on a fresh EL7.2 system:

```
Resolving Dependencies
--> Running transaction check
---> Package intel-compilers-devel-ohpc.x86_64 0:2016-19.1 will be installed
--> Processing Dependency: intel-compxe >= 2016 for package: intel-compilers-devel-ohpc-2016-19.1.x86_64
--> Running transaction check
---> Package intel-compxe.noarch 0:2016.3-067 will be installed
--> Finished Dependency Resolution

Dependencies Resolved

================================================================================
 Package                        Arch       Version        Repository       Size
================================================================================
Installing:
 intel-compilers-devel-ohpc     x86_64     2016-19.1      ZN_7.2_arch     7.8 k
Installing for dependencies:
 intel-compxe                   noarch     2016.3-067     ZN_7_arch       4.2 k

Transaction Summary
================================================================================
Install  1 Package (+1 Dependent package)

Total download size: 12 k
Installed size: 3.2 k
Downloading packages:
(1/2): intel-compilers-devel-ohpc-2016-19.1.x86_64.rpm     | 7.8 kB   00:00     
(2/2): intel-compxe-2016.3-067.noarch.rpm                  | 4.2 kB   00:00     
--------------------------------------------------------------------------------
Total                                              123 kB/s |  12 kB  00:00     
Running transaction check
Running transaction test
Transaction test succeeded
Running transaction
  Installing : intel-compxe-2016.3-067.noarch                               1/2 
  Installing : intel-compilers-devel-ohpc-2016-19.1.x86_64                  2/2 

Scanning top-level dir = /opt/intel
  Verifying  : intel-compilers-devel-ohpc-2016-19.1.x86_64                  1/2 
  Verifying  : intel-compxe-2016.3-067.noarch                               2/2 

Installed:
  intel-compilers-devel-ohpc.x86_64 0:2016-19.1                                 

Dependency Installed:
  intel-compxe.noarch 0:2016.3-067                   

```

The package has a dep on intel-compxe, but this doesn't guarantee a working icc binary as required for the postinstallation script. So the automatic installation of OpenHPC from RPM packages doesn't really work. I think it would be better to ship with a static module file in this package.

LaHaine:
[...]

PR #349, opened 2016-12-01
use find -L to follow symlinks

LaHaine:
This helps in case Intel MPI was installed on a shared file
system. See issue #254, that was fixed the same way.


koomie:
Thanks @LaHaine, I should have caught this back with #254, thanks for adding for the MPI compatibility module as well.

Issue #351, opened 2016-12-07
Add Listening Interface to wwnodescan command


nirmalasrjn:
This prevents wwnodescan from adding nodes from an interface other than the internal/provisioning interface.

koomie:
Documentation updated for 1.2.1 to include use of `--listen` option in example syntax.

PR #352, opened 2016-12-07
Add Listening Interface to wwnodescan command (Issue #351)

nirmalasrjn:
If the listen option is not included, then wwnodescan adds nodes
from the interface other than the provisioning interface.


koomie:
Landed on 1.2.1 branch. Thanks.

PR #358, opened 2017-01-10
mpiP: AArch64 fix

pawosm-arm:


pawosm-arm:
The backtrace() unwinder as used in mpiP doesn't work on AArch64. I managed to make internal setjmp()-based unwinder to work instead - with this patch mpiP test case passes on AArch64.

PR #362, opened 2017-01-12
EasyBuild 3.0.2 (WIP)

boegel:
(please don't consider this a blocker for OpenHPC v1.2.1, I'm just targeting that branch because I assume it's the most recent, feel free to change)

WIP because of a couple of questions I have:

* The OpenHPC module file is now in Tcl syntax, is that intentional? Should we convert it to Lua syntax instead (`EasyBuild/3.0.2.lua`), since Lmod is used as modules tool anyway?

* I can probably rework `easybuild.spec` to not rely on `bootstrap_eb.py` at all (and use `easy_install` instead), the actual installation itself will be identical; except for an EasyBuild-generated module file, but that seems to get in the way now anyway... Maybe that makes sense in this setup? Using the bootstrap script is desirable if you want an EasyBuild module file, but since OpenHPC is taking care of that itself...

* Even though this is a major version bump (`2.9.0` -> `3.0.2`), it's basically backwards compatible, see http://easybuild.readthedocs.io/en/latest/Release_notes.html#v3-0-0-november-16th-2016; especially in the OpenHPC setup where Lmod is being used already

* I removed the patches because they are no longer relevant, all those changes have been merged upstream a while ago, and are included in EasyBuild v3.0.2.


crbaird:
Landed on 1.3 branch. Thanks for the PR.

boegel:
@crbaird There's an EasyBuild v3.1.0 release already, and an upcoming v3.1.1 (sometime next week).

How much of a time window do I have to bump to the latest version for OpenHPC 1.3?

crbaird:
@boegel we don't have a hard release date, so let's see if we can land the 3.1.1

boegel:
@crbaird OK, I'll try and keep it in mind when I push out the v3.1.1 release of EasyBuild

PR #369, opened 2017-02-08
fix building on CentOS 7.3

adrianreber:
The included changes are necessary to get the openHPC packages building
on CentOS 7.3.

The result of this can be seen at:

https://copr.fedorainfracloud.org/coprs/adrian/ohpc/packages/

To actually get it working further changes are necessary. This first
commit includes missing or wrong named BuildRequires.


koomie:
Thank you @adrianreber for the consistency cleanup regarding the macro file definitions and other tweaks for CentOS 7.3

PR #382, opened 2017-02-28
enable jobcomp_elasticsearch plugin in slurm

LaHaine:
The plugin is built automatically if curl-config is found.


koomie:
Thanks @LaHaine, landed on 1.3 branch.

PR #386, opened 2017-03-08
EasyBuild v3.1.1 (WIP)

boegel:
(WIP because I haven't checked whether this passes the integration tests)

### DIFF
--- a/components/dev-tools/easybuild/SPECS/easybuild.spec
+++ b/components/dev-tools/easybuild/SPECS/easybuild.spec
@@ -15,12 +15,12 @@
 %define pname easybuild
 %define PNAME %(echo %{pname} | tr [a-z] [A-Z])
 
-%define vsc_base_ver 2.5.5
-%define vsc_install_ver 0.10.21
+%define vsc_base_ver 2.5.7
+%define vsc_install_ver 0.10.25
 
 Summary:   Build and installation framework
 Name:      EasyBuild%{PROJ_DELIM}
-Version:   3.0.2
+Version:   3.1.1
 Release:   1
 License:   GPLv2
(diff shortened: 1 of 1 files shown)


components/dev-tools/easybuild/SPECS/easybuild.spec
```diff
@@ -66,7 +66,6 @@ cp %{_sourcedir}/*py .
 
 export EASYBUILD_BOOTSTRAP_SOURCEPATH=%{_sourcedir}
 export EASYBUILD_INSTALLPATH=%{install_path}
-export EASYBUILD_MODULE_SYNTAX=Tcl
```
boegel: I included this change sort of as a reminder to self.

This controls the syntax used by EasyBuild to generate module files, in particular in the syntax of the EasyBuild module file that is generated as a part of the bootstrap procedure.

I'm not sure if matters at all, because the module file generated by EasyBuild is removed later (see the `rm %{version}` below, which is now likely going to fail since the module file generated by EasyBuild will now be `3.1.1.lua`).

The OHPC module file that is created is written in Tcl syntax however (see line 80 & onwards in this `.spec`).

Is that intentional?

Even though Lmod can consume module files in both Tcl and Lua, Lua should be preferred since that involves less overhead (Lmod translates Tcl module file to Lua on the fly via a `tcl2lua` tool every time a Tcl module file is consumed).

cc @crbaird

crbaird:
hi @boegel. The decision to go with TCL module files in OpenHPC is a legacy holdover from convincing sites to use lmod in the first place. We may eventually switch to lua flavored modules, but we'd like to do it consistently across the board.

PR #387, opened 2017-03-13
Add support to build openHPC packages with different gnu compilers versions

adrianreber:
The following commits are a bit larger change. This adds the possibility to change between the original gnu compiler (5.4.0) a gnu 6.x compiler based on the software collection devtoolset-6 (https://www.softwarecollections.org/en/scls/rhscl/devtoolset-6/) and a gcc 7 based gnu-compiler package.

To change the used compiler the compiler_family variable needs to be set to either gnu (5.4), dts6 (6.x) or gnu7 (pre gcc 7).

The results of the rebuilds can be found in my copr repositories:

https://copr.fedorainfracloud.org/coprs/adrian/ohpc-gcc7/packages/
https://copr.fedorainfracloud.org/coprs/adrian/ohpc-dts6/packages/

To make it easier to add additional compilers many (almost) identical macros have been move from the individual SPEC files to the common OHPC_macros files. Thus every compiler dependent SPEC had to be changed. So, alltough many files had to be changed the good thing is that a lot of code could be removed:

45 files changed, 617 insertions(+), 1702 deletions(-)

As this change also required adapting the OHPC_setup_compiler file I moved this file to a central package so that it exists only once and does not need to be part of every SPEC file.
The two new base packages are ohpc-filesystem which is required now by every openHPC package. This package only provides the base directories of the openHPC tree. Instead of every package owning /opt/ohpc this is now part of this new package. Details in that corresponding commit message.
The other new base package is ohpc-buildroot which provides OHPC_setup_compiler and OHPC_setup_mpi. In addition ohpc-buildroot also requires the lmod-ohpc package. This package seems to be automatically pulled in by the OpenSuse Build Service. Instead of relying on implicit functionality of the build system those dependencies are now explicitly listed.

The new macros are now %ohpc_compiler which replaces the initial compiler section of almost every SPEC file. The second new macro is %ohpc_setup_compiler which replaces the explicit call to OHPC_setup_compiler by a single macro.

In addition unnecessary parts of the spec files have been removed. ldconfig never needs to be executed as the libraries are not in paths ldconfig looks into. '%clean' section are also not required in RPM for a long time. BuildRoot is also always set by RPM automatically.

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -24,6 +24,7 @@ BuildRequires: ohpc-buildroot
 
 # Set the compiler to devtoolset-6 if it is on a RHEL 7 system
 %if 0%{?rhel} == 7
+	%{!?compiler_family: %global compiler_family dts6}
 	# OBS define's rhel_version to 700 for RHEL, in case we are
 	# not running in OBS
 	%global rhel_version 700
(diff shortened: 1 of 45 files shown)


adrianreber:
Rebased to avoid conflicts

koomie:
@adrianreber, you may have noticed I landed this on 1.3.1. In the end, I wasn't able to use it 100% as is in our build system, but I tried to keep the spirit of the changes and expanded the centralization effort for MPI families as well.  I created #447 to track some follow on work and document the latest approach. It's similar to what you had, but uses variables to toggle the compiler/mpi family settings (kind of like your `ohpc_bootstrap` flag).

Issue #388, opened 2017-03-15
slurm installs both LSB init script and systemd unit files


LaHaine:
I've noticed that slurm-ohpc in ohpc 1.2.1 installs both the LSB init script in /etc/init.d and the systemd unit files. This is a possible conflict, if the init script was registered by chkconfig and the administrator is calling systemctl with the unit file, there are misleading errors caused by the LSB compatibility of systemd. I propose a complete switch to systemd unit files.

PR #389, opened 2017-03-15
use systemd units in slurm

LaHaine:
This is my proposed fix for issue #388 Maybe the LSB init scripts can be removed later as well, I am not sure about SLES.


adrianreber:
Just had a look at the current `slurm.spec` file in the 2.x branch and it seems this has been resolved some time ago.

PR #433, opened 2017-04-26
fix if statment in conman spec file

rberg2:
Signed-off-by: Richard Berg <rchrd.berg@gmail.com>


koomie:
Thanks for this fix. We'll get it queue up for the next planned release.

PR #434, opened 2017-04-26
valgrind: AArch64 fix (imported from Valgrind SVN)

pawosm-arm:
This is to fix following issue: https://bugs.kde.org/show_bug.cgi?id=369459


koomie:
Thanks @pawosm-arm

PR #439, opened 2017-05-02
Make Lmod conflicts to Modules

pliniosilveira:
Make Lmod conflicts to Modules (Suse's environment modules).


koomie:
Thanks for this suggestion. Landed on 1.3.1 brach.

PR #445, opened 2017-05-10
Remove post-build-checks conflicts for slurm,munge,warewulf

a74nh:
slurm, munge and warewulf all use BuildConflicts for post-build-checks.
I noticed that almost every other package in ohpc has #!BuildIgnore instead.
I'm not sure if this difference is intentional or not.

Privately, I'm using OBS 2.8 (build.openhpc.community uses OBS 2.7), and these three packages fail to build unless I make these changes.

According to this url, OBS previously didn't handle buildconflicts correctly and now does: https://github.com/openSUSE/open-build-service/issues/2004


koomie:
Thanks for these. Landed on the 1.3.1 branch.

PR #449, opened 2017-05-15
compiler-families: LLVM compiler family

pawosm-arm:


pawosm-arm:
https://github.com/openhpc/submissions/issues/18

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
```diff
@@ -0,0 +1,191 @@
...
+# It may have been modified from the default version supplied by the underlying
+# release package (if available) in order to apply patches, perform customized
+# build/install configurations, and supply additional files to support
+# desired integration conventions.
+#
+#----------------------------------------------------------------------------eh-
+
+%include %{_sourcedir}/OHPC_macros
+
+%global gnuver 6.3.0
```
pawosm-arm: The trouble with our targeted Linux distros is that they operate on a bit ancient versions of gcc/g++. Although it may be possible to build LLVM compilers (with all their runtime libraries) using gcc 4.8 (as in Centos 7.3), it does not necessary mean that this build is correct. For example, a part of LLVM project is libomp - OpenMP runtime library which is heavily using compiler builtins and c++11 atomics. This runtime library is built along with (possibly in parallel) the rest of the LLVM project using exactly the same (system?) compiler that builds the other parts. We need to be careful here, as for AArch64 architecture we may face effects of gcc bug like this: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=65697
The version selected was 6.3.0 because that was a valid selection before 15th of May 2017 when things changed to 7.1.0 due to commit 6ea6464b0c3115af379734cf08273e56dd5760bb.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
```diff
@@ -0,0 +1,191 @@
...
+%global cmakemajor 3.8
+%global cmakever 3.8.1
```
pawosm-arm: It's a very good idea. Can one submission cover two things, namely compiler-family and CMake? Otherwise, just to be able to move from place, I'll stick with this inner cmake for a while until cmake appear in OpenHPC repo.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
```diff
@@ -0,0 +1,191 @@
...
+
+cd llvm-%{version}.src/projects
+ln -s ../../openmp-%{version}.src openmp
+cd ../..
+
+ln -s llvm-%{version}.src llvm
+ln -s cmake-%{cmakever} cmake
+
+%build
+
```
pawosm-arm: For some reason many of the OpenHPC packages do it that way. I just followed the order that I encountered.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
```diff
@@ -0,0 +1,191 @@
...
+module-whatis "Category: compiler, runtime support"
+module-whatis "Description: LLVM Compiler Infrastructure"
+module-whatis "URL: http://www.llvm.org"
+
+set     version                            %{version}
+
+prepend-path    PATH                %{install_path}/bin
+prepend-path    MANPATH             %{install_path}/share/man
+prepend-path    INCLUDE             %{install_path}/include
+prepend-path    LD_LIBRARY_PATH     %{gnu_path}/lib64:%{install_path}/lib
```
pawosm-arm: Most of the LLVM and Clang is writen in C++, if we build it with non-distro compiler, it needs to have this path to libs provided.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
```diff
@@ -0,0 +1,191 @@
...
+## version file for %{pname}-%{version}
+##
+set     ModulesVersion      "%{version}"
+EOF
+
+%post
+/sbin/ldconfig
+
+%postun
+/sbin/ldconfig
```
pawosm-arm: good point

pawosm-arm:
There is one thing missing here: the Fortran compiler. Now we can freely add it:
https://github.com/flang-compiler which I plan to do for the second iteration of this PR.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
adrianreber: Ah, ok, makes sense. I am still worried about the implicit dependency on the gcc compiler in the module file. Is there a way to build llvm in such a way that it becomes independent of the compiler it has been built with?

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
adrianreber: Understandable. I have been trying to correct it in a few spec files lately, that is the reason I have seen it.

adrianreber:
I have started to compile the openHPC packages with your llvm package:

https://copr.fedorainfracloud.org/coprs/adrian/ohpc-clang/packages/

Not much right now, but I will continue. The main changes are to disable fortran if the compiler_family is set to llvm.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
pawosm-arm: as a static library. I doubt we like it, it's bloated.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
```diff
@@ -0,0 +1,191 @@
...
+%define debug_package %{nil}
+
+BuildRequires: gnu-compilers%{PROJ_DELIM}
+BuildRequires: make
+BuildRequires: perl
+BuildRequires: python
+BuildRequires: pkgconfig
+BuildRequires: binutils-devel
+Requires:      binutils
+Requires:      gnu-compilers%{PROJ_DELIM} = %{gnuver}
```
pawosm-arm: I don't see how (or maybe it's not 1.3.1 branch yet)

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
adrianreber: Static linking sounds wrong, yes. But a boostrap build like described here:

http://releases.llvm.org/4.0.0/docs/AdvancedBuilds.html

Would that work and remove the gcc dependency?

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
adrianreber: Opening a github issue might be a good idea to track the cmake bundling and the need for a separate cmake package.

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
adrianreber: I did the following:

```--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -77,6 +77,10 @@ Requires:      gnu-dts6-compilers%{PROJ_DELIM}
 BuildRequires: gnu-7-compilers%{PROJ_DELIM}
 Requires:      gnu-7-compilers%{PROJ_DELIM}
 %endif
+%if "%{compiler_family}" == "llvm"
+BuildRequires: llvm-compilers%{PROJ_DELIM}
+Requires:      llvm-compilers%{PROJ_DELIM}
+%endif
 
 %endif
 
```

```
--- a/components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
+++ b/components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
@@ -8,6 +8,7 @@
 #
 #----------------------------------------------------------------------------eh-
 
+%define ohpc_compiler_dependent 1
 %include %{_sourcedir}/OHPC_macros
 
 %global gnuver 6.3.0
```
This will pull in the compiler define in compiler_family.

Github's code reviews are sometimes hard to read. Following lines can be removed

```
-Source10:  OHPC_setup_compiler
-%define debug_package %{nil}
-
-BuildRequires: gnu-compilers%{PROJ_DELIM}
-Requires:      gnu-compilers%{PROJ_DELIM} = %{gnuver}
```

And this change:

```
@@ -100,8 +96,7 @@ ln -s cmake-%{cmakever} cmake
 %build
 
 %install
-export OHPC_COMPILER_FAMILY=gnu
-. %{_sourcedir}/OHPC_setup_compiler
+ %ohpc_setup_compiler
 
 mkdir build-cmake
 cd build-cmake
```

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
pawosm-arm: Done: https://github.com/openhpc/submissions/issues/21

pawosm-arm: and this: https://build.openhpc.community/package/show/home:pawel:branches:OpenHPC:1.3:Update1:Factory/cmake

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
pawosm-arm: Smells like gcc building process (which takes ages), one thing that I like about LLVM is that I don't need to do this multi-stage builds.
Besides, gcc still produces better code than LLVM.
I'm opting to use OHPC gnu-compilers for building LLVM for now and leave comment that we need to drop it later, when all the targeted Linux distros start to use more recent gcc (e.g. 5.3+).

components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
pawosm-arm: I won't be able to derive gnu_path not having gnuver. Since it's used in the module file, installed gnu-compilers version must match with the one referred. This is ensured by following line:

`Requires:      gnu-compilers%{PROJ_DELIM} = %{gnuver}
`

pawosm-arm:
Improved version force-pushed

adrianreber:
One more question concerning the used bootstrap compiler as there was mail thread on openhpc-devel about problems on aarch64.

Did you try building llvm with the OS compiler on CentOS 7.3 and you get the failure you mentioned or was it a failure on CentOS 7.2. If possible I still would like to avoid the gcc-7 dependency.

adrianreber:
Closing as the 1.3.x branch is inactive for a long time.

PR #455, opened 2017-05-24
Also simplify the mpi selection

adrianreber:
With the new flag %%ohpc_mpi_dependent the macro %%ohpc_setup_compiler
can automatically run the necessary things for MPI dependent packages.

With this change OHPC_setup_mpi is also removed from all SPEC files
as the ohpc-buildroot supplied version can be used by all packages.

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -56,7 +56,7 @@ BuildRequires: ohpc-buildroot
 # Compiler dependencies
 %if 0%{?ohpc_compiler_dependent} == 1
 
-%if "%{compiler_family}" == "gnu" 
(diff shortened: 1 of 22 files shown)


koomie:
More nice work here @adrianreber, thanks for the improvements. I landed this on 1.3.1 with the exception of bumping the release string. In the ohpc builds, OBS is managing that string for us so I like to leave it a canonical value. I realize we have some other specs that are not consistent, so I'll try to clean those up separately.

adrianreber:
@koomie Thanks for merging. I was not aware of OBS's release handling. Thanks for letting me know.

PR #456, opened 2017-05-24
adding newer CMake as needed by LLVM

pawosm-arm:


adrianreber:
Thanks for the separate cmake package. I guess the 

compiler-families: LLVM compiler family

commit does not really belong in this PR. I will test the package tomorrow, but I have a few comments until then.

components/dev-tools/cmake/SPECS/cmake.spec
```diff
@@ -0,0 +1,94 @@
...
+URL: https://cmake.org/
+Source: https://cmake.org/files/v%{major_version}/cmake-%{version}.tar.gz
+BuildRequires:  gcc-c++
+BuildRequires:  ncurses-devel
+BuildRequires:  curl-devel
+BuildRequires:  xz-devel
+BuildRequires:  zlib-devel
+BuildRequires:  pkgconfig
+
+%define install_path %{OHPC_PUB}/%{pname}/%version
```
adrianreber: Not sure where the cmake in the OpenHPC package hierarchy belongs, but I would recommend not a top level directory but rather %{OHPC_PUB}/apps/%{pname}/%version . I see that other packages are also installed directly in the top level pub directory, so it makes kind of sense if looking at the other packages but apps still seems nicer as a top-level directory.

components/dev-tools/cmake/SPECS/cmake.spec
```diff
@@ -0,0 +1,94 @@
...
+%{__cat} << EOF > %{buildroot}/%{OHPC_MODULES}/%{pname}/.version.%{version}
+#%Module1.0#####################################################################
+##
+## version file for %{pname}-%{version}
+##
+set     ModulesVersion      "%{version}"
+EOF
+
+%clean
+%{__rm} -rf $RPM_BUILD_ROOT
```
adrianreber: %clean is no longer needed with recent RPMs. CentOS 7 for sure and I guess the same is true for SLES.

components/dev-tools/cmake/SPECS/cmake.spec
```diff
@@ -0,0 +1,94 @@
...
+
+%files
+%defattr(-,root,root,-)
+%{OHPC_PUB}
```
adrianreber: I would say %{OHPC_PUB}/apps/%{pname} instead of %{OHPC_PUB} and another line for the modulefiles

components/dev-tools/cmake/SPECS/cmake.spec
```diff
@@ -0,0 +1,94 @@
...
+
+%changelog
```
adrianreber: I like changelogs

components/dev-tools/cmake/SPECS/cmake.spec
```diff
@@ -0,0 +1,94 @@
...
+
+%define pname cmake
+
+%define major_version 3.8
+%define minor_version 1
+
+Summary: CMake is an open-source, cross-platform family of tools designed to build, test and package software.
+Name: %{pname}%{PROJ_DELIM}
+Version: %{major_version}.%{minor_version}
+Release: %{_rel}%{?dist}
```
adrianreber: is %{_rel} defined anywhere?

pawosm-arm:
Hi,
The second commit indeed doesn't belong to this PR, still I'm working on the same branch on the LLVM compiler (fortunately, there's some progress in it).

pawosm-arm:
modified.

chuckatkins:
Some of the dependent packages are missing.  CMake will still build since it uses it's own internal copies of things but to properly use system supplied dependencies, the full list of dev-packages is: `bzip2-devel`, `curl-devel`, `expat-devel`, `ncurses-devel`, `jsoncpp-devel`, `libarchive-devel`, `xz-devel`, `rhash-devel`, `libuv-devel`, and `zlib-devel`. 

Also, `configure` is just a wrapper around the bootstrap script.  Really you'll
want to call it directly instead: `./bootstrap --prefix=${install_path} --system
-libs --no-qt-gui` (actually without the `--system-libs` option none of the dependencies get used anyways).

Lastly, since we're updating, at this point the 3.8.2 patch release is available
 instead of 3.8.1.

pawosm-arm:
Hi,

I wouldn't consider myself as regular either.

Changes pushed:
- Updated 3.8.1 to 3.8.2
- used ./bootstrap with provided flags instead of ./configure wrapper
- added devel-dependencies but only those that resolve on OpenHPC OBS

adrianreber:
@koomie I would say this is ready to be merged. Probably the 1.3.2 branch? Do we need to go through submissions?

koomie:
Since this is a dependency for an item already accepted I don't think we need to file a separate submission. We're a bit heads down at the moment to prepare for a tutorial at the PEARC conference, but I will get a 1.3.2 build area setup in OBS and we can land this there.  Thanks.

crbaird:
pulled in 0a88af3

PR #459, opened 2017-05-31
OHPC_macros: fix build outside of OBS

adrianreber:
When building the OpenHPC packages outside of OBS some builds for
CentOS/RHEL are failing as the necessary macros are not available. This
sets the 'rhel_version' macro to 700 if the 'rhel' macro equals 7.


koomie:
Thanks.

PR #460, opened 2017-06-01
singularity: few small fixes to singularity.spec

adrianreber:
The main reason for the changes was that the package was not building
without Source1: OHPC_macros outside of OBS.

Additionally:
 * removed %%_rel macro as it is not used in OpenHPC
 * removed unused definition of PNAME
 * removed BuildRoot as it is automatically set by RPM
 * moved building from the %%install to the %%build section
 * replaced %%{?mflags_install} in %%install with
   %%{?_smp_mflags} in %%build
   * neither %%{?mflags} nor %%{?mflags_install} is defined
     on CentOS:
     # rpm --eval '%{?mflags}'

     # rpm --eval '%{?mflags_install}'

   * but %%{?_smp_mflags} is defined:
     # rpm --eval '%{?_smp_mflags}'
     -j2
 * removed %%clean section (automatically inserted by RPM)

### DIFF
--- a/components/runtimes/singularity/SPECS/singularity.spec
+++ b/components/runtimes/singularity/SPECS/singularity.spec
@@ -8,12 +8,10 @@
 #
 #----------------------------------------------------------------------------eh-
 
-%{!?_rel:%{expand:%%global _rel 0.1}}
 %include %{_sourcedir}/OHPC_macros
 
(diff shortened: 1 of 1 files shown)


crbaird:
Thanks, @adrianreber

PR #477, opened 2017-07-04
nagios: fix building on RHEL/CentOS

adrianreber:
Building the nagios package outside of OBS failed as it was missing the
BR on systemd. This adds the necessary BuildRequires and cleaned up the
nagios spec file a bit:

 * removed setting PROJ_DELIM as it comes from OHPC_macros
 * removed setting PNAME as it is not used
 * removed BuildRoot - automatically defined by RPM
 * fixed systemd Requires and BuildRequires
 * removed German summary as no other package has localized summaries
 * removed trailing whitespaces
 * removed cleaning of buildroot - done by RPM
 * use predefined systemd macros for service control in preun/post/postun

### DIFF
--- a/components/admin/nagios/SPECS/nagios.spec
+++ b/components/admin/nagios/SPECS/nagios.spec
@@ -9,11 +9,9 @@
 #----------------------------------------------------------------------------eh-
 
 %include %{_sourcedir}/OHPC_macros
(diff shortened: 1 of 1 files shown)


adrianreber:
I have not tested this on suse. I guess (hope) that the systemd rpm macros are the same.

PR #478, opened 2017-07-04
[RFC] openmpi: support multiple Open MPI versions in parallel

adrianreber:
In order to support the installation of multiple Open MPI versions in
parallel and the compilation of dependent software packages, the
Open MPI spec file, OHPC_macros and OHPC_setup_mpi has been enhanced.

In addition to defining mpi_family like mvapich2, mpich and openmpi the
mpi_family can be defined as openmpiv1.10, openmpiv2.0, openmpiv2.1 and
openmpiv3.0 (rc1 currently). This enables to compile and install the
complete Open MPI dependent software stack in parallel to each other.

Using 'modules av' gives now the following after loading the compiler
module:

            ------ /opt/ohpc/pub/moduledeps/gnu ------
 openmpi/1.10.7    openmpi/2.0.3    openmpi/2.1.1    openmpi/3.0.0rc1 (D)

Loading one of the Open MPI modules makes software based on that version
available:

            -------- /opt/ohpc/pub/moduledeps/gnu-openmpiv2.1 ------
 imb/4.1

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -73,6 +73,9 @@ BuildRequires: ohpc-buildroot
 	%global mpi_family openmpiv1.10
 %endif
 
+# This returns 'openmpi' without version suffix, if openmpi* is selected.
+%global openmpi %(eval echo -n %{mpi_family} | cut -c 1-7)
+
 # Compiler dependencies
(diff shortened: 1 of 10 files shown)


adrianreber:
Updated with additional fixes to build all packages with all Open MPI versions.

PR #481, opened 2017-07-06
fix building tau and cube and some cleanups

adrianreber:


koomie:
Thanks @adrianreber. Landed on 1.3.2.

PR #519, opened 2017-08-03
Update the install guide for Aarch64

a74nh:
I was browsing the Aarch64 install guide and noticed a few differences compared to x86:

1) The "Enable forwarding of system logs" section was in a different place.

2) The beegfs section was missing. It's possible this was intentional, however the rest of the aarch64 doc does reference beegfs throughout.

3) Some of the tables in the appendix "Package Manifest" were out of date. This was only for the tables without gnu7 references in them.

The document is also incorrect for 1.3.1 and 1.3 in the same ways, but it's probably too late to update now.

### DIFF
--- a/docs/recipes/install/centos7/aarch64/warewulf/slurm/manifest/admin.tex
+++ b/docs/recipes/install/centos7/aarch64/warewulf/slurm/manifest/admin.tex
@@ -8,6 +8,13 @@
 {\bf RPM Package Name} & {\bf Version} & {\bf Info/URL}  \\ 
 \midrule
 
+% <-- begin entry for clustershell-ohpc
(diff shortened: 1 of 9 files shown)


koomie:
@a74nh, note that I landed this on the 1.3.2 branch in preparation for the next release but reverted commit c45629606928297cc7e2610fb7c4cb55334b88bb.  The reason beegfs was not in there is because we are currently instructing users to leverage public beegfs repos available at https://www.beegfs.io/release/beegfs_6/dists/

At present, they do not have aarch64 builds published, so we didn't call out the installation in the recipe.  

PBS was not likely in the manifest as we don't have a test cluster for it, but happy to put it in there.  Note that the manifest versions will be changing here as we update for the 1.3.2 repo configuration.

PR #522, opened 2017-08-08
Adding xCAT stateful recipe

mrmhodak:
Adding xCAT stateful recipe -  CentOS Slurm only for now. Will add more later.

The provisioning strategy is to have have xCAT first install a generic OS image and do OpenHPC installation and other customization using a postbootscript run by xCAT after OS install and reboot. The process pretty much exactly follows chroot image creation in stateless recipes, with the exception that commands  are added to script the file for later execution (rather than executed in chroot environment).  Also, all services on the compute nodes need to be explicitly started because the nodes will not be rebooted again. Additionally, local repo mirrors are used for installation on the nodes so that nodes do not have to go to the Internet to get packages.

Multiple other changes including: 
- Using xCAT command wherever possible- psh to run commands on all nodes, rcons to access SOL consoles, rpower and rsetboot to start provisioning process. 

- Adjustments to parse_doc.pl needed to extract version number to download the right tarball. (BTW, I cannot see any way for the script to find the full version number (i.e. 1.3.1). That is stored in ohpc-doc.sty, which is not read by the script. For now I have created \VERLONG command in steps.tex)

Several files have been split into "_stateful" version even though changes are minor, but I could not find a toggle reliably working for "lslisting" environment.

Le me know if anything needs to be changed/fixed.

### DIFF
--- a/docs/recipes/install/common/reset_computes_xcat.tex
+++ b/docs/recipes/install/common/reset_computes_xcat.tex
@@ -37,4 +37,4 @@ c4:  08:53:36 up 5 days, 18:30,  0 users,  load average: 0.00, 0.01, 0.05
 \end{lstlisting}
 Another  useful way of monitoring provisioning is to use  the \texttt{rcons} command,
 which displays serial console for a selected node. Note that the escape sequence
-is \texttt{CTRL-e CTRL-c .} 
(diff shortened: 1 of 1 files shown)


mrmhodak:
#575 is the request to merge, this is now obsolete

crbaird:
tests passing in CI

PR #569, opened 2017-10-11
Update NFS mount options in docs

jabl:
- By not setting rsize/wsize, the client and server will autonegotiate
  the maximum sizes, leading typically to the best performance.

- cto is the default NFS-standard conforming semantics so no need to
  specify it explicitly.

- nodev,nosuid are for security. If /opt/ohpc/pub contains no suid
  binaries, one could add nosuid to that mount as well.

- noatime is good for performance, and atime updates are typically
  useless.

### DIFF
--- a/docs/recipes/install/common/warewulf_chroot_customize_centos.tex
+++ b/docs/recipes/install/common/warewulf_chroot_customize_centos.tex
@@ -17,8 +17,8 @@ example configuration.
 [sms](*\#*) cat ~/.ssh/cluster.pub >> $CHROOT/root/.ssh/authorized_keys
 
 # Add NFS client mounts of /home and /opt/ohpc/pub to base image
-[sms](*\#*) echo "${sms_ip}:/home /home nfs nfsvers=3,rsize=1024,wsize=1024,cto 0 0" >> $CHROOT/etc/fstab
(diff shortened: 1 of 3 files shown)


adrianreber:
Just a quick comment on nosuid. Singularity and pbspro rely for a few binaries on suid

https://github.com/openhpc/ohpc/blob/obs/OpenHPC_1.3.3_Factory/components/runtimes/singularity/SPECS/singularity.spec#L154

So that nosuid sounds problematic for those two cases at least.

jabl:
Thanks for the heads-up. Yes, that means that /opt/ohpc/pub should not be mounted nosuid, although it shouldn't be a problem for /home.

Now that you mention Singularity, does nodev break mounting container images? If so, /home shouldn't be mounted nodev I guess? Ostensibly, nodev should prevent character or block special files, but container images are regular files. However inside the container image there's /dev, /proc and such, might that be a problem?

jmstover:
@gmkurtzer
Can you comment on the nodev option to a mount with Singularity? I... don't know about that.

gmkurtzer:
The `nodev` mount option should not be a problem for Singularity. The only NFS mount options that have been a problem is `no_root_squash` on a directory where a Singularity image file exists, and `nosuid` on an exported directory where Singularity has been installed.

PR #572, opened 2017-10-12
fix building prun and meta-packages due to missing files

adrianreber:


adrianreber:
This is pretty simple as it only adds the missing LICENSE files to the SOURCES directory. This could also be a link or we could generally move the LICENSE file to the SOURCES directory instead of being in one directory above.

adrianreber:
I pushed some more fixes. The second commit tries to add '3's in two more openmpi dependent spec files.

PR #575, opened 2017-10-13
Adding xCAT stateful recipe

mrmhodak:
This is an equivalent to original  request #522 ,  but updated to 1.3.3 with minor adjustments. I will close the old request.

Here is a full description from the previous PR:

Adding xCAT stateful recipe - CentOS Slurm only for now. Will add more later.

The provisioning strategy is to have have xCAT first install a generic OS image and do OpenHPC installation and other customization using a postbootscript run by xCAT after OS install and reboot. The process pretty much exactly follows chroot image creation in stateless recipes, with the exception that commands are added to script the file for later execution (rather than executed in chroot environment). Also, all services on the compute nodes need to be explicitly started because the nodes will not be rebooted again. Additionally, local repo mirrors are used for installation on the nodes so that nodes do not have to go to the Internet to get packages.

Multiple other changes including:

Using xCAT command wherever possible- psh to run commands on all nodes, rcons to access SOL consoles, rpower and rsetboot to start provisioning process.

Adjustments to parse_doc.pl needed to extract version number to download the right tarball. (BTW, I cannot see any way for the script to find the full version number (i.e. 1.3.1). That is stored in ohpc-doc.sty, which is not read by the script. For now I have created \VERLONG command in steps.tex)

Several files have been split into "_stateful" version even though changes are minor, but I could not find a toggle reliably working for "lslisting" environment.

Le me know if anything needs to be changed/fixed.

### DIFF
--- a/docs/recipes/install/common/add_xcat_hosts_intro.tex
+++ b/docs/recipes/install/common/add_xcat_hosts_intro.tex
@@ -122,9 +122,6 @@ accomplished as follows:
 [sms](*\#*) makedhcp -n
 [sms](*\#*) makedns -n
 
-# Associate desired provisioning image for computes
-[sms](*\#*) nodeset compute osimage=BOSVER-x86_64-netboot-compute
-
 \end{lstlisting}
 
 %%% If the Lustre client was enabled for computes in \S\ref{sec:lustre_client}, you
--- a/docs/recipes/install/common/synclist_osimage.tex
+++ b/docs/recipes/install/common/synclist_osimage.tex
@@ -7,7 +7,7 @@ The following associates the prepared \texttt{synclist} with the \installmode{} 
 
 % begin_ohpc_run
 % ohpc_comment_header Import files \ref{sec:file_import}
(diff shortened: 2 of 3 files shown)


koomie:
We're having some issues with the current xCAT stateful recipe which I need to iron out for the next 1.3.3 release. Consequently, will schedule this stateful recipe for 1.3.4 milestone.

mrmhodak:
I have updated the xCAT stateful recipe to the 1.3.4 branch and will send a pull request shortly.

In the process, I have changed the approach: Previous iterations defined a post-install script for OpeHPC component installation, which was run by xCAT as a part of the nodes’ installation process. This works, but makes finding and fixing errors in the script difficult (execution happens away from user’s eyes).

The new approach is to install OS on the nodes early in the process by using a generic OS image automatically created by xCAT. OpenHPC components are installed once the nodes are installed and running. Effectively, this is very similar to what happens in a stateless approach except that instead of installing into a chroot environment,  the installations happen on running nodes (accomplished by using xCAT’s parallel execution command psh).

I will submit in the next few days, after I do one more test on my main test server, which is currently down.

In the meantime, let me know if you have any comments/questions.

Thanks.

mrmhodak:
Closing this one, it is obsolete now. PR #643 should be used instead.

PR #576, opened 2017-10-16
[docs] Add simple checks to base install commands

rengolin:
This is a first attempt at improving the command lines in the documents. 

Most of them just allow safe copy and paste across by checking already updated files, thus avoiding duplicated commands. The bulk is, therefore, uncontroversial, as long as we're happy to have the checks on most places. If this is the case, and this patch is accepted, I'll create a pull request to all others.

The more complicated changes will be submitted after the initial cleanup.

Reference repo: https://git.linaro.org/leg/hpc/ohpc-scripts.git/

This change: https://git.linaro.org/leg/hpc/ohpc-scripts.git/tree/centos7/00_base_install.sh

### DIFF
--- a/docs/recipes/install/centos7/aarch64/warewulf/slurm/steps.tex
+++ b/docs/recipes/install/centos7/aarch64/warewulf/slurm/steps.tex
@@ -82,8 +82,10 @@
 % ohpc_validation_newline
 % ohpc_validation_comment Disable firewall 
 \begin{lstlisting}[language=bash,keywords={}]
-[sms](*\#*) systemctl disable firewalld
-[sms](*\#*) systemctl stop firewalld
+[sms](*\#*) if systemctl status firewalld; then
+              systemctl disable firewalld
+              systemctl stop firewalld
+            fi
 \end{lstlisting}
 % end_ohpc_run
 
--- a/docs/recipes/install/common/bos.tex
+++ b/docs/recipes/install/common/bos.tex
@@ -13,7 +13,9 @@ installed the BOS, there may be an adequate entry already defined
 in \path{/etc/hosts}. If not, the following addition can be used to identify
 your SMS host.
 \begin{lstlisting}[language=bash,keywords={}]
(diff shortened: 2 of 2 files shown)


koomie:
I would suggest to not continue this particular approach as it will complicate/bloat the recipe examples and will, in the end, still result in something that is not 100% re-runnable. Recall that the target audience assumption called out specifically in the docs is for experienced Linux administrators. In the same we expect they can understand regex'es, I think it's reasonable to assume they understand runlevel services and not appending a file twice.

That being said, I understand the motivation and desire to potentially have a companion "installer" which is re-runnable. We provide a bash script now that we can 100% validate in CI which is good, but as you know, it does not preserve state. When we have talked about this before (spurred by discussion at last year's SC BoF), the thought was that seasoned users are likely to put all of this under configuration management control, and that's the right way to maintain system state. Unfortunately, there is no single config management system we could choose to make everyone happy, so we prefer a simple example script (a la what we provide with recipe.sh) for validation.  

There was some interest expressed in creating an installer that could more automatically apply an OpenHPC recipe (possibly using one or more configuration management systems). I started a prototype of that earlier this year, but haven't had a chance to work on that lately. The general approach though is to introduce more keyword options in our latex recipes (so as opposed to your systemd checks, we introduce additional keyword macros like `(*\enableservice*) httpd.service` and `(*\appendfile*)` and to append to a file.  Those can then be leveraged by a separate installer to implement stateful implementations (and could potentially help a less seasoned user walk thru the installation easier).  With this approach, we could still have a single input (our latex docs) that would generate the bash recipe that we have today, along with a companion input script for an installer that would ultimately allow folks to implement key system state changes in multiple ways.

Having started some of this, I realize it is non-trivial effort. Once we get through the current high-priority items for v1.3.3, we can dig in deeper and I'll show you what I have done on this front.

rengolin:
Hi Karl,

Makes sense. I have to work on Ansible scripts for some of those, so I'll just abandon the docs changes and will do that internally, so that later we can share and collaborate on some standard way of doing that.

Thanks!
--renato

PR #577, opened 2017-10-17
Update warewulf stateful install doc to use ext4

jabl:
Also increase the size of /boot in the example to 1 GB.

### DIFF
--- a/docs/recipes/install/common/stateful.tex
+++ b/docs/recipes/install/common/stateful.tex
@@ -18,9 +18,9 @@ filesystem layout will look like.
 
 \begin{lstlisting}[language=bash,literate={-}{-}1,keywords={},upquote=true]
 # Update node object parameters
-[sms](*\#*) export sda1="mountpoint=/boot:dev=sda1:type=ext3:size=500"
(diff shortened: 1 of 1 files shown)


koomie:
Have you tested with ext4?

jabl:
Err, actually I'm having some problems on my virtual test cluster to get the stateful provisioning to work at all; it kernel panics on boot. Anyway, I get the same error message with the original ext3 instructions, so at least my patch didn't cause this. :)

But, I don't see any reason why ext4 wouldn't work. In /etc/warewulf/bootstrap.conf one can see that the ext4 driver is included; actually on many modern distros including centos7 the kernel is compiled with CONFIG_EXT4_USE_FOR_EXT23=y meaning that the ext4 driver is used for ext2/3 and there are no separate ext2/3 drivers.

jabl:
Well, anyway, it seems that as of 752666cacfb84ccc4edc9798b97c47e9819b7490 the openhpc docs reference the new warewulf 3.8 way of doing things, so this PR is moot. Closing.

Issue #589, opened 2017-10-27
Source URL of python-Cython  is wrong


takekato:
Hi,

I found that the URL in the spec file of python-Cython.
Apparently, the download URL format of python-Cython might change after the spec file was created.
Please refer following diff and if it is correct, would you please merge the patch 
from https://github.com/takekato/ohpc/tree/tkato-python-Cython.

diff --git a/components/distro-packages/python-Cython/SPECS/python-Cython.spec b/components/distro-packages/python-Cython/SP
index 102cefb..0cc6de3 100644
--- a/components/distro-packages/python-Cython/SPECS/python-Cython.spec
+++ b/components/distro-packages/python-Cython/SPECS/python-Cython.spec
@@ -33,7 +33,7 @@ Url:            http://www.cython.org
 Summary:        The Cython compiler for writing C extensions for the Python language
 License:        Apache-2.0
 Group:          Development/Languages/Python
-Source:         http://pypi.python.org/packages/source/C/Cython/Cython-%{version}.tar.gz
+Source:         https://pypi.python.org/packages/b3/89/e8dddc86bf4f144407962e04c4018f53dfe4d6ae5010d34573d8521e810e/Cython-
 Source1:        python-Cython-rpmlintrc
 Patch1:         python-Cython-c++11.patch
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build

adrianreber:
I am not sure what our procedure is, as this was reported against 1.3.2. Now that @koomie fixed it on 1.3.3 this could be closed. I do not seem to have the necessary rights to close this.

takekato:
Sorry, I confirmed the URL work today.
It might be temporary problem or my fault.
I'm sorry to bother you.
Thanks.

PR #590, opened 2017-10-27
Fix URL of SOURCE.

takekato:
As I mentioned on issue #589 , I found that the URL in the spec file of python-Cython.
Apparently, the download URL format of python-Cython might change after the spec file was created.


adrianreber:
Which error do you see? This resulting URL works for me:

https://pypi.python.org/packages/source/C/Cython/Cython-0.23.4.tar.gz

koomie:
Maybe the issue is that the current spec has `http` instead of `https`?  It does look like the http variant has stopped working, so I updated to use http with the same path.

adrianreber:
Same here. Fixed in 1.3.3 and could be closed now.

PR #600, opened 2017-11-09
Use ip instead of ifconfig

jabl:
net-tools hasn't seen a release since 2001, whereas iproute2 is
actively developed and included by default.  By modifying the setup
instructions to use ip instead of ifconfig net-tools can thus be
removed from the dependencies.


koomie:
I'm pretty sure we tried this a while back and while `ip` worked fine for standard physical interfaces, we ran into problems using it on aliased interfaces (e.g. `eth0:0`). I don't remember the exact details but I do remember having to back it out as it broke the CI system which does have some reliance on aliasing.  

I will try to recreate again to remember the details and perhaps we can find a solution.

PR #630, opened 2017-12-05
environment modules based dependencies

adrianreber:
In order to avoid base OS breakage due to duplicate or wrong
dependencies, this introduces a completely different dependency model
for OpenHPC.

The main problem with the automatically generated SO based dependencies
is that OpenHPC packages can have the same SO 'Requires' or 'Provides'
as packages installed by the base OS.
As the OpenHPC packages are installed outside of the dynamic loader's
scope the dependencies are correctly solved in the RPM database but not
in the running system.

To avoid those problems this patch disables the RPM internal dependency
generator and replaces it with a environment module based
implementation.

This new environment module based generator creates dependencies like:

 # repoquery --requires openblas-gnu7-ohpc
 lmod-ohpc
 ohpc-filesystem
 ohpc-module(gnu7)
 # repoquery --provides openblas-gnu7-ohpc
 ohpc-module(openblas)
 ohpc-module(openblas/0.2.20)
 openblas-gnu7-ohpc = 0.2.20-1.ohpc
 openblas-gnu7-ohpc(x86-64) = 0.2.20-1.ohpc
 # repoquery --requires tau-gnu7-openmpi3-ohpc
 lmod-ohpc >= 7.6.1
 ohpc-module(gnu7-openmpi3)
 ohpc-module(papi)
 ohpc-module(pdtoolkit)

The created dependencies can directly be used in the 'module load'
syntax.

The new dependency generators are looking for environment module files
in 'modulefiles/' and 'moduledeps/' and create the dependencies based
on paths, filenames, prepend-path and depends-on.

Besides the infrastructure changes in this commit, the following changes
in the RPM spec files can be made to simplify dependency handling.

If the variable 'ohpc_required_modules' is set to a valid environment
modules name the necessary 'BuildRequires:' and 'Requires:' statements
will be generated:

 llvm-compilers.spec: %global ohpc_required_modules cmake
 # rpm -qpR ~/rpmbuild/SRPMS/llvm5-compilers-ohpc-5.0.0-1.ohpc.src.rpm  | grep cmake
 ohpc-module(cmake)
 # repoquery --provides cmake-ohpc
 cmake-ohpc = 3.9.2-1.ohpc
 cmake-ohpc(x86-64) = 3.9.2-1.ohpc
 ohpc-module(cmake)
 ohpc-module(cmake/3.9.2)

And if using the macro '%ohpc_load_modules' in the '%build' and
'%install' section the previously defined required modules will
be automatically loaded.

The resulting binary RPM will only require 'ohpc-module(cmake)' if it is
mentioned in the environment modules file.

Right now this has only been compile tested and the results can be seen
here: https://copr.fedorainfracloud.org/coprs/adrian/ohpc-modules-final/

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -34,6 +34,8 @@
 
 DocDir:    %{OHPC_PUB}/doc/contrib
 
+%global lmod_version 7.6.1
+
 %if 0%{?rhel} == 7
 	# OBS define's rhel_version to 700 for RHEL, in case we are
 	# not running in OBS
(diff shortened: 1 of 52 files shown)


adrianreber:
This needs to be rebased on top of recent version changes, but I wanted to submit a PR for today's TSC meeting.

PR #637, opened 2017-12-19
Opafm variable in input.local

PaulUTK:
The input.local files were missing the opafm variable being defined. The recipe.sh script checks for these variables in order to install the opafm package and enable the service. Adding the variable to the input.local file will allow a user to select that parameter now for use.


koomie:
Thanks for the catch and fix.

PR #638, opened 2017-12-20
Recipe.sh lustre-client-ohpc-module does not exist

PaulUTK:
Changing lustre-client-ohpc-modules to kmod-lustre-client-ohpc as the former does not exist anymore.


koomie:
Thanks for this catch. I also realized at some point that the kernel module packaging changed, but ended up modifying the docs to just rely on `lustre-client-ohpc` which pulls in the `kmod-lustre-client-ohpc` package as a dependency.

PR #639, opened 2017-12-20
Recipe nagios perl command checks for /bin/mail instead of /usr/bin/mail

PaulUTK:
Fixing perl script to check for /usr/bin/mail instead of /bin/mail and changes it to /usr/bin/mailx.


adrianreber:
Closing this as the 1.3.x has seen no activity in a long time and 3.x is not using Nagios any more.

PR #642, opened 2018-01-06
Specific uid and gid for slurm system user

mrmhodak:
Slurm requires that uid/gid are the same across the cluster, this patch follows the example of munge spec file (which uses uid/gid 200) and sets uid and gid to 201.

Differing ids are a problem for stateful xCAT install, where ids end up being different - stateful xCAT image installs different packages than default install on the SMS.
 Regardless, given that this is a requirement ( https://slurm.schedmd.com/accounting.html) it is a good practice to enforce this.

### DIFF
--- a/components/rms/munge/SPECS/munge.spec
+++ b/components/rms/munge/SPECS/munge.spec
@@ -146,10 +146,10 @@ rm -rf "$RPM_BUILD_ROOT"
 # karl.w.schulz@intel.com (9/10/18) - provide specific uid/gid to deal with 
 # possibility of getting alternate ownership within Warewulf
 /usr/bin/getent group munge >/dev/null 2>&1 || \
-  /usr/sbin/groupadd -r munge -g 200
+  /usr/sbin/groupadd -r munge -g 201
 /usr/bin/getent passwd munge >/dev/null 2>&1 || \
   /usr/sbin/useradd -c "MUNGE authentication service" \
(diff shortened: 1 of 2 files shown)


adrianreber:
I like this as I already hit the problem of different UID/GID when using OpenHPC without any provisioning system. I was using ansible to set up the master and the compute nodes and I just created the slurm user before installing the RPM. But something like this makes it of course even easier.

I am just unsure about the actual UID. Looking at munge 201 makes sense but looking how Fedora/CentOS/RHEL handles static ID distribution I am not sure we should use 200/201:

* https://pagure.io/setup/blob/master/f/uidgid
* https://fedoraproject.org/wiki/Packaging:UsersAndGroups

Maybe this is something for the future to use higher IDs from system account ID range:

```
$ cat /etc/login.defs
# System accounts
SYS_UID_MIN               201
SYS_UID_MAX               999
```

mrmhodak:
Good point. I checked Suse and their system uid/guids start from 100. So we should go with 201 and 202 for munge and slurm, respectively. I will update the PR.

There is also a possibility that the 200 number was chosen to avoid conflict with other packages (note that if uid/gid of 200 already exist, the user/group creation fails), but using numbers outside the expected range is not a good solution.

mrmhodak:
Updated. Munge has uid/guid of 201, slurm 202. This should work for both CentOS and Suse.

mrmhodak:
This has been resubmitted as #673 to satisfy new submission requirements.

PR #643, opened 2018-01-08
xCAT stateful guide for 1.3.4

mrmhodak:
Adding xCAT stateful Guide. 

Compute nodes are installed with BOS as soon as xCAT is installed. OpenHPC components are added to both SMS and nodes when nodes are up and running.  

Also, minor updates to existing xCAT documentation to use xCAT commands where possible.


koomie:
Thank you for updating this against the latest revision. I'm getting ready to work on starting to land this - can I make one request though?  We have just recently adopted a model to enforce signed requests to acknowledge a Certificate of Origin. Additional information is available here: https://github.com/openhpc/ohpc/wiki/Contributions

Assuming you agree, can you update the PR to include the 'Signed-off-by' option. 

Thanks.

mrmhodak:
I feel stupid asking this, but how to do it retroactively? 

I agree with the terms, but presumably, all the 9 commits have be changed, and I am not sure how.

koomie:
Let's see if the suggestion mentioned https://gist.github.com/kwk/d70f20d17b18c4f3296d might work?  You would have to change the commit count in the example from 2 to 9, and add the correct email, etc, but that should get your local fork updated. 

If the local commit logs look good after that, I think you will have to do a forced push to your fork (which I believe will update the PR as well).  In theory, that would look something like:

```
git push -f -u origin mrmhodak:OpenHPC_1.3.4_add_doc
```

mrmhodak:
That wants a "ref" to rewrite. I love how git can make simple things hard.

koomie:
Sorry for the difficulty here.  Perhaps another, simpler option might be to just create a new pull request? You could create a new local branch off of `openhpc:obs/OpenHPC_1.3.4_Factory` and then copy in the files you modified.  Then, `git commit -s`, push to your fork, and open another pull request.

mrmhodak:
That is what going to do

mrmhodak:
Closing, superseded by #672 .

PR #650, opened 2018-01-16
admin/ohpc-filesystem: modified approach for creation of ohpc.attr fi…

koomie:
…le. Leverage OHPC_HOME macro at build time in ohpc-filesystem as opposed to using potentially undefined macros in .attr file. Follow up to #645.

Signed-off-by: Karl W. Schulz <karl.w.schulz@intel.com>


adrianreber:
👍 I was also thinking about doing it this way.

Issue #631, opened 2017-12-08
PBS Pro (v14.1.2)


scottaltair:
https://github.com/pbspro/pbspro

Altair has made available PBS Professional v14.1.2 (open source) which addresses a number of defects and a few minor features. They are in the process of updating pbspro.org, but they provided us early information.

koomie:
Tests passing in CI.

PR #660, opened 2018-01-26
#631 Updating pbspro.spec for 14.1.2 release and then some

scottaltair:
- Updating pbspro.spec to refer to 14.1.2 (latest open source version)
- Cleaned up variable name consistencies (pname vs pbs_name, etc)
- Altair move the location of the packages; change to Source0
- Altair agreed to create 'make dist' tar.gz, which OpenHPC will be relying on. No requirement to run autogen.sh

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -91,7 +91,7 @@ Requires:      llvm-compilers%{PROJ_DELIM}
 
 %endif
 
-# Disable RPM symlink analysis on files in %{OHPC_HOME}. 
+# Disable RPM symlink analysis on files in %%{OHPC_HOME}. 
 %global __libsymlink_exclude_path  %{OHPC_HOME}/.*$
 
 # MPI dependencies
--- a/components/compiler-families/gnu-compilers/SPECS/gnu-compilers.spec
+++ b/components/compiler-families/gnu-compilers/SPECS/gnu-compilers.spec
@@ -11,7 +11,7 @@
 %include %{_sourcedir}/OHPC_macros
 
 %if "%{compiler_family}" == "gnu7"
-%global gnu_version 7.2.0
+%global gnu_version 7.3.0
 %global gnu_major_ver 7
 %global gnu_release 1
 %global pname gnu7-compilers
--- a/components/dev-tools/numpy/SPECS/python-numpy.spec
+++ b/components/dev-tools/numpy/SPECS/python-numpy.spec
@@ -24,6 +24,7 @@ Requires:      openblas-%{compiler_family}%{PROJ_DELIM}
 %if 0%{?sles_version} || 0%{?suse_version}
 %define python_module() python-%{**} python3-%{**}
 %else
+%define __python3 /usr/bin/python3.4
 %define python_module() python-%{**} python34-%{**}
 %endif
 Name:           python-%{pname}-%{compiler_family}%{PROJ_DELIM}
--- a/components/distro-packages/python-rpm-macros/SOURCES/specfile_search.patch
+++ b/components/distro-packages/python-rpm-macros/SOURCES/specfile_search.patch
@@ -1,12 +1,24 @@
---- a/functions.lua	2017-12-07 11:46:04.907000000 -0800
-+++ b/functions.lua	2017-12-18 20:29:50.982000000 -0800
-@@ -56,10 +56,19 @@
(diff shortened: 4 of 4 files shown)


adrianreber:
@scottaltair : Question about the pbspro spec file. It prints out a lot of text during installation, which seems unnecessary as OpenHPC provides the documentation in its recipes. This seems to be the only OpenHPC package which prints this kind of information during installation and I think it could be removed.

scottaltair:
@adrianreber sorry, I don't believe I am following your concern about the pbspro.spec file. Can you provide me with specific example?

adrianreber:
So actually this is unrelated to this PR. So sorry for hijacking it but I was always thinking if this is necessary. I am talking about these messages:

%postun -n %{pname}-%{pbs_server}%{PROJ_DELIM}
echo
echo "NOTE: /etc/pbs.conf and the PBS_HOME directory must be deleted manually"
echo

%postun -n %{pname}-%{pbs_execution}%{PROJ_DELIM}
echo
echo "NOTE: /etc/pbs.conf and the PBS_HOME directory must be deleted manually"
echo

%postun -n %{pname}-%{pbs_client}%{PROJ_DELIM}
echo
echo "NOTE: /etc/pbs.conf must be deleted manually"
echo

mike0042:
@adrianreber : The spec file is directly responsible for printing one message after PBS Pro has been uninstalled. The remainder of the messages come from a script called pbs_postinstall that is part of the PBS Pro package itself. We could redirect stdout/stderr from pbs_postinstall to /dev/null or an installation log file.

docs/recipes/install/common/pxse_enabled_builds.tex
```diff
@@ -46,8 +46,13 @@
 % ohpc_command if [[ ${enable_opa} -eq 1 ]];then
 % ohpc_indent 10
 \begin{lstlisting}[language=bash,keywords={},upquote=true,keepspaces]
+<<<<<<< .merge_file_CHumzE
+# Optionally, choose Omni-Path enabled MPI builds. Otherwise, skip to retain default MPI stacks
+[sms](*\#*) (*\install*) openmpi-psm2-intel-ohpc mvapich2-psm2-intel-ohpc
+=======
 # Optionally, choose the Omni-Path enabled build for MVAPICH2. Otherwise, skip to retain IB variant
 [sms](*\#*) (*\install*) mvapich2-psm2-intel-ohpc
+>>>>>>> .merge_file_juf65L
```
adrianreber: there is merge conflict marker in this file

adrianreber:
@mike0042 Thanks. I was thinking about many more messages. Right now it seems so little. But thanks for pointing out it comes from a script.

Right now I am still asking if this is necessary. Maybe someone else from the community can also comment if they think it is useful or if it is not required.

docs/recipes/install/common/pxse_enabled_builds.tex
scottaltair: Sorry, @adrianreber . I must not have had the latest changes in my local branch when I created the PR. This had nothing to do with what I was working on. I will need to rebase next time.

mike0042:
@adrianreber : I added those messages to the PBS Pro spec file when we went open source. The spec file was later adapted to work with OpenHPC. I felt it was important to inform the admin there were files that the package left behind. In the case they weren't planning to install another version of PBS Pro, they could manually delete them.

adrianreber:
@mike0042 I personally think that it is not necessary as we have a lot of packages leaving old files around during uninstall. I think that there is not much value in printing information during install/uninstall of RPM packages as those messages might never be read.

Anyway, this PR is the wrong place to discuss this as it is unrelated to the PR. I should probably open a new issue, but it was more an observation.

There is still the merge conflict marker in the documentation which needs to be cleaned up before this PR can be merged.

mike0042:
@scottaltair : You'll want to remove the three %postun blocks in the spec file and redirect stdout/stderr to /dev/null in the three places where pbs_postinstall is called. That should silence all messages. Rebasing your branch should git rid of the merge conflict that was pointed out.
@adrianreber : Please reference the new ticket from this PR once you open it. Thanks!

scottaltair:
This PR is no longer valid. Will be opening a new PR. Attempted to rebase and clean up, but made it worse.

PR #664, opened 2018-01-29
Updating pbspro.spec for v14.1.2 and cleaning up misc

scottaltair:
- Updating pbspro.spec to refer to 14.1.2 (latest open source version)
 - Cleaned up variable name consistencies (pname vs pbs_name, etc)
 - Altair move the location of the packages; change to Source0
 - Altair agreed to create 'make dist' tar.gz, which OpenHPC will be relying on. No requirement to run autogen.sh
 - Reduced verbosity by commenting out %postun and redirecting pbs_postinstall to /devnull
 - Addressing the OBS Badness values. Nothing wrong with build.


scottaltair:
@adrianreber and @mike0042 - I have closed out the previous PR. Here is the new one. See the description above that addresses the comments in the old PR (#631 )

adrianreber:
Thanks for the reduction in verbosity.

Issue #668, opened 2018-02-07
PAPI (v5.6.0)


koomie:
http://icl.cs.utk.edu/papi/software

koomie:
Tests passing in CI.

PR #669, opened 2018-02-07
perf-tools/papi: bump version to v5.6.0 (#668)

koomie:


koomie:
Testing DCO check. Did not like me switching emails on the commit so going to redo.

PR #672, opened 2018-02-10
xCAT stateful guide for 1.3.4

mrmhodak:
Adding xCAT stateful Guide.

Compute nodes are installed with BOS as soon as xCAT is installed. OpenHPC components are added to both SMS and nodes when nodes are up and running.

Also, minor updates to existing xCAT documentation to use xCAT commands where possible.

This replaces PR#643.

Signed-off-by: mhodak <mhodak@lenovo.com>


mrmhodak:
Please note that it would be good to merge #642 at the same time, otherwise slurm ends with different user ids on the nodes compared to SMS. I will update that PR to include the required signed-off message.

PR #678, opened 2018-02-21
macros for OHPC-style builds of multiple python flavored packages from a single spec

crbaird:
This PR adds the notion of a python family similar to our compiler and mpi families to allow a single spec file to produce multiple rpms for both python2.7 and python3.4. It defaults to python3, and a _link file in OBS can be used to use python2 as an alternative.

Signed-off-by: Reese Baird <reese.baird@intel.com>

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -35,9 +35,9 @@
 DocDir:    %{OHPC_PUB}/doc/contrib
 
 %if 0%{?rhel} == 7
-	# OBS define's rhel_version to 700 for RHEL, in case we are
-	# not running in OBS
-	%global rhel_version 700
+    # OBS define's rhel_version to 700 for RHEL, in case we are
+    # not running in OBS
+    %global rhel_version 700
 %endif
 
 # OpenHPC packages require ohpc-filesystem which defines the basic
@@ -59,6 +59,7 @@ BuildRequires: ohpc-buildroot
 
 %{!?compiler_family: %global compiler_family gnu7}
 %{!?mpi_family: %global mpi_family openmpi3}
+%{!?python_family: %global python_family python3}
 
 # Compiler dependencies
 %if 0%{?ohpc_compiler_dependent} == 1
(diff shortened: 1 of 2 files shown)


components/OHPC_macros
```diff
@@ -35,9 +35,9 @@
 DocDir:    %{OHPC_PUB}/doc/contrib
 
 %if 0%{?rhel} == 7
-	# OBS define's rhel_version to 700 for RHEL, in case we are
-	# not running in OBS
-	%global rhel_version 700
+    # OBS define's rhel_version to 700 for RHEL, in case we are
+    # not running in OBS
+    %global rhel_version 700
```
adrianreber: These whitespace changes seem unrelated to the actual python3 change.

components/OHPC_macros
```diff
@@ -115,9 +116,32 @@ Requires:      openmpi3-%{compiler_family}%{PROJ_DELIM}
 %endif
 %endif
 
+# Python dependencies and macros
+%if 0%{?ohpc_python_dependent} == 1
+%if "%{python_family}" == "python3"
+%global __python %__python3
```
adrianreber: Where is %__python3 defined?

components/OHPC_macros
```diff
@@ -115,9 +116,32 @@ Requires:      openmpi3-%{compiler_family}%{PROJ_DELIM}
...
-	. %{OHPC_ADMIN}/ohpc/OHPC_setup_compiler %{compiler_family} \
-	%if 0%{?ohpc_mpi_dependent} == 1 \
-		. %{OHPC_ADMIN}/ohpc/OHPC_setup_mpi %{mpi_family} \
-	%endif \
+    . %{OHPC_ADMIN}/ohpc/OHPC_setup_compiler %{compiler_family} \
+    %if 0%{?ohpc_mpi_dependent} == 1 \
+        . %{OHPC_ADMIN}/ohpc/OHPC_setup_mpi %{mpi_family} \
+    %endif \
 }
+
```
adrianreber: More whitespace changes.

components/OHPC_macros
crbaird: it's in EPEL's python3-rpm-macros, a requirement of python34-devel

adrianreber:
Thanks for the clarification. :+1: 

One idea would be to move the BuildRequires to this file, so that we can just set ohpc_python_dependent to 1 and all BuildRequires and Requires are automatically set like with the compilers and MPI. Just an idea. Maybe even your intel compiler python34-build-patch requirement could be in the central macro file.

Issue #605, opened 2017-11-28
slurm (17.11.5)


crbaird:
Available at: http://slurm.schedmd.com/download.html

crbaird:
tests passing in CI

koomie:
External PMIx linkage has been re-introduced into SLURM build and tests passing in CI.

PR #679, opened 2018-02-23
rms/slurm: require gtk2 to trigger build of sview (#605)

crbaird:
A user on the mailing list requested sview.

Signed-off-by: Reese Baird <reese.baird@intel.com>


koomie:
I don't think most folks are going to want to have to pull in all of gtk2 for a default install. Instead, what I might suggest if we want to include build support for `sview`, is to package that up in a separate sub package (e.g. `slurm-sview`) so that the gtk dependencies can be isolated to an optional package.

adrianreber:
:+1: You could squash those two commits two one commit. I do not think that steps during this review need to be part of the repository.

PR #708, opened 2018-04-25
Add missing zlib-devel runtime dependency to phdf5

chuckatkins:


crbaird:
Thanks, @chuckatkins. Is this causing an issue for you? The automatic dependency generator in rpmbuild should be handling this, and it seems like it is working in our CI testing:

rpm -q --requires phdf5-gnu7-openmpi3-ohpc | grep libz
libz.so.1()(64bit)

chuckatkins:
It is.  The dependency generator will make sure the zlib runtime package gets pulled in but not the dev package, so you can end up with linker errors from missing libz.so (not libz.so.1).  I ran into the issue when building a minimal container for circle CI: https://github.com/ornladios/ADIOS2/blob/master/scripts/circle/el7-gnu7-openmpi-ohpc/Dockerfile .  If I don't explicitly add the zlib-devel package then linker errors "unable to find -lz" show up when building with hdf5.

chuckatkins:
@crbaird ping?

koomie:
Can I ask if you were using the hdf5 generated wrapper scripts (e.g. `h5pcc`) or compiling/linking with the ohpc module env vars?

I built your docker container (after removing zlib-devel and disabling the libfabric build) and can compile hdf5 examples ok using the module provided vars as follows:

```
$ module list

Currently Loaded Modules:
  1) gnu7/7.3.0   2) openmpi3/3.0.0   3) phdf5/1.10.1

$ mpicc $HDF5_DIR/share/hdf5_examples/c/h5_write.c -I$HDF5_INC  -L$HDF5_LIB -lhdf5
```

However, I do get the linkage failure with `h5pcc` and I think the rub is that the wrappers are adding the `-lz` linkage when they really don't need to since it is pulled in dynamically.

``` ldd $HDF5_LIB/libhdf5.so | grep libz
	libz.so.1 => /lib64/libz.so.1 (0x00007f525fbe2000)
```

chuckatkins:
The issue is indeed with the hdf5 compiler wrappers.  CMake searches for h5cc or h5pcc and interogates them to get the necessary include dirs and libraries needed to use with your regular compilers instead of the wrappers.  So while libz.so might not actually be needed, h5cc says it is.  So really adding zlib-devel as a dependency is a workaround for h5cc not properly hiding private depencies.  Dropping private depencies really needs to be addressed in upstream hdf5 but until that happens the h5cc and h5pcc wrappers provided by the ohpc package are broken without the zlib-devel depency.

koomie:
Since we are building to support dynamic linkage only, how would you feel about leaving the zlib dependency as is, but I patch the wrappers (`h5cc`, `h5pc`, `h5c++` and their parallel equivalents) to not include these unneeded private link options?

I'm testing locally and the patches are straight forward since the wrappers are autoconf'ed.

chuckatkins:
Sounds good.  I can look into submitting a fix upstream.  So long as the wrappers work with the appropriate options to interogate the necessary dependencies then it should be fine.

koomie:
Updated approach in #749.  I think a generic fix for upstream is more complicated - we know we want dynamic linkage - but if someone was trying to do a build with static libs then the wrappers would need to retain that logic.

chuckatkins:
Thanks for the fix.  It should be pretty simple to add the "if building static" logic.  Regardless, this works good for here in the meantime.

PR #711, opened 2018-04-26
EasyBuild v3.6.0 + updated bootstrap script

boegel:
updated bootstrap script deals with some changes that were made to PyPI


crbaird:
Thanks as always, @boegel

PR #717, opened 2018-05-11
Add SLURM dependency on hwloc

ktpedre:
This will find the base OS version, not the OpenHPC hwloc version.
If rebuilding SLURM RPMs and the base OS does not have hwloc-devel installed, SLURM will be built without hwloc support. Adding the --with-hwloc forces a dependency on hwloc-devel.

Signed-off-by: Kevin Pedretti ktpedre@sandia.gov


adrianreber:
@ktpedre The text in your first comment should be part of your commit message, especially the Signed-off-by and then the DCO check should not fail.

ktpedre:
Closing, as this was merged in #814

PR #747, opened 2018-05-30
Add GEOPM and msr-safe to OpenHPC

cmcantalupo:
We are planning on tagging GEOPM version 0.5.0 tomorrow or Thursday, and we will update this pull request to point to the new release at that time.  For now we are just looking for any feedback this request which is based on GEOPM version 0.4.0, the latest stable release.


components/perf-tools/geopm/SPECS/geopm.spec
```diff
@@ -0,0 +1,570 @@
...
+#  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+#  A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+#  OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+#  SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+#  LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+#  DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+#  THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+#  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY LOG OF THE USE
+#  OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+#
```
adrianreber: @koomie I am curious about license text in our SPEC files. Isn't OpenHPC content supposed to be Apache 2.0 licensed? Does it make sense to have a different license text in SPEC files? I understand that the actual license of the source code is different, that makes sense. But I think the SPEC file should be Apache 2.0 and without an additional license text in the SPEC file.

components/perf-tools/geopm/SPECS/geopm.spec
```diff
@@ -0,0 +1,570 @@
...
+
+EOF
+
+%files
+%defattr(-,root,root,-)
+%{OHPC_PUB}
+%doc README COPYING VERSION
+
+%changelog
+* Fri Jan 12 2018 Christopher M. Cantalupo <christopher.m.cantalupo@intel.com> v0.4.0
```
adrianreber: @koomie I think we also should have a rule about the %changelog section. Most of our SPEC files have no %changelog at all. I would prefer to have a changelog, but as all is documented in git it is not really necessary.

If we have a changelog it should, however, only contain the changes to the SPEC file and not about the software packaged. This %changelog, from my point of view, is much too long and detailed, especially for a new OpenHPC package.

components/perf-tools/geopm/SPECS/geopm.spec
```diff
@@ -0,0 +1,570 @@
...
+%define docdir %{OHPC_PUB}/doc/contrib/%{pname}-%{compiler_family}-%{mpi_family}-%{version}
+%define install_path  %{OHPC_LIBS}/%{compiler_family}/%{mpi_family}/%{pname}/%version
+
+Name:          %{pname}-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
+Summary:       Global Extensible Open Power Manager
+Version:       0.4.0
+Release:       1
+License:       BSD-3-Clause
+Group:         %{PROJ_NAME}/parallel-libs
+Vendor:        Intel Corporation
```
adrianreber: We usually do not set the Vendor tag

components/perf-tools/geopm/SPECS/geopm.spec
```diff
@@ -0,0 +1,570 @@
...
+URL:           http://geopm.github.io/geopm
+Source0:       https://github.com/geopm/geopm/releases/download/v%{version}/geopm-%{version}.tar.gz
+Source1:       OHPC_macros
+BuildRoot:     %{_tmppath}/%{pname}-%{version}-%{release}-root
+DocDir:        %{OHPC_PUB}/doc/contrib
+Prefix:        %{_prefix}
```
adrianreber: BuildRoot, DocDir, Prefix are unnecessary or set by the global macro file OHPC_macros

components/perf-tools/geopm/SPECS/geopm.spec
```diff
@@ -0,0 +1,570 @@
...
+Requires:      msr-safe
+BuildRequires: autoconf
+BuildRequires: automake
+BuildRequires: hwloc-devel
+BuildRequires: libtool
+BuildRequires: libtool-ltdl-devel
+BuildRequires: ohpc-buildroot
```
adrianreber: Also set by the global macro file

components/perf-tools/msr-safe/SPECS/msr-safe.spec
```diff
@@ -0,0 +1,133 @@
...
+
+export INSTALL_MOD_PATH=$RPM_BUILD_ROOT
+export INSTALL_MOD_DIR=extra/%{name}
+for flavor in %flavors_to_build ; do
+        make -C %{kernel_source $flavor} modules_install \
+                M=$PWD/obj/$flavor
+done
+
+%clean
+rm -rf %{buildroot}
```
adrianreber: clean section are automatically included

components/perf-tools/msr-safe/SPECS/msr-safe.spec
```diff
@@ -0,0 +1,133 @@
...
+# release package (if available) in order to apply patches, perform customized
+# build/install configurations, and supply additional files to support
+# desired integration conventions.
+#
+#----------------------------------------------------------------------------eh-
+
+# msr-safe.spec
+
+%include %{_sourcedir}/OHPC_macros
+%{!?PROJ_DELIM: %global PROJ_DELIM -ohpc}
```
adrianreber: this is provided by the global macro file

components/perf-tools/geopm/SPECS/geopm.spec
cmcantalupo: Agreed, the %changelog in the spec file is intended to track changes to the spec file itself, not the software it package.  Appending the geopm/ChangeLog to the spec file is not appropriate here.  We can start a log for changes to the OpenHPC spec file and remove what is there.

components/perf-tools/msr-safe/SPECS/msr-safe.spec
cmcantalupo: Just to be clear, you are saying we should remove lines 82 and 83.

components/perf-tools/msr-safe/SPECS/msr-safe.spec
cmcantalupo: Remove line 14?

components/perf-tools/msr-safe/SPECS/msr-safe.spec
```diff
@@ -0,0 +1,133 @@
...
+
+# Base package name
+%define pname msr-safe
+%define PNAME MSR-SAFE
+
+
+Name:           msr-safe
```
cmcantalupo: @crbaird The Name: should be:
msr-safe%{PROJ_DELIM}

This also needs to change in the geopm.spec Requires: field.

components/perf-tools/msr-safe/SPECS/msr-safe.spec
adrianreber: Yes, the %clean section is not needed in SPEC files for quite some time now.

components/perf-tools/msr-safe/SPECS/msr-safe.spec
adrianreber: Yes, the PROJ_DELIM macro is set by the global macro file.

components/perf-tools/geopm/SPECS/geopm.spec
bgeltz: Chris and I don't have approval (OSPDT) to release GEOPM under any other license than the BSD-3-clause, so I'm not sure we're empowered to change this if it is deemed necessary.

components/perf-tools/geopm/SPECS/geopm.spec
bgeltz: Done. Will be in next pull request rev.

components/perf-tools/geopm/SPECS/geopm.spec
bgeltz: Done. Will be in next pull request rev.

components/perf-tools/geopm/SPECS/geopm.spec
bgeltz: Done. Will be in next pull request rev.

components/perf-tools/geopm/SPECS/geopm.spec
bgeltz: Done. Will be in next pull request rev.

components/perf-tools/msr-safe/SPECS/msr-safe.spec
bgeltz: Done. Will be in next pull request rev.

components/perf-tools/msr-safe/SPECS/msr-safe.spec
bgeltz: Done. Will be in next pull request rev.

components/perf-tools/msr-safe/SPECS/msr-safe.spec
bgeltz: Done. Will be in next pull request rev.

components/perf-tools/geopm/SPECS/geopm.spec
koomie: I agree, this is a recapitulation of the license for the source release itself which is flagged using normal RPM semantics via the `License:       BSD-3-Clause` line and inclusion of formal `LICENSE` file which ships with the package.  We are not 100% consistent, but the vast majority of the .specs in ohpc do not do this.  Would you be amenable to contributing a spec with the standard ohpc .spec header?

bgeltz: Thanks for the feedback.  I'm going to remove the license header that's there now and replace it with just the standard ohpc header.  I'll have that up in a few minutes.

cmcantalupo:
We believe that this pull request is ready to be merged.  We have tagged and released GEOPM version 0.5.0 (https://github.com/geopm/geopm/releases/tag/v0.5.0), and we have updated the spec file accordingly.  We have also given the test content some exercise in our cluster, but we are not able to check all combinations of tool chain / MPI / resource manager.  

Note that GEOPM has a requirement that the MPI processes are pinned to distinct CPUs (thread migration is not allowed).  The prun script does not enforce this requirement, which is enforced by the GEOPM launcher during normal use, so we forked the script to add the options we require to the command line when an environment variable is not available.

tests/perf-tools/geopm/geopm-prun
```diff
@@ -0,0 +1,273 @@
...
+
+function launch_mpich () {
+
+    if [[ ${RM} == "slurm" ]];then
+	if [[ ${OHPC_MPI_LAUNCHERS} =~ "pmix" ]];then
+            verify_launcher_avail srun
+            cmd="srun --mpi=pmix --cpu-bind=cores $@"
+	else
+	    verify_launcher_avail mpiexec.hydra
+	    cmd="mpiexec.hydra -bind-to core -bootstrap slurm $@"
```
koomie: I understand why you went the route to add some additional command-line options here, but I definitely would not like to maintain two prun variants.  Ideally, it would be nice to just set additional env vars in the test script(s) to get desired binding. I did look at MPICH and see that they have a `HYDRA_BINDING` var which is supposed to accommodate this, but it doesn't look to work in the latest stable release. I did find a patch that will query this var and will include it in the next mpich build for v1.3.5 (https://github.com/openhpc/ohpc/issues/750).  

I think you can do most things in openmpi with env vars as well, but I didn't try this exact case. So, we may not need any mods to prun at all.  If we can't do everything with env vars, I would suggest that as opposed to having a hacked prun separately, that we expand prun to allow you to append additional desired command line arguments and we can commit to that for the 1.3.6 branch if it's necessary.

Long story short, I would suggest you drop `geopm-run` and assume that (1) you can get core binding with env vars or (2) we'll extend the existing prun to allow you add the desired command-line options.  If you are ok with that route and update the PR to remove geopm-prun, I'll land it on the v1.3.6 branch.

bgeltz: Chris and I have been working up a solution today that does not involve forking prun.  I've tested it and it's working as well as the solution currently in this PR.  I'll update the PR with that work in a few minutes.

For reference, the patch is here: c0a383802cdeea066536e7f5276fa1f03d3d5fe3.

bgeltz: Ok, the latest update removes geopm-prun and adds the necessary command line args for core binding through export bash function overrides for mpiexec.hydra, mpirun, and srun.  Let me know what you think on this approach.

koomie: I see this as an improvement but still not crazy about it since it will be sensitive to job launch changes in the future, and that's the motivation for having something like `prun` in the first place. I'll land this on 1.3.6 and we can take the time to fix MPICH and prun if necessary to make the testing process clean and hopefully semi-future proof.  

Thanks for the work on this.

Issue #755, opened 2018-06-14
PBS Pro (v18.1.3)


scottaltair:
Altair released PBS Professional 18.1.2 (https://github.com/PBSPro/pbspro/releases/tag/v18.1.2). PBS Professional 18.1.x a significant release from 14.1.x. The 18.1.2 release addresses upgrade issues from 14.1.0.

scottaltair:
I have successfully built x86_64 and aarch64 with 18.1.2 source on the OBS.

koomie:
tests passing in CI.

PR #756, opened 2018-06-14
Update pbspro.spec to support PBS Professional 18.1.2 and other details

scottaltair:
Addressing #755 
- Updated copyright formatting and year
- Updated pbs_version to 18.1.2
- Removed pbs.null.patch and reference in pbspro.spec
- Removed systemd.patch and references in pbspro.spec - per the upstream pbspro.spec
- Updated pbspro.spec with Cray CLE logic
- Cleaned up preun logic
- Updated pbspro.spec with posttrans to correct unconditional rm of /etc/init.d/pbs


scottaltair:
Missing sign-off, again.

PR #757, opened 2018-06-14
Update pbspro.spec to support PBS Pro 18.1.2 and other details

scottaltair:
Addressing #755
- Updated copyright formatting and year
- Updated pbs_version to 18.1.2
- Removed pbs.null.patch and reference in pbspro.spec
- Removed systemd.patch and references in pbspro.spec - per the upstream pbspro.spec
- Updated pbspro.spec with Cray CLE logic
- Cleaned up preun logic
- Updated pbspro.spec with posttrans to correct unconditional rm of /etc/init.d/pbs

Signed-off-by: Scott Suchyta <scott@altair.com>


components/rms/pbspro/SPECS/pbspro.spec
```diff
@@ -11,55 +11,56 @@
...
+# or contact the Altair Legal Department.
+#
+# Altair’s dual-license business model allows companies, individuals, and 
+# organizations to create proprietary derivative works of PBS Pro and distribute 
+# them - whether embedded or bundled with other software - under a commercial 
+# license agreement.
+#
+# Use of Altair’s trademarks, including but not limited to "PBS™", 
+# "PBS Professional®", and "PBS Pro™" and Altair’s logos is subject to Altair's 
+# trademark licensing policies.
```
adrianreber: As already mentioned in another pull request. Does the SPEC file really have to have this license text?

OpenHPC SPEC files are all, as far as I know, licensed under the Apache 2.0 license. I think this license text should be removed from the SPEC file.

scottaltair: @koomie, please advise.  We are using the Altair supplied spec file with OHPC modes to it.

mike0042: The Altair legal department has instructed us to add the license text to all files we maintain and distribute. If this goes against OpenHPC policy I can follow up with legal. How would you prefer to proceed @adrianreber?

billnitzberg: Hi @mike0042,  I'll give you a call to discuss (and I'll follow up with Altair legal as needed).  Copyrights and licenses depend on which files and where they reside, and I want to ensure we are able to follow the OpenHPC guidelines, which has standardized on Apache 2.0.  Thx!

adrianreber: Actually, no idea. I had a look at http://openhpc.community/about-us/openhpc-charter/ and it says
```
All new inbound code contributions to the Project shall be made under the Apache License Version 2.0```

But it also says many other things and I have no idea what exactly means what.

Is it clear that this is only about the SPEC file? This does not change the license of PBS Pro at all.

From my point of view, it would make most sense if all SPEC files are under the same license. Especially as they are also adapted by other people in the community.

But as I am not the person to decide about this, I leave it to others.

Comment (scottaltair) [reply components/rms/pbspro/SPECS/pbspro.spec]: @billnitzberg , @mike0042 , @adrianreber the SPEC file used based off of what Altair publishes in the source. The difference in the pbspro.spec file used at OpenHPC is that it incorporates the ohpc modifications.

Comment (billnitzberg) [reply components/rms/pbspro/SPECS/pbspro.spec]: Hmmm... I just looked deeper and noticed that the standard PBS Pro copyright and dual-license text are already included in OpenHPC.  There really is no change being proposed here (in the license).  It seems including the PBS Pro OSS license text verbatim was accepted in the past, so it should be fine, unless something has changed.

Comment (koomie) [reply components/rms/pbspro/SPECS/pbspro.spec]: Hi all,

Agreed that the discussion and request here is not about changing the license under which pbs professional is released. That is your business and it was included previously with Affero GPL .

This is about the designation of the build instructions of the .spec file that is part of the ohpc repo. We have not been 100% consistent initially on this, but have been trying to clean up a bit along the way. Standard RPM semantics for identifying the license of the software being built is to leverage the License metadata tag and to include the formal license text as part of the install. That first bit is already covered in the current implementation in the .spec via:

```
License:   AGPLv3 with exceptions
```

I don't think we are currently installing the LICENSE file so regardless, we should probably amend the ohpc .spec to include a line like the following in the %files section

```
%doc COPYRIGHT INSTALL LICENSE NEWS 
```

The point being that with this, the resulting RPM and install correctly identifies the license and you don't need to recapitulate the underlying source license again in the text of the build instructions.

So I guess the question is would ya'll be amenable to leaving that bit out in the .spec file that we use and test in the ohpc repo and just rely on the standard ohpc header?  We have to maintain a separate .spec in ohpc anyway, so you wouldn't necessarily need to alter the .spec included in the pbs release tarball (although adding the `%doc` section is probably a reasonable addition).

billnitzberg: Hi @koomie,

OK, I understand and that seems reasonable.  Since it's a change, let me double check with our legal team to see if we can just delete the whole text or they want to retain the copyright and just delete the GPLv3 bits or something else.  Should hear back in a day or two.  Thx!

scottaltair: @billnitzberg were you able to get an update from the Altair Legal team?

billnitzberg: @scottaltair -- thanks; I've pinged our legal team again.

billnitzberg: Hi @scottaltair & @koomie,

I received approval from Altair Legal to contribute the pbs.spec file (without the PBS Pro OSS license text) to OpenHPC with the understanding that it would then be covered by the Apache 2.0 license.  If there are a couple other files that also need to be contributed (sans PBS Pro OSS license), that will be OK, just let me know which ones.

Thanks!

adrianreber: Thanks for your efforts to get this solved!

mike0042: Please go ahead and update the spec file @koomie. Thanks for checking with us.

PR #761, opened 2018-06-25
Update geopm to 0.5.1

cmcantalupo:
We published a hotfix for a few issues.  This patch updates the spec file to point to version 0.5.1.


crbaird:
thanks, @cmcantalupo

PR #783, opened 2018-06-29
Remove noatime from NFS mount options

jabl:
In commit 67141e1a30912ee852fceca40fc6d6a0b6f20b19 I, among other
things, added noatime to the NFS mount options.  However, it turns out
that the Linux NFS client, in fact, ignores any atime related mount
options, and instead unconditionally implements its own relaxed atime
semantics.  For more detauils, see
http://man7.org/linux/man-pages/man5/nfs.5.html .


jabl:
Oh, what a mess, I'll close this and do a new one...

PR #784, opened 2018-06-29
Remove noatime from NFS mount options

jabl:
In commit 67141e1 I, among other things, added noatime to the NFS
mount options. However, it turns out that the Linux NFS client, in
fact, ignores any atime related mount options, and instead
unconditionally implements its own relaxed atime semantics. For more
details, see http://man7.org/linux/man-pages/man5/nfs.5.html .


koomie:
Thanks - I didn't realize it was a NOOP for NFS mounts.

PR #787, opened 2018-07-09
msr-safe: comment out unknown kernel module macros

adrianreber:
Not sure if this is correct, but those macros are undefined and the
lustre SPEC has them also commented out.

### DIFF
--- a/components/perf-tools/msr-safe/SPECS/msr-safe.spec
+++ b/components/perf-tools/msr-safe/SPECS/msr-safe.spec
@@ -33,14 +33,14 @@ Source5:        OHPC_macros
 DocDir:         %{OHPC_PUB}/doc/contrib
 Prefix:         %{_prefix}
 BuildRoot:      %{_tmppath}/%{name}-%{version}-build
-BuildRequires:  %kernel_module_package_buildreqs
+#BuildRequires:  %kernel_module_package_buildreqs
 BuildRequires:  systemd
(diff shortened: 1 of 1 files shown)


adrianreber:
msr-safe does not build for me outside of OBS. I saw that the two kernel module macros are undefined. In the lustre module it is just commented out. Not sure if this is the right 'fix'.

cmcantalupo:
Are you building on a SLE system, and if so do we need to add a section like:

    %if 0%{?sles_version} || 0%{?suse_version}
    BuildRequires:  %suse_kernel_module_package_buildreqs
    %suse_kernel_module_package default
    %else
    BuildRequires: %kernel_module_package_buildreqs
    %kernel_module_package default
    %endif

https://forums.opensuse.org/showthread.php/487046-rpmbuild-cannot-understand-kernel_module_package-macro

I'm just guessing here since I don't know what OS was causing the problem, and the post was made 5 years ago.

adrianreber:
Actually I am only trying to build the SRPM on Fedora 28.  This has not been a problem until now.

Now I see that lustre-client also does not build on Fedora 28 for the same reason. The reason why it did not matter for lustre-client is that nothing requires lustre-client in contrast to msr-safe which is required by geopm.

I stumbled over this as the script misc/build-order.sh fails now with msr-safe/geopm dependency problems.

So I think this PR can be closed and I have to deal with not building this package on Fedora 28.

PR #791, opened 2018-07-09
Replaces INCLUDE with CPATH in the module files of all libraries providing headers.

hoelzlw:
The module files for nearly all libraries set the wrong environment variable `INCLUDE` for the `include`-directory containing the header files.

The point is, that no compiler actually uses that variable.

Instead, `gcc` and `llvm` for instance use the following three variables:
1. `CPATH`
    Paths given via this variable apply to all supported languages and are appended at the end after all paths given via `-I`.
2. `C_INCLUDE_PATH`
    Paths given via this variable apply to `C` only and are appended at the end after all paths given via `-isystem`, which are processed after all paths given via `-I`.
3. `CPLUS_INCLUDE_PATH`
    This variable is the same as above, but for `C++`.

The question now is: What was the initial intention when that code was written?
- Did you want to bring the headers to a common path which could be included easily?
Because in this case, it is necessary to strip the colons out of `$INCLUDE` and replace them with multiple `-I`.
So people need something like
    ```bash
    INCPATHS="-I ${INCLUDE//:/ -I }"
    ```
    to get a proper command line for `gcc`.
- Since the first point is documented nowhere, I believe the intention was to bring the headers to a common path which the compiler searches by default.
In that case, the current situation is a bug and `INCLUDE` should be replaced by the appropriate of the variables above.

I used `CPATH` as the appropriate variable, so one has not to step down what is C++ and what is C.

I did not change the spec `.../components/io-libs/netcdf-Fortran/SPECS/netcdf-fortran.spec` because it is Fortran and there is no `CPATH`-alternative for that language.
I am not sure, whether I may have overlooked some other libraries which are actually Fortran rather than C/C++.

Also, I did not change the spec `.../components/compiler-families/gnu-compilers/SPECS/gnu-compilers.spec`.
`gcc` knows where its own headers are located, so they should not be part of `CPATH`.
In this case, I think one could delete the line prepending `gcc`'s include directory to `INCLUDE` entirely.
Nobody is ever going to make use of this.

However, this was just the first step.

Note, that there is also a variable called `LIBRARY_PATH` which points to directories with static libraries which would otherwise be added to the linker command via multiple `-L` switches.

My impression of environment modules is, that developers compiling their code on the head node before running it on compute nodes do not have to know that header files and libraries are located at some awkward paths such as `/opt/ohpc/pub/utils/valgrind/3.13.0/lib/valgrind/`.
They should be able to just write `include "..."` in their code and use `-l` switches on their linker commands with the compiler finding the appropriate files on its own.

Right now, developers have to know some specific paths, because for instance the static library `libcoregrind-amd64-linux.a` is only available under the path above and there is no dynamic version of it.

So, is there a specific reason, why `LIBRARY_PATH` is missing?
Right now, only EasyBuild is setting it, but it does not seem to be necessary.
At the same time sionlib, scorep, autotools and valgrind have static libraries without `LIBRARY_PATH` set.
As before, `gcc` has static libraries but does not need the variable.
`gcc` knows, where its libraries are.

Should I change this, too and create a new pull request?

Best regards
Wolfgang

### DIFF
--- a/components/dev-tools/python/SPECS/python.spec
+++ b/components/dev-tools/python/SPECS/python.spec
@@ -293,7 +293,7 @@ set             version         %{version}
 prepend-path    PATH                %{install_path}/bin
 prepend-path    PYTHONPATH          %{install_path}/lib/python2.7:%{install_path}/lib/python2.7/lib-dynload:/usr/lib64/python2.7/site-packages:/usr/lib/python2.7/site-packages
 prepend-path    MANPATH             %{install_path}/share/man
-prepend-path    INCLUDE             %{install_path}/include
+prepend-path    CPATH               %{install_path}/include
 prepend-path    LD_LIBRARY_PATH     %{install_path}/lib
 
 setenv          PYTHONHOME          %{install_path}
--- a/components/io-libs/adios/SPECS/adios.spec
+++ b/components/io-libs/adios/SPECS/adios.spec
@@ -211,7 +211,7 @@ set             version             %{version}
 depends-on phdf5
 
 prepend-path    PATH                %{install_path}/bin
-prepend-path    INCLUDE             %{install_path}/include
+prepend-path    CPATH               %{install_path}/include
 prepend-path    LD_LIBRARY_PATH     %{install_path}/lib
 prepend-path	PYTHONPATH          %{install_path}/python/lib64/python2.7/site-packages
 
--- a/components/io-libs/hdf5/SPECS/hdf5.spec
+++ b/components/io-libs/hdf5/SPECS/hdf5.spec
@@ -124,7 +124,7 @@ module-whatis "%{url}"
 set     version			    %{version}
 
 prepend-path    PATH                %{install_path}/bin
-prepend-path    INCLUDE             %{install_path}/include
+prepend-path    CPATH               %{install_path}/include
 prepend-path	LD_LIBRARY_PATH	    %{install_path}/lib
 
 setenv          %{PNAME}_DIR        %{install_path}
--- a/components/io-libs/netcdf-CXX/SPECS/netcdf-cxx4.spec
+++ b/components/io-libs/netcdf-CXX/SPECS/netcdf-cxx4.spec
@@ -156,7 +156,7 @@ set             version             %{version}
 
 prepend-path    PATH                %{install_path}/bin
 prepend-path    MANPATH             %{install_path}/share/man
-prepend-path    INCLUDE             %{install_path}/include
+prepend-path    CPATH               %{install_path}/include
 prepend-path    LD_LIBRARY_PATH     %{install_path}/lib
 
 setenv          %{PNAME}_DIR        %{install_path}
--- a/components/io-libs/netcdf/SPECS/netcdf.spec
+++ b/components/io-libs/netcdf/SPECS/netcdf.spec
@@ -160,7 +160,7 @@ depends-on phdf5
 
 prepend-path    PATH                %{install_path}/bin
 prepend-path    MANPATH             %{install_path}/share/man
-prepend-path    INCLUDE             %{install_path}/include
+prepend-path    CPATH               %{install_path}/include
 prepend-path    LD_LIBRARY_PATH     %{install_path}/lib
 
 setenv          %{PNAME}_DIR        %{install_path}
--- a/components/io-libs/phdf5/SPECS/hdf5.spec
+++ b/components/io-libs/phdf5/SPECS/hdf5.spec
@@ -131,7 +131,7 @@ module-whatis "%{url}"
 set     version			    %{version}
 
 prepend-path    PATH                %{install_path}/bin
-prepend-path    INCLUDE             %{install_path}/include
+prepend-path    CPATH               %{install_path}/include
 prepend-path	LD_LIBRARY_PATH	    %{install_path}/lib
 
 setenv          %{PNAME}_DIR        %{install_path}
--- a/components/io-libs/pnetcdf/SPECS/pnetcdf.spec
+++ b/components/io-libs/pnetcdf/SPECS/pnetcdf.spec
@@ -100,7 +100,7 @@ set     version                     %{version}
 
 prepend-path    PATH                %{install_path}/bin
 prepend-path    MANPATH             %{install_path}/share/man
-prepend-path    INCLUDE             %{install_path}/include
+prepend-path    CPATH               %{install_path}/include
 prepend-path    PKG_CONFIG_PATH     %{install_path}/lib/pkgconfig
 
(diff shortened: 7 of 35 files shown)


adrianreber:
This is no comment about the content but I wonder why it passed the DCO check.

All commits are expected to carry a 'Signed-off-by:' and I do not see it. @koomie

hoelzlw:
That wondered me too.
I learned about DCO like five minutes ago and amended and force pushed the DCO signed commit right now.

adrianreber:
Ah, I can see it also now. Strange.

hoelzlw:
But you were still right.
The commit still passed the DCO check _before_ I new what DCO is and before I amended the commit.

koomie:
I'm perplexed as well, perhaps a temporary glitch in the matrix? I just created an unsigned PR #792 and the check seemed to behave as I expect.

koomie:
Thanks for bringing this up.  Agree that having the modules set `INCLUDE` is not helping folks much and we don't rely on it for testing (it's likely there based on historical baggage). The high level intent for the resulting modules is really to provide the end user a consistent set of env variables that they can rely on for application compilation and to setup dynamic library search paths. The vars we strive to provide for consistency are of the form FOO_DIR, FOO_BIN, FOO_LIB, and FOO_INC. So, for a version of HDF5, this might map to:

```
HDF5_DIR -> /opt/ohpc/pub/libs/gnu7/hdf5/1.10.2")
HDF5_BIN -> /opt/ohpc/pub/libs/gnu7/hdf5/1.10.2/bin")
HDF5_LIB -> /opt/ohpc/pub/libs/gnu7/hdf5/1.10.2/lib")
HDF5_INC -> /opt/ohpc/pub/libs/gnu7/hdf5/1.10.2/include")
```

With the module loaded, the typical use case is for the user to leverage these vars in their build system (a la things like `-I${FOO_INC}/foo.h`, `-L${FOO_LIB} -lfoo`, etc) and this is what we leverage for integration testing.

Let me bring this up for discussion with the technical steering committee - strikes me we could either remove setting `INCLUDE` at all or amend to set `CPATH` as you suggest which has potential different implications for expected end-user interaction and testing.

koomie:
Just wanted to follow up on this to let you know that we've had several discussions on this topic in recent TSC meetings. Overall, there is a strong sentiment to keep the existing convention of exposing a consistent set of environment variables (`_DIR`, `_INC`, `_LIB`, etc) that users can count on to access headers and libraries in their own build systems and to be very transparent about the paths.

That being said, we looked at ways a site could potentially leverage these settings to optionally augment `CPATH` and `LIBRARY_PATH`, ideally without touching the existing modulefiles.  One of the suggestions was to explore the use of a custom module hook in _Lmod_ that could trigger off of an environment variable to optionally add to `CPATH` and `LIBRARY_PATH`.  It turns out we needed one change in _Lmod_ to support this and it is doable with the most recent 7.8.1 release (and we'll use that for the next ohpc release).  

To highlight a working example that could hopefully give you the functionality you were after with the PR, below is code that could be added to the site-wide _Lmod_ config which defines load/unload hooks that will scan for `_INC` and `_LIB` style env vars during the module load/unload process and use the values to include in `CPATH`/`LIBRARY_PATH` respectively.  In this example, the code  is added to:

/opt/ohpc/admin/lmod/lmod/libexec/SitePackage.lua

```
-- Example load/unload hooks to manage CPATH and LIBRARY_PATH
-- environment variables to paths provided by OpenHPC modulefiles
-- (using _INC and _LIB env variable conventions).

local hook = require("Hook")

function ohpc_cpath_set(t)

	 if (os.getenv("OHPC_MOD_SET_CPATH") == "1" ) then

	 	 local modFile = io.open(t.fn,"r")

		 local includeDir=""
	 	 local libDir=""

	 	 for line in modFile:lines() do

	     	     includeDir=string.match(line,"^setenv%s*%w*_INC%s*(%S*)")
		     if includeDir ~= nil then
		   	prepend_path("CPATH",includeDir)
		     end		
		     
	     	     libDir=string.match(line,"^setenv%s*%w*_LIB%s*(%S*)")
		     if libDir ~= nil then
		   	prepend_path("LIBRARY_PATH",libDir)
		     end		
	    	 end

		 modFile:close()
	end
end

function ohpc_cpath_unset(t)

	 if (os.getenv("OHPC_MOD_SET_CPATH") == "1" ) then

	 	 local modFile = io.open(t.fn,"r")

	 	 local includeDir=""
		 local libDir=""

	 	 for line in modFile:lines() do
	     	     includeDir=string.match(line,"^setenv%s*%w*_INC%s*(%S*)")
		     if includeDir ~= nil then
			remove_path("CPATH",includeDir)
		     end		
		     
	     	     libDir=string.match(line,"^setenv%s*%w*_LIB%s*(%S*)")
		     if libDir ~= nil then
		   	remove_path("LIBRARY_PATH",libDir)
		     end		
	    	 end
		 modFile:close()
	end
end

hook.register("load",  ohpc_cpath_set)
hook.register("unload",ohpc_cpath_unset)
```

Now, let's see an example. With this modification, we can set `OHPC_MOD_SET_CPATH=1` in order to trigger the extra logic in the modulehooks for existing ohpc-style modulefiles. 

#### Set env variable to opt-in to set CPATH and LIBRARY_PATH

```
$ export OHPC_MOD_SET_CPATH=1
```
##### NOTE: nothing set initially
```
$ echo $CPATH $LIBRARY_PATH

```

##### Now, load openblas

```
$ module load openblas
$ echo $CPATH $LIBRARY_PATH
/opt/ohpc/pub/libs/gnu7/openblas/0.2.20/include /opt/ohpc/pub/libs/gnu7/openblas/0.2.20/lib
```

##### Unload - back to being empty

```
$ module unload openblas
$ echo $CPATH $LIBRARY_PATH

```

Should you want to play with this approach locally, there is a newer lmod-ohpc version build in the ohpc factory. URL for x86/centos is:

http://build.openhpc.community/OpenHPC:/1.3:/Update6:/Factory/CentOS_7/x86_64/

hoelzlw:
This is nice and elegant, and actually obsoletes the PR.
I don't speak Lua, so I didn't thought of fiddling around with lmod.

I think I can close the PR now.

Regards

aflyhorse:
> Should you want to play with this approach locally, there is a newer lmod-ohpc version build in the ohpc factory. URL for x86/centos is:
> 
> http://build.openhpc.community/OpenHPC:/1.3:/Update6:/Factory/CentOS_7/x86_64/

@koomie I was trying to create a PR before read your post. That's an elegant solution, but did it came to live? I'm using mod-ohpc 8.5.22 from ohpc 2.4 repo but seems this snippet was not included in the factory version. I still need to manually add this script locally.

PR #794, opened 2018-07-11
SPEC file cleanup and gcc-8 builds fixes

adrianreber:
This started as a cleanup attempt to remove unneeded SPEC file tags and attributes.

While testing it I found some gcc-8 errors which I fixed on the way. For me everything builds now except adios. The adios problem is related to python. I tried to fix it but I am not sure if it compiles with python3.

Let me know if I should this PR in two PRs.

One cleanup PR and one fix gcc-8 PR. As my build fixes depend partly on my cleanups it would be more difficult, but I could remove the build fix commits from my branch and once the cleanup is hopefully merged I could open a second PR. Just wanted to show that I have tested it.


components/OHPC_macros
```diff
@@ -34,6 +34,11 @@
 
 DocDir:    %{OHPC_PUB}/doc/contrib
 
+# Instead of having Source[1-9]: OHPC_macros in every SPEC file,
+# this sets Source999 for all SPEC files which include OHPC_macros.
+# Setting it to 999 and hoping it does not conflict with anything.
+Source999: OHPC_macros
+
```
koomie: If we did this, I wonder if it would be worthwhile to cook up a special PR check that scans .spec files to confirm they didn't also define `Source999`?  Or should it be Source42 to ensure consistency with life, the universe, and everything.

adrianreber: Interesting. 42 :smile: .  The chances that Source42 will be reached is also pretty small.

A first idea how to do that would be to use travis to do quick check on the SPEC file. We could also check for other things like %changelog. If you want I can look at a travis integration.

koomie: At first I was thinking of standing up a separate web validator, but travis probably makes more sense. Would be awesome if you'd want to have a look at leveraging travis for some additional checks.

adrianreber: I will set up travis with my clone of ohpc and will send you a link to have a look at it once it works.

PR #801, opened 2018-07-12
Small change to spack default package install path

ECoulter:
Just a couple of changes to the default Spack configuration, so that admins will be able to install user-accessible software without having to make changes to the spack config. 

## Addition of the spack\_install\_path variable
Allows for Spack to create packages in a publicly-accessible
location by default. Currently 'spack install foo' results 
in foo being installed under `/opt/ohpc/admin/...`, which is 
inaccessible to users.
Modified in the %install section via
```
sed -i "s@  install_tree:.*@  install_tree: %{spack_install_path}@" %{buildroot}%{install_path}/etc/spack/defaults/config.yaml
```
(sed takes any delimiter - '@' doesn't conflict with path specification)

## Modifications to the default spack module install path
Installing modulefiles under default OHPC tree for now, with a spack
prefix - something like the 
below may be more appropriate, however, to differentiate between those
built by spack, and those provided by OpenHPC.
Putting spack modules in a custom location would also require modification of the 
spack module, to add those to the modulepath.
```
sed -i "s@    tcl:.*@  tcl: %{spack_install_path}/modules/@" %{buildroot}%{install_path}/etc/spack/defaults/config.yaml
```

Just modifying tcl path for now; lmod is not enabled by default in spack currently...
 discussion about default module system ongoing in the spack project.
 (https://github.com/spack/spack/pull/5955)

Please let me know if this should've been raised as an issue first! (This time, at least, I've signed off on the commit properly...)


components/dev-tools/spack/SPECS/spack.spec
```diff
@@ -41,6 +41,7 @@ Requires: python2-mock
 DocDir:    %{OHPC_PUB}/doc/contrib
 
 %global install_path %{OHPC_ADMIN}/%{pname}/%version
+%define spack_install_path %{OHPC_PUB}/%{pname}/%version
```
adrianreber: Please use %global and not %define, in this case %define works, but we try to go with %global everywhere

components/dev-tools/spack/SPECS/spack.spec
```diff
@@ -84,3 +89,5 @@ rm -rf $RPM_BUILD_ROOT
 %defattr(-,root,root,-)
 %{OHPC_HOME}
 %doc LICENSE README.md
+
+%changelog
```
adrianreber: We just removed the %changelog section from all SPEC files. Please do not add it back.

components/dev-tools/spack/SPECS/spack.spec
ECoulter: Done - thank for the heads up! Didn't know define could lead to weird behaviour.

components/dev-tools/spack/SPECS/spack.spec
ECoulter: Oops - artifact from forking before the cleanup got pushed - fixed!

components/dev-tools/spack/SPECS/spack.spec
adrianreber: If it really leads to weird behaviour is heavily discussed. But %define has a different scope than %global and to avoid ambiguities, sticking to %global is easier.

adrianreber:
Could you squash your commits and force push, just that we do not introduce commits changing things back and forth.

adrianreber:
Thanks for all your quick fixes!

PR #803, opened 2018-07-13
travis: configure travis

adrianreber:
Our initial travis test is a small python script which only looks at the
SPEC files of the pull request and checks if they include any lines we
do not need. Currently this includes:

 * '^Source999.*$' (That is reserved for OHPC_macros)
 * '^Source.*OHPC_macros$' (That is not necessary anymore)
 * '^%changelog.*' (We removed all changelogs)
 * '^DocDir.*' (Set by OHPC_macros)
 * '^BuildRoot.*' (Not necessary anymore)
 * '^%defattr\(-,root,root\).*' (Setting the default value is not necessary)

With this initial travis setup running we can easily include more and
more checks in the future.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+sudo: required
+language: c
+env:
+  - TEST=check_spec
+services:
+  - docker
+script:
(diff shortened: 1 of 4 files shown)


koomie:
This is awesome, thanks for getting it going.

koomie:
FYI we should be up and running with this travis config now for future PR's. I spot checked a few things with #805, seemed groovy to me.

adrianreber:
:+1:

PR #804, opened 2018-07-13
OHPC_macros: set disttag to current branch name

adrianreber:
This sets the disttag to the current branch name. This replicates OBS's
behavior to append the current release string to the release.

Currently this is ".ohpc.1.3.6"


adrianreber:
@koomie: would this be something worth carrying and upgrading it with each new branch, so that outside of OBS built packages also get the correct disttag?

koomie:
Users rebuilding from .src RPMs will have the correct disttag directly, but we can update this per branch along the way for consistency for folks starting straight from git repo.

Issue #774, opened 2018-06-28
FFTW (3.3.8)


crbaird:
Available at: http://www.fftw.org/download.html

koomie:
Tests passing in CI.

PR #806, opened 2018-07-16
Update fftw to 3.3.8

adrianreber:
Updated fftw to 3.3.8 (#774) and a few small SPEC file cleanups.

### DIFF
--- a/components/parallel-libs/fftw/SPECS/fftw.spec
+++ b/components/parallel-libs/fftw/SPECS/fftw.spec
@@ -30,7 +30,6 @@ Source0:   http://www.fftw.org/fftw-%{version}.tar.gz
 %define mpi           1
 
 BuildRequires:        perl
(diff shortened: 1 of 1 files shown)


adrianreber:
@rengolin Quick question concerning optimization. With this PR we turned on ```--enable-sse2 --enable-avx``` for x86_64.

In the help of fftw's configure I see ```--enable-neon```. Is this something to turn on for OpenHPC's fftw package?

rengolin:
Yes. All Arm64 machines have NEON, so we want that enabled anywhere. Even Arm32 nowadays all have neon, including Rpi2, so not a big deal. The only board without it that you can buy nowadays is Rpi1, but honestly, we're safer ignoring it altogether. :)

adrianreber:
@rengolin thanks, will open a new PR with ```--enable-neon``` for aarch64

PR #807, opened 2018-07-16
travis: add rebuild_srpm travis test

adrianreber:
This adds a second travis test which rebuilds all RPMs in a pull
request, for all SPEC files the pull request includes.

To reduce code duplication the parts of the Dockerfiles which would be
same have been factored out and the actual Dockerfiles are created
before running travis dynamically.

### DIFF
--- a/components/serial-libs/plasma/SPECS/plasma.spec
+++ b/components/serial-libs/plasma/SPECS/plasma.spec
@@ -87,15 +87,19 @@ plasma-installer_%{version}/setup.py              \
 %if %{compiler_family} == gnu8
     --cflags="${RPM_OPT_FLAGS} -fopenmp ${PIC_OPT} -I${OPENBLAS_INC}" \
     --fflags="${RPM_OPT_FLAGS} -fopenmp ${PIC_OPT} -I${OPENBLAS_INC}" \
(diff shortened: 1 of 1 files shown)


adrianreber:
A successful travis run of a test commit changing two spec files can be seen here:

https://travis-ci.org/adrianreber/ohpc/builds/404461556

adrianreber:
Ah, the newly added test fails in this PR. Because the Dockerfile ends in check_spec and that is also my check if there is a spec file in the PR. Maybe I should also check that the file is under components.

adrianreber:
Concerning what to rebuild. We could rebuild only the changed RPMs of a pull request to make sure we get the results in a reasonable time. We could, additionally, add a 'allow_failures' test case which rebuilds all packages starting from the changed package in the dependency chain.

So right now we could just merge this PR and I will fix the SPEC file detection in a follow up PR.

And then I can work on a rebuild-everything-starting-with-this-package-in-the-dependency-chain travis test case with 'allow_failures' just as an additional data point about the pull request.

As travis is x86_64 only (except for software emulated qemu VMs) it will take some time, but not two days.

Using travis does not mean we cannot merge a PR before travis finishes. We can still merge it any time  we want. But if something breaks somewhere we can always look at the travis results later to see if the breakage was already detected by travis.

PR #808, opened 2018-07-16
fftw: also build single and long double precision

adrianreber:
With this change fftw is now built three times. For single, double, and
long double precision. This does not yet enable quad precision as quad
precision does not support building with MPI support.

This commit also enable '--enable-neon' for aarch64 builds.


adrianreber:
I have not tested this on aarch64 and we also talked about additional test cases which this does not include.

Issue #685, opened 2018-03-11
introduce arm-hpc compiler family builds


koomie:
The initial back-end OBS infrastructure changes have been completed. Next, build issues need to be worked through for all of the relevant ohpc components to build successfully with a clang-based compiler.

koomie:
Pushing this to v1.3.7 milestone - the ARM folks indicate they will be releasing an updated HPC compiler suite to better address fortran incompatibility issues and have recommended to hold off with builds until the next compiler release.

koomie:
initial capability now included in installation recipes for aarch64

PR #809, opened 2018-07-18
add compatibility package for arm-hpc compilers

koomie:
work towards arm-hpc family builds (#685)

### DIFF
--- a/components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
+++ b/components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
@@ -24,7 +24,6 @@ BuildArch: aarch64
 #!BuildIgnore: post-build-checks
 
 Requires: lmod%{PROJ_DELIM}
-Provides: %{pname}%{PROJ_DELIM}
 
 %description
 
--- a/components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
+++ b/components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
@@ -1,0 +1,150 @@
+#----------------------------------------------------------------------------bh-
+# This RPM .spec file is part of the OpenHPC project.
+#
+# It may have been modified from the default version supplied by the underlying
+# release package (if available) in order to apply patches, perform customized
+# build/install configurations, and supply additional files to support
+# desired integration conventions.
+#
+#----------------------------------------------------------------------------eh-
+
+%include %{_sourcedir}/OHPC_macros
+%{!?PROJ_DELIM: %global PROJ_DELIM -ohpc}
+
+%define pname arm-compilers-devel
+%define year 2017
+
+Summary:   OpenHPC compatibility package for Arm HPC compiler
+Name:      %{pname}%{PROJ_DELIM}
+Version:   %{year}
+Release:   1
+License:   Apache-2.0
+URL:       https://github.com/openhpc/ohpc
+Group:     %{PROJ_NAME}/compiler-families
+BuildArch: aarch64
+Source1:   OHPC_macros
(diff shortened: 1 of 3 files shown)


components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
```diff
@@ -0,0 +1,159 @@
...
+#
+# It may have been modified from the default version supplied by the underlying
+# release package (if available) in order to apply patches, perform customized
+# build/install configurations, and supply additional files to support
+# desired integration conventions.
+#
+#----------------------------------------------------------------------------eh-
+
+%include %{_sourcedir}/OHPC_macros
+%{!?PROJ_DELIM: %global PROJ_DELIM -ohpc}
```
adrianreber: PROJ_DELIM is part of OHPC_macros, that should be part of one of the travis tests

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
```diff
@@ -0,0 +1,159 @@
...
+
+%define pname arm-compilers-devel
```
adrianreber: I prefer using %global to avoid scope issues

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
```diff
@@ -0,0 +1,159 @@
...
+License:   Apache-2.0
+URL:       https://github.com/openhpc/ohpc
+Group:     %{PROJ_NAME}/compiler-families
+BuildArch: aarch64
+
+#!BuildIgnore: brp-check-suse
+#!BuildIgnore: post-build-checks
+
+Requires: lmod%{PROJ_DELIM}
+Provides: %{pname}%{PROJ_DELIM}
```
adrianreber: This looks unnecessary. The is the same as Name: and usually provided by RPM automatically.

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
```diff
@@ -0,0 +1,159 @@
...
+
+%description
+
+Provides OpenHPC-style module compatability for use with the Arm HPC compiler suite.
+
+%prep
+
+%build
+
```
adrianreber: Never actually tried, but instead of defining empty %prep and %build could this bis just removed?

adrianreber:
I would also recommend squashing the commits where you fix errors. No need to introduce commits with errors to have them fixed in the same pull request. Excited to see the arm-hpc compiler. Cool.

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
koomie: I added this check; also revamped slightly to also check for a few items that we do want to be in each spec file.

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
koomie: fixed.

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
koomie: fair enough - i removed these two empty sections.

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
koomie: I feel I had to do this for a specific reason, but can't recall the details. I'll pull that out for now.

PR #815, opened 2018-07-30
Use of recipe.sh without ipmitool or with version 1.5

erbap:
Signed-off-by: Eric Baplue <eric.baplue@gmail.com>


adrianreber:
Sorry that nobody reacted on this PR. I am closing it now. If you still want to  see this change in OpenHPC please open a PR against the 3.x branch and this time we will act much faster on it.

PR #820, opened 2018-08-10
Update pdsh project URL

jabl:


koomie:
Thanks for updated URL. Note that the value in the .spec file is the authoritative source for us - the other files get created and updated automatically as part of the release process (ie. those manifest tables are machine generated by querying the final release RPMs). 

Long story short should the URL need to be updated again in the future, it can just be done in the .spec file, but we can land this since those manifest files will all get updated for 1.3.6 anyway.

Thanks.

PR #821, opened 2018-08-13
Update config files per GEOPM requirements.

bgeltz:
- Disable intel_pstate kernel driver via the kernel cmdline.
- Enable SLURM's task/affinity plugin for per rank CPU pinning.

Signed-off-by: Brad Geltz <brad.geltz@intel.com>


crbaird:
thanks, @bgeltz

PR #822, opened 2018-08-14
Arm HPC Compiler and Arm Performance Libraries

pawosm-arm:


components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
```diff
@@ -73,7 +82,11 @@ if [ -z "${versions}" ]; then
     exit 1
 fi
 
-%post
+%build
```
adrianreber: No need to introduce an empty %build section. In @koomie's previous pull request I especially asked to have it removed. If it is empty, you can just leave it away.

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
```diff
@@ -24,17 +25,25 @@ BuildArch: aarch64
 #!BuildIgnore: post-build-checks
 
 Requires: lmod%{PROJ_DELIM}
+Provides: %{pname}%{PROJ_DELIM}
```
adrianreber: Provides: %{pname}%{PROJ_DELIM} should not be necessary to be listed explicitly.

components/compiler-families/arm-compilers-devel/SPECS/arm-compilers-devel.spec
```diff
@@ -24,17 +25,25 @@ BuildArch: aarch64
...
+Provides: libompstub.so()(64bit)
+Provides: libomp.so()(64bit)
+Provides: libomp.so(VERSION)(64bit)
+Provides: libarmpl.so()(64bit)
+Provides: libarmpl_lp64.so()(64bit)
+Provides: libarmpl_ilp64.so()(64bit)
+Provides: libarmpl_mp.so()(64bit)
+Provides: libarmpl_lp64_mp.so()(64bit)
+Provides: libarmpl_ilp64_mp.so()(64bit)
+AutoReq:  no
```
adrianreber: @koomie : Do these 'Provides' need the (ohpc) coloring?

pawosm-arm: After spending hours on trial-and-error I came up with a list that fulfills expected dependencies

koomie: I would suggest that we take an alternative approach to take to deal with the inconsistency of these vendor rpms. Fundamentally, the issue is that these .so's (and others) are provided by the ARM HPC compiler rpms, but the rpms themselves have disabled the `Provides` metadata ostensibly breaking any downstream entity trying to provide RPMs with dependencies. To highlight the issue, note that compiler rpm does own **libflang.so** as follows:

```
# rpm -qlp arm-hpc-compiler-18.3_Generic-AArch64_RHEL-7_aarch64-linux-18.3-24.aarch64.rpm | grep libflang.so
warning: arm-hpc-compiler-18.3_Generic-AArch64_RHEL-7_aarch64-linux-18.3-24.aarch64.rpm: Header V4 RSA/SHA1 Signature, key ID d0a6194c: NOKEY
/opt/arm/arm-hpc-compiler-18.3_Generic-AArch64_RHEL-7_aarch64-linux/lib/libflang.so
```

However, when you query the associated provides for said rpm, all of the .so's it includes are missing:

```
# rpm -qp --provides arm-hpc-compiler-18.3_Generic-AArch64_RHEL-7_aarch64-linux-18.3-24.aarch64.rpm 
warning: arm-hpc-compiler-18.3_Generic-AArch64_RHEL-7_aarch64-linux-18.3-24.aarch64.rpm: Header V4 RSA/SHA1 Signature, key ID d0a6194c: NOKEY
arm-hpc-compiler-18.3_Generic-AArch64_RHEL-7_aarch64-linux = 18.3-24
arm-hpc-compiler-18.3_Generic-AArch64_RHEL-7_aarch64-linux(aarch-64) = 18.3-24
```

So, you are then stuck trying to hack around this by either faking the necessary .so's in a package that doesn't own them (your approach) or disabling all .so dependency tracking, or something in between (which I will suggest instead).  

Note: this really is an issue with the arm hpc rpms packaging and it is one of the issues I pointed a while ago and you get to see the pain points directly now.  It's nice to ship RPMs, but disabling proper dependencies means nobody can really use them as intended to provide additional rpms that leverage them without hackery.

With that in mind, I would suggest removing these fake provides for .so's provided by the arm hpc compilers, and augment OHPC_MACROS instead to tell the RPM elf dependency analysis mechanism to skip them when building downstream ohpc packages. It might look something like:

```
%if "%{compiler_family}" == "arm"
BuildRequires: arm-compilers-devel%{PROJ_DELIM}
Requires:      arm-compilers-devel%{PROJ_DELIM}
%global __requires_exclude ^libflang\\.so.*$|^libflangrti\\.so.*$|^libompstub\\.so.*|^libomp\\.so.*$
%endif

pawosm-arm: As you can see, I had to add two versions of libstdc++.so.6 to the list: GLIBCXX_3.4.20 and GLIBCXX_3.4.21. It seems scary adding `|^libstdc++.so.*$` to the list above...
If you don't mind, I'd rather stick to my current solution until we fix our RPM's in (one of) upcoming releases.

adrianreber:
Closing as the 1.3.x branch has seen no activity in a long time. For 3.x we had a successful build of almost all packages against the ARM compiler.

PR #844, opened 2018-09-06
CMake: Bump to 3.12.1 and use more system libraries as dependencies

chuckatkins:
3.12 addressed an important bug introduced into 3.11 that causes the Intel C compiler to not be able to properly use -std=gnu11 and instead always use -std=c11, which forces ANSI C compatibility and breaks POSIX functionality.

### DIFF
--- a/components/dev-tools/cmake/LICENSE
+++ b/components/dev-tools/cmake/LICENSE
@@ -1,5 +1,5 @@
 CMake - Cross Platform Makefile Generator
-Copyright 2000-2017 Kitware, Inc. and Contributors
+Copyright 2000-2018 Kitware, Inc. and Contributors
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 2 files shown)


adrianreber:
@chuckatkins could you open the pull request against 1.3.6? OpenHPC branching is a bit confusing, but the current devel branch (1.3.6 right now) is where the newest code lives. Once 1.3.6 is released it is merged back into master.

chuckatkins:
@adrianreber I think this should be good now.

crbaird:
thanks, @chuckatkins

koomie:
@chuckatkins, have you built this latest version with SLES12? The CentOS build went fine but seeing an error in SLES, example log is below.

https://build.openhpc.community/package/live_build_log/OpenHPC:1.3:Update6:Factory/cmake/SLE_12/x86_64

```
[ 75%] Building CXX object Source/CMakeFiles/ccmake.dir/CursesDialog/cmCursesOptionsWidget.cxx.o
[ 1010s] In file included from /home/abuild/rpmbuild/BUILD/cmake-3.12.1/Source/CursesDialog/cmCursesOptionsWidget.h:8:0,
[ 1010s]                  from /home/abuild/rpmbuild/BUILD/cmake-3.12.1/Source/CursesDialog/cmCursesOptionsWidget.cxx:3:
[ 1010s] /home/abuild/rpmbuild/BUILD/cmake-3.12.1/Source/CursesDialog/cmCursesStandardIncludes.h:18:18: fatal error: form.h: No such file or directory
```

It's erring on `form.h`; I thought perhaps the build might be missing ncurses-devel, but that is present in the build. I also see that the cmake src has it's own `form.h` so I'm not sure if this error is in reference to an included source, or to one that should be resolved externally. Just curious if you might have a thought before diving in further.

crbaird:
The CentOS ncurses-devel puts a form.h in a default include path while the SLE ncurses-devel does not. This is now patch in OHPC 1.3.6 Factory. Cmake should really check for the header file at config time.

PR #853, opened 2018-10-02
Add spank plugin to msr-safe to enable save/restore for jobs.

cmcantalupo:
- Include patch from msr-safe mainline that came in before
  the spank plugin as well.


crbaird:
thanks, @cmcantalupo

PR #857, opened 2018-10-10
Check that intel_pstate driver is disabled on the compute node.

cmcantalupo:


adrianreber:
Slightly adapted and applied in 3.x

PR #869, opened 2018-11-03
Explicitly install msr-safe package.

cmcantalupo:
- The msr-safe-ohpc package is a dependency of the msr-safe-slurm-ohpc
  which is already part of the GEOPM install instructions.
- Therefore, this modification is not a functional change to which
  packages will be installed if the instructions are followed.
- Explictly listing the msr-safe-ohpc package will avoid a problem
  with using the msr-safe kernel module in the case where a user
  chooses not to intall the slurm plugin package.


koomie:
Had to cherry-pick this patch but it has landed on 1.3.9 branch (52e76e977073e1e72ccb0be4ead64c743fc1ba6e)

PR #870, opened 2018-11-05
perf-tools/paraver: update build to use system gcc; gui is not a comp…

koomie:
…iler/mpi

variant package

Signed-off-by: Karl W. Schulz <karl@ices.utexas.edu>


koomie:
Note: Travis check seemed to be confused as commit also re-named a .spec file. Will need to update this at some point to deal with this possibility.

PR #874, opened 2018-11-25
SPEC file cleanups

adrianreber:
This mainly removes conditionals from SPEC files for distributions (EL5, EL6, Fedora) which OpenHPC does not support. It also cleans up trailing whitespace and commented out lines.

PR #877, opened 2018-12-11
Add support to build packages on RHEL8 beta

adrianreber:
This includes all commits from my previous cleanup PR (https://github.com/openhpc/ohpc/pull/874). So it would make sense to merge the cleanup PR first.

This is a rebuild of all OpenHPC packages against the RHEL8 beta. I rebuilt all packages against CentOS 7.6 to make sure nothing breaks with the RHEL8 beta changes.

As pbspro is broken with CentOS 7.6 (https://github.com/PBSPro/pbspro/pull/910) this PR also includes a fix for pbspro.

Some OpenHPC packages failed to rebuild against the RHEL8 beta. The most problems are related to python2.

I skipped following OpenHPC packages on the RHEL8 beta because they are part of the OS already in the same version:
 * lua-filesystem
 * lua-bitop
 * luaposix
 * autoconf
 * automake
 * lsof
 * valgrind
 * libtool

Following packages failed to build on the RHEL8 beta because they seem python2 only:
 * singularity (it seems there is a patch in Fedora for python3 builds)
 * pbspro
 * genders
 * easybuild (https://github.com/easybuilders/easybuild-framework/issues/133)
 * spack
 * ganglia
 * mpiP
 * slurm
 * shine
 * geopm
 * tau
 * plasma
 * petsc
 * trilinos
 * adios

Following packages did not build because there is no EPEL8:
 * cmake (not sure why any more)
 * warewulf-ipmi
 * warewulf-provision
 * warewulf-vnfs
 * nagios
 * warewulf-cluster
 * nagios-plugins
 * wxparaver

The following packages are also not building for various other reasons:
 * sigar
 * msr-safe
 * sensys
 * extrae
 * pdsh (needs slurm)
 * mfem (requires petsc)
 * slepc (requires petsc)

Another problem I found during rebuild on the RHEL8 beta is related to annobin, a GCC plugin which
creates annotated binaries (https://fedoraproject.org/wiki/Changes/Annobin).

For all OpenHPC packages I disabled annobin globally in the SPEC files by setting '%undefine _annotated_build'.

If I am now starting to build python packages with the OpenHPC compiler using the default python3 from RHEL8 I will get CFLAGS from python3 which also has the annobin usage hard-coded. The problem right now is that there is no way to disable annobin for any python3 packages built with a compiler which does not support annobin:

https://bugzilla.redhat.com/show_bug.cgi?id=1536569
https://bugzilla.redhat.com/show_bug.cgi?id=1634784

The current work-around is this:

export CFLAGS="-iplugindir=/usr/lib/gcc/x86_64-redhat-linux/8/plugin -fplugin=annobin -fplugin-arg-annobin-disable"

This first changes the plugin search path to the plugin path of the system compiler it then loads the annobin plugin to disable it. So there is no way to disable the plugin without enabling it first.

Unfortunately this means that we are now loading the GCC plugins from the system compiler and not from the OpenHPC compiler. Right not that is not problematic as it is almost the same version, but this could become a problem in the future. I have also not tested in any way how this interacts with non GCC compilers.


adrianreber:
CI failure due to missing group tag in python-Cython.

boegel:
@adrianreber Isn't Python 2.7 also available for RHEL8? EasyBuild will be happy if `python2` is available. Also, Spack supports Python 3 afaik.

adrianreber:
@boegel There might be python2 for some limited time, not sure. If you look at the RHEL 7.5 release notes at:

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/7.5_release_notes/chap-red_hat_enterprise_linux-7.5_release_notes-deprecated_functionality

it says: "Python 2 will be replaced with Python 3 in the next Red Hat Enterprise Linux (RHEL) major release". So I would recommend using python3. I think OpenHPC should try to only use python3 (if possible) if there will be an OpenHPC release for CentOS 8. But as we do not know when CentOS 8 will be available and what it will look like we still have some time. It also not decided if and when OpenHPC will support CentOS 8. This is just my very early effort to prepare OpenHPC to be ready for CentOS 8.

I would try to avoid anything python2 for this effort, but all of this depends on what the TSC thinks about it.

Looking at https://pythonclock.org/ it does not really make sense to work with python2 any more.

adrianreber:
@boegel That page also says "It won’t be in RHEL 8 Beta, but there will come a day when support for Python 2 will end. If you maintain Python 2 code, you should think about porting it to Python 3."

boegel:
@adrianreber Sure, and I fully agree with that, but as far as I can tell, Python 2 will still be available in RHEL8, at least in the first release. Otherwise we would have already received stronger signals, I think...

boegel:
@adrianreber Agreed. I wasn't trying to push for Python 2, let me be clear. I hope to fix the compatibility issues in EasyBuild soon.

adrianreber:
@koomie rebased and force pushed. Now with annobin. Using annobin would make it a lot easier for us, as we can just use the default RPM build flags. I added a bootstrap annobin flag to the annobin SPEC file to be able to build annobin without having annobin installed. So first builds needs the bootstrap flag set to '1' and then it can stay at '0'. For every major gcc release we probably need to do a bootstrap annobin build.

koomie:
I believe we can close this out now that we are building against 8.1 proper, yes?

PR #879, opened 2018-12-11
Update ohpc-find-requires

LaHaine:
fix error printing in case no search path was provided

Signed-off-by: Götz Waschk <goetz.waschk@desy.de>


adrianreber:
The commit is from `LaHaine <LaHaine@users.noreply.github.com>` but signed off by `Signed-off-by: Götz Waschk goetz.waschk@desy.de`, that is why the DCO check fails. Can you fix that? Besides that it looks correct.

Issue #943, opened 2019-03-09
Make munge log to syslog


jabl:
By default munge writes logs by itself to /var/long/munge/munged.log. This prevents those logs from being forwarded to the log server (e.g. the OpenHPC SMS/admin machine), and in case those logs for one reason or another grows rapidly (e.g. https://github.com/dun/munge/issues/26 ) it can be a problem for diskless nodes (or compute nodes with small partitions where /var/log reside).

Fortunately, munge has an option to instead log to syslog, --syslog, which would solve this problem. To add this option, there's a couple of options

- Just add --syslog to the ExecStart line in /usr/lib/systemd/system/munge.service

- The munge package has /etc/sysconfig/munge but that isn't actually used anymore since the systemd service file doesn't specify to load environment variables from there. To do this, one could add --syslog to a default DAEMON_OPTS, and add a "EnvironmentFile=-/etc/sysconfig/munge" directive to /usr/lib/systemd/system/munge.service. Apparently upstream is planning something like this in https://github.com/dun/munge/issues/64

koomie:
This seems look a reasonable idea to me to have it go to syslog. Want to try creating a PR for this change to the ohpc build of munge?

jabl:
Sure, lets see...

koomie:
Thanks! With your patch and slight tweak to apply it, we should now have munge dumping to syslog.  Example from CI environment:

```
# grep munge /var/log/messages 
Mar 12 22:54:57 c2 munged[8659]: Running on "c2.localdomain" (172.16.1.18)
Mar 12 22:54:57 c2 munged[8659]: PRNG seeded with 128 bytes from "/dev/urandom"
Mar 12 22:54:57 c2 munged[8659]: Enabled PRNG entropy pool enhanced stirring
Mar 12 22:54:57 c2 munged[8659]: Updating supplementary group mapping every 3600 seconds
Mar 12 22:54:57 c2 munged[8659]: Enabled supplementary group mtime check of "/etc/group"
Mar 12 22:54:57 c2 munged[8659]: Starting munge-0.5.13 daemon (pid 8659)
Mar 12 22:54:57 c2 munged[8659]: Created 2 work threads
Mar 12 22:54:57 c2 munged[8659]: Found 4 users with supplementary groups in 0.001 seconds
Mar 12 22:54:58 c1 munged[8664]: Running on "c1.localdomain" (172.16.1.17)
[...]

PR #945, opened 2019-03-11
Make munged log via syslog

jabl:
By default munge writes logs by itself to
/var/long/munge/munged.log. This prevents those logs from being
forwarded to the log server (e.g. the OpenHPC SMS/admin machine), and
in case those logs for one reason or another grows rapidly (e.g.
https://github.com/dun/munge/issues/26) it can be a problem for
diskless nodes or compute nodes with small partitions for /var/log.

This should fix https://github.com/openhpc/ohpc/issues/943


koomie:
This looks close but you also need to _apply_ said patch that is introduced, i.e. we need `%patch4` later down in the .spec file. I tested this by hand and will add this in.

PR #950, opened 2019-03-25
Upgrade Intel MPI benchmarks to 2019.2

### LINKED ISSUE #946: intel-imb - MPI Benchmarks (v2019.2)
The Intel MPI benchmarks version 2019.2 is available.  Going forward, the release tarballs will have a new naming convention (https://github.com/intel/mpi-benchmarks/issues/15).

amblakem:
Should fix #946

Signed-off-by: Aaron Blakeman <aaron.blakeman@intel.com>


koomie:
There are multiple issues with this PR - I'm guessing you did not try to build before submission.

1. the setup macro fails since the leading directory name is changed.  Looks like new dirname includes `IMB` and an extra `v` in the path.

```
cd mpi-benchmarks-2019.2
/var/tmp/rpm-tmp.zvwJjR: line 35: cd: mpi-benchmarks-2019.2: No such file or directory
```

2. After fixing above, the current Patch1 being applied to allow use of `mpicc` for compilation no longer applies in this latest version.  

```
Patch #1 (imb.cc.patch):
+ /usr/bin/cat /root/rpmbuild/SOURCES/imb.cc.patch
+ /usr/bin/patch -p0 --fuzz=0
can't find file to patch at input line 3
Perhaps you used the wrong -p or --strip option?
The text leading up to this was:
--------------------------
```

3. After disabling patch, build fails since  the `src` directory is no longer present.

```
+ cd src
/var/tmp/rpm-tmp.X6CTfJ: line 36: cd: src: No such file or directory
```

I've stopped here - not sure what all else is problematic but this needs work. Also, looking at the upstream GitHub for this newer release I see at least one other issue with the updated build system that is introduced in 2019.x. In particular, the Makefile (https://github.com/intel/mpi-benchmarks/blob/master/Makefile) has hard-coded top-level targets that are calling `make -j` with parallel options directly within it's own Makefile. This is not going to behave as expected for standard packaging which can call the top-level with parallel options as well.

You probably want to get that addressed first so that users (and packagers) can call top-level build with parallel options. Typical mechanism for RPM is `make %{?_smp_mflags}`

ablakeman:
Thank you for the feedback.  I did build this locally, but it looks like I made a mistake in my process.  I'm sorry for that.  I'll address the issues that I can, and work with the upstream author regarding the hard-coded uses of make.

amblakem:
Thank you for the feedback. I did build this locally, but it looks like I made a mistake in my process. I'm sorry for that. I'll address the issues that I can, and work with the upstream author regarding the hard-coded uses of make.

amblakem:
Again, sorry about my mistake with the initial PR.  I see where my process broke down and I'll try not to do that again.  With this latest commit I've manually tested against a couple compiler\MPI variants.  I tried to use other packages for reference when doing these updates.  

I don't have a good way for testing against all compiler\MPI combinations.  If you have any specific guidance on how to do that I can make sure that future submissions are more thoroughly tested.

amblakem:
> In particular, the Makefile (https://github.com/intel/mpi-benchmarks/blob/master/Makefile) has hard-coded top-level targets that are calling make -j with parallel options directly within it's own Makefile. 

I apologize for taking so long to follow up on this point.  I'm told that this is fixed now, but this fix won't be available until the next IMB release later in the year.  If you'd like, we can close this PR and #946 and I can resubmit when this fix is in place.

koomie:
I'd suggest letting the upstream folks fix the release first as it will be beneficial for all potential users. I will close this PR and leave #946 open so we can revisit when there is a new release.

amblakem:
That does sound like the best path forward.  Thank you @koomie

Issue #968, opened 2019-04-22
Charliecloud (v0.15)


koomie:
https://github.com/hpc/charliecloud

j-ogas:
Greetings. We maintain a charliecloud spec file and would be happy to submit it for this issue.

koomie:
There is a .spec file with the ohpc naming conventions as well (e.g. https://github.com/openhpc/ohpc/blob/obs/OpenHPC_1.3.8_Factory/components/runtimes/charliecloud/SPECS/charliecloud.spec) - certainly welcome updates to this variant (and to include relevant changes from your upstream version).

crbaird:
We also manually pre-build the man pages and source them in. The sphinx dependencies were just too much across CentOS/SLE and x86/ARM.

j-ogas:
> The sphinx dependencies were just too much across CentOS/SLE and x86/ARM.

Agreed. As of 0.9.9, we include pre-built man pages in the source tarball, e.g., `https://github.com/hpc/%{name}/releases/download/v%{version}/%{name}-%{version}.tar.gz`. Updating `Source0` with this new path should help reduce the spec file complexity.

To further simplify packaging, starting in 0.9.10, we will also provide pre-built html pages in the source URL above.

koomie:
bumped to v0.15

koomie:
Tests passing in CI.

PR #990, opened 2019-04-29
add charliecloud 0.9.9

j-ogas:
Addresses #968.

This pull request introduces changes from 0.9.9 that reduce packaging complexity.

The goal is to collaborate with you folks to include relevant changes from our upstream version now and more closely in future.

A couple notes:

-  **patches removed** — 0.9.9 accounts for aarch64 and includes pre-built man pages in the source tarball (obtained from `source0`). In 0.9.10, the tarball will include pre-built html files. For the time being, we've been packaging a readme with the URL to the html documentation online.  
- **added -test subpackage**— we feel like the test suite is an important part of the charliecloud package. If possible, we'd like to include it going forward. The rpmlintrc has been updated accordingly.


components/runtimes/charliecloud/SPECS/charliecloud.spec
```diff
@@ -13,56 +13,61 @@
 # Base package name
 %define pname charliecloud
 
+# libexec path
+%define _libexecdir %{_prefix}/libexec
+
```
adrianreber: I am bit unsure about this. Most files are installed in /opt/ohpc. This moves some files under /usr.

We have a few packages which are installed under /usr, but this is probably something new. Some files under /opt/ohpc some under /usr. If some other charliecloud package (without the -ohpc suffix) would be installed this files would probably collide.

components/runtimes/charliecloud/SPECS/charliecloud.spec
```diff
@@ -99,8 +104,45 @@ EOF
...
+Charliecloud comes with a fairly comprehensive Bats test suite. For testing
+instructions visit: https://hpc.github.io/charliecloud/test.html
+EOF
+
 %{__mkdir_p} ${RPM_BUILD_ROOT}/%{_docdir}
 
+%check
+# Don't try to compile python files with /usr/bin/python
+%{?el7:%global __python %__python3}
+
```
adrianreber: The `%check` section seems useless.

components/runtimes/charliecloud/SPECS/charliecloud.spec
```diff
@@ -99,8 +104,45 @@ EOF
...
+  echo user.max_user_namespaces=3171 >/etc/sysctl.d/51-userns.conf
+  reboot
+
+Note for versions below RHEL7.6, you will also need to enable user namespaces:
+
+  grubby --args=namespace.unpriv_enable=1 --update-kernel=ALL
+  reboot
+EOF
+
```
adrianreber: Are user namespaces working on SLES 12 out of the box?

adrianreber:
Please add a Signed-Off-By, amend your commit and force push to your branch.

components/runtimes/charliecloud/SPECS/charliecloud.spec
j-ogas: Good point. I've addressed this in my local branch by defining libexecdir as `%{install_path}/libexec`.

components/runtimes/charliecloud/SPECS/charliecloud.spec
j-ogas: Removed the `%check` section in my local branch. I've also moved the global python macro to the top. I believe you guys have this covered in `OHPC_macros` but we'd like to keep it.

components/runtimes/charliecloud/SPECS/charliecloud.spec
j-ogas: I believe they do, however, let me check with a member of the cray team here to be certain.

Issue #987, opened 2019-04-23
PLASMA (v18.11.1)


koomie:
https://bitbucket.org/icl/plasma

koomie:
@adrianreber; I know you looked at a newer plasma version before in #919 and there were build issues at the time. Do you want to take another look with their latest v18.11.1 release and see if it might be in better shape?

adrianreber:
@koomie : sure, makes sense as I already had a look at it.

koomie:
Builds complete in OBS.

koomie:
@adrianreber, we are seeing CI failures with this build of the latest plasma version. One gotcha is the lack of an f90 module interface to link against. Looking at the cmake log from the build, and futzing with it some more locally, it's not obvious to me that the cmake setup has any support to enable the fortran interface.

There are some linkage issues with our previous C tests too, but given the fortran issue I wonder if we should continue to hold off on this and keep the previous variant?

adrianreber:
@koomie : reverting sounds right. Upstream still seems not really ready with their changes. I was kind of surprised myself when it actually did build. There does not seem to be much documentation and everything I did required looking at the cmake definitions. Let's use the old version.

koomie:
Ok thanks - I reverted back.

PR #995, opened 2019-05-13
plasma: Update to 18.11.1

adrianreber:
This is real big update. The plasma build-system is completely different
now. This change includes running 'plasmatest' during %check to
verify that this newly build plasma does what it is supposed to do.


adrianreber:
@koomie: plasma has a completely different build system but it seems to work. Do we have some tests using plasma? To verify that this plasma still behaves as the old versions? I was not able to test against the Intel compiler.

Fixes: #987

koomie:
We do have some basic tests included in the integration test suite.   I will land this now and if all goes south in integration testing, we can revisit to decide whether to revert back or not.

PR #996, opened 2019-05-13
lmod: Support OpenSUSE 15.1 and RHEL 8

amblakem:
Hi,

I understand that OpenSUSE 15.1 support is targeted for OpenHPC 2.0 and as such the following packages may need to be updated, but I'm submitting this PR against the current versions to document some of the changes that may be required.  I completely understand if this PR is not merged in favor of an alternate approach.  My initial testing shows that once the changes below are committed and `lmod-ohpc` is installed, about 40% of the OpenHPC packages can be built against OpenSUSE 15.1 without modification.  That number increases to about 50% when packages that rely on /etc/SuSE-release[1] are fixed, but I don't yet have a PR ready for those changes.

I've tested this PR on the current beta OpenSUSE 15.1, SLE12.4, and CentOS 7.

[1] `/etc/SuSE-release` was [deprecated](https://en.opensuse.org/Etc_SuSE-release), which breaks packages that depend on that file existing for detecting a SUSE distribution

### DIFF
--- a/components/admin/lmod/SPECS/lmod.spec
+++ b/components/admin/lmod/SPECS/lmod.spec
@@ -14,11 +14,23 @@
 
 %define pname lmod
 
-%if 0%{?suse_version} <= 1220
-%define luaver 5.1
+%if 0%{?rhel} > 7 || 0%{?suse_version} >= 1500
+%define next_distribution_release 1
 %else
-%define luaver 5.2
+%define next_distribution_release 0
 %endif
+
+%if %{next_distribution_release}
+%define luaver 5.3
+%else
+%if 0%{?suse_version} > 1220 && 0%{?suse_version} < 1500
+%define luaver 5.2
+%else
+%define luaver 5.1
+%endif
+%endif
+
+
 %define lualibdir %{_libdir}/lua/%{luaver}
 %define luapkgdir %{_datadir}/lua/%{luaver}
 %define LUA_CPATH ?.so;?/?.so;%{lualibdir}/?.so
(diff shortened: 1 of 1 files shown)


adrianreber:
As you are mentioning OpenHPC 2.0. I just saw that lua-posix and lua-filesystem can be installed via yum on RHEL8. Do those packages maybe also already exist for SLES 15 (or openSUSE)?

amblakem:
Good point, thank you @adrianreber.  OpenSUSE 15.1 does have `lua5[13]-lua{filesystem,posix}` available but with a newer version than what's currently in OpenHPC.  If you think it would be better to leverage the distribution provided packages, we can close this and I can try and make that work instead.  There is also a `lua-lmod` package available to OpenSUSE 15.1 but I'm assuming that we'd still want the OpenHPC specific package?

I tried to build `lmod-ohpc` against the distribution provided `lua5[13]-{filesystem,posix}`, but this was unsuccessful.  From what I can tell, `lua5[13]-luaposix` doesn't actually provide `/usr/lib64/lua/5.?/posix.so` .  I can look into that a bit more if I should be leveraging the distribution packages instead of building from source.

adrianreber:
For #877 (trying to build OpenHPC on RHEL8) I was using -filesystem and -posix from the OS and it worked. If both target operating systems have the packages and if they work it does not seem to be necessary to provide OpenHPC rebuilds. Not sure what the right approach for OpenHPC is, just was wanted to mention it and I was curious.

koomie:
I believe we did have to provide this previously due to lack of functionality provided in SLES. It would be worth a little investigation to see if we could use distro lua variants on a newer OpenSUE 15.x release, if not we can continue to include necessary dependence in ohpc.

koomie:
@amblakem, do you mind digging a little bit more to see if you can get lmod to build with openSUSE 15.x variant using distro lua packages?

amblakem:
Not at all, I'd be happy to.  I'll report back as soon as I have more information.

koomie:
Thanks - I'd say don't go crazy if it doesn't seem doable; but if can be done, I'd like to not have to carry around the extra lua dependencies in 2.x series.

amblakem:
I was able to use the distribution packages and build `lmod-ohpc` on OpenSUSE 15.1.  So as not to conflict with what @adrianreber has done, I used his work from #877 to create a spec file for `lmod-ohpc` that also works with OpenSUSE [[1]](https://gist.github.com/amblakem/c68341219c9c2de0c21258161b412de5).  With this, I was able to build and install `lmod-ohpc` and use `module` to load `gnu8` and a few libraries.  I also confirmed that this spec file still works on SLE12.  Through my basic testing, everything looks good.

If you'd like me to make a PR against @adrianreber's branch or something, please let me know. I'm not sure the best way to submit these changes.

One thing to note, the `lua5?-lua{filesystem,posix}` packages appear to be in OpenSUSE 15.1 but not in SLE15.1.  I'm not super familiar with the new relationship between OpenSUSE and SLE, so the packages available to OpenSUSE 15.1 may be available to SLE15.1.   The SUSE software page[[2]](https://software.opensuse.org/package/lua53-luaposix ) shows that SLE15 distribution has an "experimental" versions of at least `lua53-luaposix` but I did not try to build against this.  

[1]  https://gist.github.com/amblakem/c68341219c9c2de0c21258161b412de5 
[2] https://software.opensuse.org/package/lua53-luaposix

adrianreber:
@amblakem : about #877 , we have no decision yet when we want to merge this. From my point of view, you could just grab my (lua, lmod) changes and include them in your PR. As your PR changes much less it is much easier to merge than my big change.

amblakem:
Thanks @adrianreber.  I went ahead and removed my previous update that built the lua modules from source.  The PR attached now updates lmod to the latest version and uses the distribution provided lua filesystem and posix modules for OpenSUSE 15.1 and RHEL8.  I tested this on OpenSUSE 15.1 and SLE12 but don't have access to RHEL8 at the moment.  I'll make sure to set up a RHEL8 environment so that I can test this (and future) PRs that target OpenHPC 2.0.

amblakem:
Now that OpenHPC 1.3.8 and OpenSUSE 15.1 are officially released, I think there may be a better approach to updating LMOD than what I did in this PR.  I'm going to go ahead and close out this in favor of work that @jcsiadal is doing, which he should be submitting soon.

Thank you @adrianreber and @koomie for your guidance.

PR #1003, opened 2019-06-10
Fix GEOPM tests

dianarg:


cmcantalupo:
Thanks @crbaird!

koomie:
Has `GEOPM_CTL=pthread` been tested and run successfully with all of the MPI variants? (openmpi, mpich, mvapich2, impi).  It doesn't seem to cooperate with much in our CI environment. 

`GEOPM_CTL=process` seems to at least work with some I tested by hand so will revert to that.

cmcantalupo:
Didn't see this come in right away, but thanks for switching it back to process mode from pthread mode.  I think this is better tested in the context of the OpenHPC test environment, and that seems to reflect your experience.  The OpenHPC tests don't use `geopmlaunch` in favor or supporting the `prun` wrapper and modifying the environment in the test scripts.

PR #1005, opened 2019-06-21
Lmod update for OpenSUSE 15

jcsiadal:
This updates the Lmod component and its required distros to add OpenSUSE 15 support. I included the preliminary changes for RHEL 8.

Summary of changes:
- Since it's used by multiple specfiles, the luaver assignment is moved to OHPC_macros
- Dropped conditionals for SLES before SP3
- Removed the lua-bit component. Nothing supported depends on it anymore.
- Add support for Lua 5.3.

Basic testing done on SLES12SP4, OpenSUSE 15.1, and CentOS 7-1810.

I got rid of conditional conflicts. I can install SUSE-built environment-modules and I assume this should be still be a conflict. Similarly, I can get Modules for CentOS.

SUSE provides Lua packages in its SLE developer language repo. It's possible to eliminate all extra lua packages in favor of those provided by the distro. Just for consideration.

I didn't update Lmod source. It gets updated every 1 or 2 weeks, and occasionally several times in a day. Someone more savvy with its development cycle should pick the right target version.

PR #1006, opened 2019-06-22
Do not check that deleted spec files conform to project guidelines

amblakem:
I noticed with #1005 that Travis was attempting to check a deleted spec file to ensure that it had the required spec file components.  I don't think this is the desired behavior.  This PR should filter out deleted items from being checked if the version of `git` being used is less than 1.8.5.  From what I can tell, travis is using a CentOS 7 docker image which is git version 1.8.3.

```
$ git --version
git version 1.8.3.1
$ grep filter tests/travis/check_spec.py
command = ['git', 'diff', '--diff-filter=ACMRTUXB', '--name-only', sys.argv[1]]
$ tests/travis/check_spec.py 90ebc7959bb698545959a2e1e4c0f8e21211d197...36214738481f5189a665d66e699040d1e2ae6ef7
About to run command git diff --diff-filter=ACMRTUXB --name-only 90ebc7959bb698545959a2e1e4c0f8e21211d197...36214738481f5189a665d66e699040d1e2ae6ef7
Checking that (^Source42.*$|^Source.*OHPC_macros$|^%changelog.*|^DocDir.*|^BuildRoot.*|^%defattr\(-,root,root\).*|^.*%global.*PROJ_DELIM.*$) does not exist

--> Scanning spec file components/admin/lmod/SPECS/lmod.spec

--> Scanning spec file components/distro-packages/lua-filesystem/SPECS/lua-filesystem.spec

--> Scanning spec file components/distro-packages/lua-posix/SPECS/luaposix.spec

No consistency errors found. Exiting.
$ cat /etc/centos-release
CentOS Linux release 7.6.1810 (Core)
```

***edit*** according to #870 this issue also exists when renaming spec files.  From my testing, this PR fixes that issue as well.

### DIFF
--- a/tests/travis/check_spec.py
+++ b/tests/travis/check_spec.py
@@ -35,7 +35,7 @@ if len(sys.argv) != 2:
     print("SKIP. Needs a git range as parameter")
     sys.exit(0)
 
-command = ['git', 'diff', '--name-only', sys.argv[1]]
+command = ['git', 'diff', '--diff-filter=ACMRTUXB', '--name-only', sys.argv[1]]
 
 print("About to run command %s" % ' '.join(command))
 


adrianreber:
Thanks for the quick fix!

PR #1007, opened 2019-06-24
OpenSUSE 15 support for openhpc-base

jcsiadal:
Update to build and install all of the packages in the openhpc-base metafile (examples, losf, pdsh, conman, cmake), and their build and installation dependencies. Lmod was installed using RPM packages created in https://github.com/openhpc/ohpc/pull/1005.

Updates include a few cosmetic changes and a required change to pdsh.

When built independently on OpenSUSE 15.1, pdsh is defaulting to lib64. Not known if this is an issue under OBS; however, the specfile explicitly uses lib, so it makes sense to force the source configurator to also use lib.


jcsiadal:
Moved all commits to PR1023.

PR #1011, opened 2019-07-03
MPI specfile updates to support OpenSUSE Leap 15

jcsiadal:

I'm hitting this issue with several specfiles. Autoconf treats OpenSUSE 15 as a multiarch OS and defaults %libdir to /lib64. OpenHPC specfiles hardcode [PREFIX]/lib in scripts and %files, so the simpler fix is to force ./configure to use the desired directory.

With the OpenMPI build and resulting RPMs working, this allows about a dozen other current packages to build correctly under OpenSUSE Leap 15.1.


jcsiadal:
Added mpich and mvapich. Same fix needed for those specfiles.

jcsiadal:
Merged all commits into PR#1023.

PR #1018, opened 2019-07-30
Use the proper path for the nologin shell

### LINKED ISSUE #1017: fix slurm rpm spec typo
ref :  https://github.com/openhpc/ohpc/blob/v1.3.8.GA/components/rms/slurm/SPECS/slurm.spec#L710
```
  -d %{_sysconfdir} -g slurm -s /bin/nologin -r slurm -u 202
````

I think should be `/sbin/nologin`, not `/bin/nologin`

amblakem:
Should fix #1017.

Signed-off-by: Aaron Blakeman <aaron.blakeman@intel.com>

### DIFF
--- a/components/rms/slurm/SPECS/slurm.spec
+++ b/components/rms/slurm/SPECS/slurm.spec
@@ -707,7 +707,7 @@ mkdir -p $RPM_BUILD_ROOT/%{_docdir}
   /usr/sbin/groupadd -r slurm -g 202
 /usr/bin/getent passwd slurm >/dev/null 2>&1 || \
   /usr/sbin/useradd -c "SLURM resource manager" \
-  -d %{_sysconfdir} -g slurm -s /bin/nologin -r slurm -u 202
+  -d %{_sysconfdir} -g slurm -s /sbin/nologin -r slurm -u 202
   
 exit 0
 


amblakem:
Alongside the `slurm.spec` file is a `spec.latest.patch` file which also has this issue[1].  This patch appears to be out of sync as it's referencing SLURM 17.11.5.  I can't find references to where this patch is used, can it be deleted?

[1] https://github.com/openhpc/ohpc/blob/obs/OpenHPC_1.3.8_Factory/components/rms/slurm/SPECS/spec.latest.patch#L533

koomie:
> Alongside the `slurm.spec` file is a `spec.latest.patch` file which also has this issue[1]. This patch appears to be out of sync as it's referencing SLURM 17.11.5. I can't find references to where this patch is used, can it be deleted?
> 
> [1] https://github.com/openhpc/ohpc/blob/obs/OpenHPC_1.3.8_Factory/components/rms/slurm/SPECS/spec.latest.patch#L533

This patch isn't used formally in the build process, but helps us keep track of changes we have to apply against the upstream .spec file. I've updated it to reflect changes against v18.08.8 and I've landed your 6ed9982 patch on the ohpc 1.3.8.1 branch.

amblakem:
Gotcha, that makes sense.  This is good information, thank you for explaining @koomie.

koomie:
This was landed in 7d240cd9a85be06271ca43d10488faaf1101d222 for v1.3.8.1 and carried forward.

PR #1020, opened 2019-08-01
Add node name patch file for mpich

mrmhodak:
Adding patch file fixing mpich issue where jobs fail if some nodes do not end with a number. From pmodels/mpich#3947


mrmhodak:
Lets wait on merging, this may not be the right patch .

mrmhodak:
After more testing, it turned out that the original patch was not correct. mpich has provided a new patch  - again pmodels/mpich#3947, which works correctly. I have updated the file and now this pull request is ready to be merged.

PR #1021, opened 2019-08-01
Apply patch to fix node name issue in mpich

mrmhodak:
Applies patch from #1020


mrmhodak:
Lets wait on merging, this may not be the right patch .

mrmhodak:
The patch has been updated in #1020, this pull request is ready to be merged

mrmhodak:
Actually, I have a feeling that applying patch can fail at build time. Let me know if that is the case. I had to adjust the patch to apply to 3.3 and I am not sure if the patch file is that or not.

koomie:
yeah, it looks like the patch is built against the v3.3.1 release. There are a few other bug fixes in that release so I'm trying a built with that now.

PR #1023, opened 2019-08-14
Support OpenSUSE LEAP 15.1

jcsiadal:
These changes update the current 2.0 branch to support OpenSUSE LEAP 15.1.

The goal was to get all components building, creating valid RPMs, and installing correctly.

All components build and install except for docs, mrsh, nagios-plugins, test-suite, llvm, and dimemas. An mrsh fix is available in PR#877 (RHEL8 beta).

**Summary of changes:**
- Since it's used by multiple specfiles, the luaver assignment is moved to OHPC_macros
- Added support for Lua 5.3. SUSE provides Lua packages in its SLE developer language repo. It's possible to eliminate all extra lua packages in favor of those provided by the distro. Just for consideration.
- Dropped conditionals for SLES before 12SP3
- Removed the lua-bit component. Nothing supported depends on it anymore.
- Removed conditional conflicts in LMOD. I can install SUSE-built environment-modules and I assume this should be still be a conflict. Similarly, I can get Modules for CentOS.
- OpenBLAS, Lustre client, and OCR were updated to the latest versions.
- In OHPC_macros, Python was updated to 3.6 for RHEL7, RHEL8, and SLE15. Also includes changes from Adrian's update for RHEL8.
- Fixed a serious build bug in pdtoolkit. The configure scripts are missing spaces around some bash conditional operators. This causes that condition to always read true, resulting in the wrong compiler being used.
- Most specfiles needed to fix library installation directory. Autoconf treats OpenSUSE 15 as a multiarch OS and defaults %libdir to /lib64. OpenHPC specfiles hardcode [PREFIX]/lib in scripts and %files, with a few exceptions that use [PREFIX]/lib64. I believe the right fix is to update ./configure and other pre-build scripts to match the hardcoded directory.

This PR replaces PR#1005, PR#1007, PR#1011. These removes any dependency for previous PRs to be merged first.

PR#1010 (Warewulf update) is also required to install the cluster; I left this as a seperate PR since it updates to a development branch.

Basic testing was done on OpenSUSE 15.1, and CentOS 7-1810. Intel components were built using Intel Parallel Studio 2019U4.


jcsiadal:
I've added a patch for Travis. The current regex expression prevents the use of PROJ_DELIM in any context of a global variable, so this fails:
`%global myvar myvalue%PROJ_DELIM`
I don't believe this was the intention. It flagged a line Karl added to lustre.spec two years ago to make sure the rpm use the correct basename.
I updated the regex so that re-definition of PROJ_DELIM is flagged as an error, but the macro can be used as a value when creating other global variables.

jcsiadal:
1. I confirmed the mrsh patch in RHEL8 PR also fixes it in SUSE 15. Nothing in mrsh is modified here.
2. Updates docs and nagios-plugins. These are building now.
3. Add more updates to fix components that were building without error, but placing libs in /lib64.
4. Rebased everything to the latest 2.0 branch.

Based on 1.3.8 RPMs, all of the components use /lib, except for numpy, mpi4py, scipy, gcc, R, opencoarrays, easybuild, and all python extensions.

jcsiadal:
Closing this request. Have synchronized the my fork and have started a new build. Will submit a new PR.

PR #1008, opened 2019-06-27
Update msr-safe to version 1.3.0

bgeltz:
The spank plugin that ships with the previous version (v1.2.1) is broken.

Signed-off-by: Brad Geltz <brad.geltz@intel.com>

PR #1025, opened 2019-08-23
Updates msr-safe to version 1.4.0

cmcantalupo:
Add a patch on top of 1.3.0 from the master branch which fixes save restore for MSRs with open write masks.  This should be applied instead of #1008.  @bgeltz can you take that PR down?  Thanks.

relates to #1008


bgeltz:
Why not update my existing PR?

cmcantalupo:
Branch has been updated to msr-safe version 1.4.0 release which includes the patch that was previously being applied.

PR #1077, opened 2019-10-22
changes for RHEL/CentOS 8

adrianreber:
For some testing on CentOS 8 I am submitting a few of my #877 changes as separate pull requests.


adrianreber:
Pushing more RHEL/CentOS 8 changes to this PR.

adrianreber:
As discussed this are the first few basic packages tested on RHEL 8. With this I can get slurm and gcc9 running using our lmod.

One of the changes to lmod is to not conflict with `environment-modules` from the OS any more, but to use the existing module files installed by the OS. Running `module load os` extends the MODULEPATH to offer the OS installed modules. In our discussions we decided to not display the OS provided modules automatically to not create confusion if there are additional MPI based modules from the OS. If the user, however, explicitly loads the `os` module, these OS based modules can also be used.

Concerning the announced slurm upgrade to 19.x for 2.0. To provide a minimal working set I just made sure 18.x builds and works on RHEL 8, we can still do the update later in the development cycle.

@koomie if possible let's get this merged to have a starting point for 2.0 (and so that I do not have to carry around so many commits locally).

jcsiadal:
Could we eliminate the double-pass build and place a BuildRequires for the OS-provided annobin and use it for bootstrap? I'm assuming here that we ultimately need to provide a newer annobin package than the one in CentOS8.

adrianreber:
> Could we eliminate the double-pass build and place a BuildRequires for the OS-provided annobin and use it for bootstrap? I'm assuming here that we ultimately need to provide a newer annobin package than the one in CentOS8.

Using the annobin from the OS will probably work as long as we stay on GCC 9. Once we switch to 10 this could probably break and we have to do the bootstrap build anyway.

Currently the main reason to have our own annobin package is to make sure we are not relying on the OS annobin as it requires to point our GCC to the OS GCC plugin path which might be a problem once we switch to GCC 10. As we rely on the RPM optflags from the OS which enable annobin we have to provide the plugin in out plugin path. Currently our annobin usage is more trying to avoid a broken compiler than the actual annobin information.

Do you still think we should use the OS annobin for bootstrap? Not sure myself, just trying to not break our compiler? If you have any changes which you think make sense I am happy to include them in this PR.

PR #1087, opened 2019-11-21
LLVM Update to 9.0.0

jcsiadal:
LLVM 9.0.0 compiler suite
Includes: clang, lld, clang-tools-extra, libcxx, libcxxabi, libunwind, compiler-rt, openmp

This is a 2-stage build using GCC8 for bootstrap. After stage 1, the entire suite is rebuilt using only the LLVM/Clang toolchain, to remove dependencies. 

There is no supported Fortran frontend for LLVM9. F18 is not yet included and flang is supported only through LLVM7. Until F18 is official, a previous release of LLVM with flang should remain active in OpenHPC.

OpenHPC tests are updated. I used gfortran as a placeholder for llvm9 and removed llvm4. I don't have a way to pre-test this before a PR. If this doesn't work, the Fortran portion of the test will need a way to conditionally disable it.

The build takes 2-3 hours on a newer dual CPU platform. LLVM is a finicky build and I kept hitting race conditions using parallel cmake. I serialized some of it to avoid this, but it's possible that it may still hit issues on slower systems.

I certainly can't say this is the ideal configuration, but it does build successfully and appears to work. In some cases, configuration choice is out of necessity; for example, when openmp is configured under runtimes, it won't receive cmake variables and won't build. Configuring openmp as a project avoids needing to patch src files. Hopefully, this LLVM config is a good base for further optimizations. I didn't spilt the runtimes into their own package--that may be a good feature-add.

### DIFF
--- a/components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
+++ b/components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
@@ -39,7 +39,7 @@ Source7:   http://releases.llvm.org/%{version}/openmp-%{version}.src.tar.xz
 Source8:   http://releases.llvm.org/%{version}/clang-tools-extra-%{version}.src.tar.xz
 
 BuildRequires: cmake%{PROJ_DELIM}
-BuildRequires: gnu8-compilers%{PROJ_DELIM}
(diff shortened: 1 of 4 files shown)


jcsiadal:
Updated to build using GNU9 for bootstrap.

PR #1091, opened 2019-11-26
Add information about OpenHPC based containers

adrianreber:
(This is still using my private quay.io registry URLs, which can be replaced later with the real location)

During SC19 there were a couple of questions about OpenHPC based containers. Although it is easy to create an OpenHPC based container (see Dockerfiles) it is even more convenient if it already exists.

Details about the containers can be found in the README.md file.

 * `make` # to build all container
 * `make push` # build and push all container to a registry
 * `make check` # run some simple checks using the containers

### DIFF
--- a/containers/README.md
+++ b/containers/README.md
@@ -1,0 +1,77 @@
+# OpenHPC container images
+
+> (Everything defaults to `podman` as it provides the possibility to build
+> containers without running as root. See 
+> https://podman.io/blogs/2019/10/28/podman-with-nfs.html for details.)
+
+The OpenHPC project currently provides the following container images:
+
+ * ohpc-gnu8:1.3.9
+ * ohpc-gnu8-mpich:1.3.9
+ * ohpc-gnu8-mvapich2:1.3.9
+ * ohpc-gnu8-openmpi3:1.3.9
+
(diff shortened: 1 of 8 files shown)


adrianreber:
@koomie @crbaird As discussed in Denver. This is a try to provide OpenHPC based containers.

containers/README.md
```diff
@@ -0,0 +1,77 @@
+# OpenHPC container images
+
+> (Everything defaults to `podman` as it provides the possibility to built
```
boegel: `built` should be `build` here?

adrianreber: Thanks. Fixed and force pushed!

koomie:
I landed this on the 2.0 branch. These will presumably need to get updated to use the `gnu9` or `gnu10` variant depending on what version ships with 2.0.  Now that there is a containers subdirectory, I moved the previous singularity example to reside here as well.

adrianreber:
> I landed this on the 2.0 branch. These will presumably need to get updated to use the `gnu9` or `gnu10` variant depending on what version ships with 2.0. Now that there is a containers subdirectory, I moved the previous singularity example to reside here as well.

@koomie Just let me know whenever I need to adapt/change anything with this.

PR #1094, opened 2019-12-13
Fixed misc/get_source.sh OHPC_macros open error

NaohiroTamura:
This patch fixed misc/get_source.sh "error: Unable to open
../SOURCES/OHPC_macros: No such file or directory" by updating
'_sourcedir' rpm macros, and also updated 'rhel' for OHPC 2.0.

Signed-off-by: Naohiro Tamura <naohirot@jp.fujitsu.com>

### DIFF
--- a/misc/get_source.sh
+++ b/misc/get_source.sh
@@ -21,7 +21,7 @@ for i in `find . -name "${PATTERN}"`; do
 	pushd `dirname ${i}` > /dev/null
 	BASE=`basename ${i}`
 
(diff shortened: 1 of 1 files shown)


adrianreber:
Good idea. That makes it much easier. It requires that the OpenHPC directory does not change from components/GROUP/PACKAGE/SPECS/. So far the layout seems to be the same for all packages.

NaohiroTamura:
@adrianreber @koomie , thanks!

PR #1097, opened 2019-12-17
Update bootstrap/core components (RPM macros, lmod, compiler/mpi setup)

jcsiadal:
Updates the core components needed to launch general build of all OpenSUSE LEAP 15.1 packages.

Tested bootstrap builds on LEAP 15.1, SLES 15, and CentOS 8.

Summary of changes:

- Removed all of the distro Lua packages. There are not needed post-SLES12, as the distros provide them.
- Moved luaver assignment to OHPC_macros. It also checks lua_version first, which is defined under OpenSUSE.
- Added updates for Lua 5.3.
- Updated python macros, using the built-ins instead guessing based on OS. This section could probably be mostly removed if we updated the macro names in the individual specs to use the built-ins (and ditched Python2 support).
- Dropped conditionals for SLES before 12SP3. Kept the SLES12SP4 just in case, but it would make the files easier to read if these were dropped also. Similarly, following new OBS docs and moving sles_version to sle_version.
- Removed conditional conflicts in LMOD. I can install SUSE-built environment-modules and I assume this should be still be a conflict. Similarly, I can get Modules for CentOS.
- Added and then removed the annobin check. There needs to be a better method to bootstrap it.


components/OHPC_macros
```diff
@@ -71,33 +76,31 @@ BuildRequires: ohpc-buildroot
...
-%if "%{compiler_family}" == "gnu7"
-BuildRequires: gnu7-compilers%{PROJ_DELIM} >= 7.2.0
-Requires:      gnu7-compilers%{PROJ_DELIM} >= 7.2.0
-%endif
+BuildRequires: gnu8-compilers%{PROJ_DELIM} >= 8.3.0
+Requires:      gnu8-compilers%{PROJ_DELIM} >= 8.3.0
+%endif # gnu8
+%if "%{compiler_family}" == "gnu9"
+BuildRequires: gnu7-compilers%{PROJ_DELIM} >= 9.2.0
+Requires:      gnu7-compilers%{PROJ_DELIM} >= 9.2.0
```
adrianreber: This looks wrong: gnu7-compilers and 9.2.0

adrianreber:
@jcsiadal Thanks for the cleanups. I like the idea of this PR. I also have been thinking about all the distribution specific macros and I thought to remove all the `_version` macros for rhel/centos and similar. My idea was to just rely on`%{rhel}`. This tells us the version of the CentOS or RHEL release and that's all we need. This way we do not rely on OBS injected macros because I think we do not need them. All we need it to detect if it is CentOS or LEAP and `%{rhel}` should be enough.

jcsiadal:
Thanks for catching the error. Also made me remember to update my build system and add gnu9 to the permutations.

jcsiadal:
Merged changes from the current 2.0 branch. Removed Python2 support entirely. Added OpenMPI4. Updated default families.

jcsiadal:
One more addition: added GNU9 and OpenMPI4 to the admin setup scripts. This is needed to support builds with updated compiler families.

PR #1098, opened 2019-12-18
Update for Intel devel pkgs

jcsiadal:

Current version in 1.3.9 and 2.0 doesn't work correctly. The %{version} tags in command substitution are not escaped, so they're substituted at build time instead of run time.

This update also updates the Intel Parallel Studio XE version to 2020. FYI: compiler version is 19.1.x in PSXE2020.


jcsiadal:
Will resubmit fixes if needed.

PR #1117, opened 2020-01-15
Annobin updates for CentOS and OpenSUSE

jcsiadal:
First, this fixes a typo for ScientificLinux and deletes the extra gnu9 entry in OHPC_macros.

This update implements the annobin package for CentOS. To avoid the chicken/egg problem, there is a flag to disable OpenHPC ohpc-annobin on any package, if needed. This allows gnu9 and ohpc-annobin to build. ohpc-annobin loads gnu9 and performs a bootstrap build; then rebuilds using itself.

I re-enabled annotation on CentOS 8. I've been rebuilding everything and it doesn't appear to have created additional build fails.

I added Requires needed for annobin to build under OpenSUSE. Have not tested that it works as expected, but it builds without error.

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -47,7 +47,7 @@ Source42: OHPC_macros
 # macros defined in each specfile.
 %if 0%{!?rhel_version:1}
 %{?centos_version: %global rhel_version=%{centos_version}}
-%{?scientificlinux_version: %global rhel_version=%{scientific_version}}
+%{?scientificlinux_version: %global rhel_version=%{scientificlinux_version}}
 %if 0%{?rhel} == 7
     # OBS defines rhel_version to 700 for RHEL7
     %global rhel_version 700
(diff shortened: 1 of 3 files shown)


adrianreber:
@jcsiadal Thanks for all your work getting annobin working on both OSes. Looks good. Not sure we really need to care about scientific Linux anymore.

There has been some discussion about dropping annobin support for now as long as nothing really breaks. My initial annobin work was motivated because it was not possible to build python packages without annobin support. I especially did not want to rely on annobin from the OS if we ever update to gcc 10. This seems to be fixed now with this : https://bugzilla.redhat.com/show_bug.cgi?id=1634784

It seems we do not have to deal with annobin right now. We probably need to decide if we just want to remove the whole annobin setup from git to not confuse people. All our work on it would then have been for nothing.

We still need to make sure to disable annobin in OHPC_macros to not enable it when building with the OpenHPC compiler.

koomie:
Yes, I will bring it up for discussion in today's TSC meeting, but I'm in favor of disabling annobin across the board for our builds with the python fix that is in rhel 8.1.

jcsiadal:
Updated this PR. Disables annotation on all OpenHPC builds, but keeps the typo fixes and SPEC update of the annobin package.

jcsiadal:
Closing. Annobin will not be included in 2.0.

PR #1119, opened 2020-01-16
Fix library location for OpenSUSE 15.1

jcsiadal:

Local builds on OpenSUSE are working, but library files will be located under "lib64". On CentOS and SLES12, these libraries are located under "lib". This update updates the configuration so that OpenSUSE 15 builds set the prefix correctly. No impact to CentOS 8 builds.

According to my notes, nothing else is needed to get these files building under OpenSUSE 15.

Also includes a few changes from "sles_version" to "sle_version", since we no longer need to support SLES 11 and earlier.


jcsiadal:
I've put in some other PRs that will conflict with this merge, so I'm going to break this into a pull for each library that still needs it.

PR #1133, opened 2020-02-10
Removed duplicated gnu9

NaohiroTamura:
gnu9 is defined from L.79 to L.82.

Signed-off-by: Naohiro Tamura <naohirot@jp.fujitsu.com>


koomie:
Oops. I think that was caught previously, but I must have messed up resolving conflicts from previous merges.  Thanks for the catch.

NaohiroTamura:
@koomie, thanks for merging!

Issue #1033, opened 2019-09-20
EasyBuild 4.0.0


boegel:
EasyBuild v4.0.0 has just been released.

The flagship feature is added support for running EasyBuild on top of Python 3.5+ (while Python 2.6 and 2.7 are still supported).

As indicated by the major version bump, it includes a couple of backwards-incompatible changes (a somewhat concise overview of the major changes is available at https://easybuild.readthedocs.io/en/latest/EasyBuild4-overview-of-changes.html).
`vsc-base` and `vsc-install` are no longer required, nor is `setuptools`.

I remember discussing with @koomie that it may make sense to only include this with OpenHPC v2.0 because the backwards-incompatible changes.

Any updates on that? Where should I send the PR to update the EasyBuild component to v4.0.0?

amblakem:
Congrats on the 4.0.0 release!

> Any updates on that? Where should I send the PR to update the EasyBuild component to v4.0.0?

I believe the `2.0` branch is the proper target for this.  The branch naming was adjusted because the additional strings embedded in the branch names weren't used[1], but I believe this is the proper target for compatability breaking changes such as this.  Additionally, I think that the `2.0` branch's target distributions are CentOS 8 and OpenSUSE LEAP 15.1 so if you need to make spec updates for those platforms I believe that will be welcomed as well.

[1] Mentioned here: https://github.com/openhpc/ohpc/wiki/files/OpenHPC-TSC-Board-20190703.pdf

boegel:
@amblakem Thanks a lot for the info!

I'll run some tests with EasyBuild 4.0 on CentOS 8 first to make sure things are working as expected before updating the `.spec` and target the `2.0` branch.

Is there any ETA for the OpenHPC 2.0 release?

amblakem:
> Is there any ETA for the OpenHPC 2.0 release?

I believe the target is still to release OpenHPC 2.0 at supercomputing this year, so mid November.

boegel:
@koomie @adrianreber Are there any other examples of Python tools that are packaged for OpenHPC?
Since EasyBuild v4.0.0 it's more straightforward to install EasyBuild like any other Python package (`pip install`), so there no more need for using the bootstrap script...

adrianreber:
[...]

PR #1136, opened 2020-02-14
update to EasyBuild v4.1.1

boegel:
cfr. #1033

I would like to get rid of the use of the bootstrap script, since `pip install EasyBuild` works just fine now, so it should be possible to significantly simplify the `.spec` file, but this will have to do for now...

(WIP because needs proper testing)


adrianreber:
Thanks a lot!

@koomie, I verified it in the 2.0 OBS and this builds on CentOS and LEAP. I also like the extended test case. Would that work in our environment?

PR #1138, opened 2020-02-14
easybuild: a macro needs a '%'

adrianreber:
@boegel @koomie 

There is a single '%' missing and thus the macro is not expanded. Simple fix.


boegel:
Thanks @adrianreber :+1:

PR #1144, opened 2020-02-19
containers: update OpenHPC containers for 2.0 release

adrianreber:
This updates the OpenHPC based container images to use the upcoming release 2.0. As this has not been released yet, the images are versioned using 1.9.9. Using these images it is, however, possible to early try out small parts of the upcoming OpenHPC 2.0 release.

In addition too the previous images there is now also an image using the OpenHPC compiler and EasyBuild. The EasyBuild integration is not yet working correctly, but let's hope it will be soon.

@koomie I would like to use this as a basis for an article where we could point users to try out the upcoming release. The whole EasyBuild integration is also experimental, but maybe a good starting point which we can use to improve it.


chuckatkins:
Rather than explicitly adding things to the `PATH` and `LD_LIBRARY_PATH` with `ENV` in the docker file, would it instead be better to setup a default set of modules to load in `/etc/profile.d`?

adrianreber:
@chuckatkins Thanks for the interest in this work.

> Rather than explicitly adding things to the `PATH` and `LD_LIBRARY_PATH` with `ENV` in the docker file, would it instead be better to setup a default set of modules to load in `/etc/profile.d`?

That was my first attempt, but if you are running a container you are not running a login shell, so all the usual scripts are not parsed. In my tries it was not possible to have any scripts parsed. Which is really strange but I did not figure it out.

I tried to tell the container to run a login shell with something like this `/bin/bash -l -c`. This would work, but if there is any space in the command passed to the container it would need to wrapped in quotes:
```
podman run --rm -v `pwd`/test:/tmp:z quay.io/ohpc/ohpc-gnu9:1.9.9 'gcc -o /tmp/hello /tmp/hello.c'
```

Which would work, but would be unexpected for the user to have all the commands be surrounded by quotes.

The problem I have is that modules are useless in a container. I have not found a way to load the modules in such a way, that the user can still pass whichever command they need into the container. 

Extending the PATH seemed to be most transparent way for a user to use the container without any additional requirements about how the command has to changed from its normal command-line use case.

If you have any solution how to do it better I am happy to include this. I still think it is really strange to not be able to have the shell parse some configuration file, but I tried a few different things and nothing worked.

chuckatkins:
This pr caught my eye because I've been using OpenHPC for a while now for the various CI images we use for https://github.com/ornladios/adios2 and I'm always tinkering with them to try to make them better.    I've essentially come up with a way to do it by

1.  setting up the default module environment in a profile.d script
2.  loading the default environment in another profile.d script
3.  wrapping the commands passed in a login shell with a dummy entrypoint shell script with the shebang `#!/bin/bash -l`

You can see it worked out in a CentOS 7 image with OHPC 1.3.9 and gnu8 + openmpi3 as the modules: https://gist.github.com/chuckatkins/994c8d88fb88f15433b08539162e8ac6

```
$ docker build foo .
...
$ docker run foo module avail

------------------------ /opt/ohpc/pub/moduledeps/gnu8 -------------------------
   openmpi3/3.1.4 (L)

-------------------------- /opt/ohpc/pub/modulefiles ---------------------------
   gnu8/8.3.0 (L)    prun/1.3

  Where:
   L:  Module is loaded

Use "module spider" to find all possible modules.
Use "module keyword key1 key2 ..." to search for all possible modules matching
any of the "keys".


$ docker run foo gcc --version
gcc (GCC) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ 
```

adrianreber:
@chuckatkins Thanks a lot. That looks very promising. I will try to include this in this PR.

chuckatkins:
Thanks! I've faced a similar problem with additional `RUN` steps in the image to build dependencies needed for the ci.  In that case though there's no `ENTRYPOINT` login shell wrapper so I instead manually source the profile to use the modules in the build step for the image, in the case of Intel,
```
RUN curl -L urlforsource | tar -xz && \
    cd srcdir && \
    . /etc/profile && \
    module load intel openmpi3 && \
    export CC=icc CXX=icpc FC=ifort && \
    ./configure ... && \
   ...
```

I'm looking forward to the 2.0 release as 1.3.x has been extremely productive for us.

adrianreber:
@chuckatkins Updated the containers to use your scripts. Works pretty good. Thanks a lot!

@koomie Can you merge this? The containers are available with the `latest` and `2.0.0` tags as described in the README.md and this could be added to the final 2.0 announcement.

On a Fedora 31 system I get the following output using one of these containers:
```
$ podman run --rm quay.io/ohpc/ohpc-gnu9-mpich:2.0.0 module avail
Trying to pull quay.io/ohpc/ohpc-gnu9-mpich:2.0.0...
Getting image source signatures
Copying blob a3ed95caeb02 done  
Copying blob a3ed95caeb02 done  
Copying blob a3ed95caeb02 done  
Copying blob d71c5eab96b2 done  
Copying blob 0b857c1c1b37 done  
Copying blob 43a6eb02a4c1 done  
Copying blob a3ed95caeb02 skipped: already exists  
Copying blob a3ed95caeb02 skipped: already exists  
Copying blob 10ee501cf0ee done  
Copying blob 1fd5ad98a54d done  
Copying blob c14bede4e69a done  
Copying blob 1cad15cc459c done  
Copying blob a5dd42d12c78 done  
Writing manifest to image destination
Storing signatures

------------------------ /opt/ohpc/pub/moduledeps/gnu9 -------------------------
   mpich/3.3.2-ofi (L)

-------------------------- /opt/ohpc/pub/modulefiles ---------------------------
   gnu9/9.2.0 (L)    libfabric/1.10.1 (L)    os    prun/2.0

  Where:
   L:  Module is loaded

Use "module spider" to find all possible modules and extensions.
Use "module keyword key1 key2 ..." to search for all possible modules matching
any of the "keys".

$ podman run --rm quay.io/ohpc/ohpc-gnu9 which gcc
Trying to pull quay.io/ohpc/ohpc-gnu9...
Getting image source signatures
Copying blob 43a6eb02a4c1 skipped: already exists  
Copying blob d71c5eab96b2 skipped: already exists  
Copying blob c14bede4e69a skipped: already exists  
Copying blob 0b857c1c1b37 skipped: already exists  
Copying blob 10ee501cf0ee skipped: already exists  
Copying blob 1fd5ad98a54d skipped: already exists  
Copying blob a3ed95caeb02 done  
Copying blob a3ed95caeb02 done  
Copying blob a3ed95caeb02 done  
Copying blob a3ed95caeb02 done  
Writing manifest to image destination
Storing signatures
/opt/ohpc/pub/compiler/gcc/9.2.0/bin/gcc
```

koomie:
Looks good to me. Thx.

PR #1145, opened 2020-02-21
easybuild integration

adrianreber:
With the help of @boegel I was able to create an integration of the OpenHPC compiler into EasyBuild.

In the end it does not really help a lot as no EasyBuild definition is yet aware of this toolchain and only used if explicitly selected (`--try-toolchain GCC,9.2.0-ohpc-gnu`). It is also not clear how EasyBuild can use the hierarchical module setup of OpenHPC (maybe by combining it into toolchain definitions???), but it is a first step towards a better integration of EasyBuild and OpenHPC.

### DIFF
--- a/components/OHPC_macros
+++ b/components/OHPC_macros
@@ -25,6 +25,9 @@
 %global OHPC_MODULES    %{OHPC_PUB}/modulefiles
 %global OHPC_MODULEDEPS %{OHPC_PUB}/moduledeps
 %global OHPC_MPI_STACKS %{OHPC_PUB}/mpi
+%global OHPC_EASYBUILD  %{OHPC_PUB}/easybuild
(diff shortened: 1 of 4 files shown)


boegel:
Looks good to me after a quick review.

W.r.t. the easyconfig files like the `GCC-9.2.0-ohpc-gnu9.eb` that is now added as a part of the EasyBuild installation: would it make sense to collect those in a separate RPM?

It shouldn't be too hard to write a script that spits out all required easyconfig files based on a template for `GCC`, etc., to populate that RPM...

boegel:
@adrianreber No interest in pursuing this?

adrianreber:
I would prefer to have this closed. This is over a year old and has not been merged and I doubt it will be merged. I am not really familiar with the changes any more. Better to have it closed than keep it forever open.

adrianreber:
Closing this, as it is old and out of date.

PR #1154, opened 2020-03-04
Update metapackages

jcsiadal:

We're internally testing OpenHPC 2.0 on CentOS 8. This update allows us to install the ohpc-(base, base-compute, and warewulf) meta-packages.

Instead of updating this meta-package independently and duplicating work, I modified the macros to grab default values from OHPC_macros.

The losf was updated to bypass a CentOS7 yum requirement.

### DIFF
--- a/components/admin/losf/SPECS/losf.spec
+++ b/components/admin/losf/SPECS/losf.spec
@@ -37,11 +37,13 @@ provides: perl(LosF_rpm_utils)
 provides: perl(LosF_utils)
 provides: perl(LosF_history_utils)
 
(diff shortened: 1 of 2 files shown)


koomie:
I had also updated the metapackages earlier today, but I like what you did to use `compiler_family` directly. I also missed the chrony change - thanks for that.

I will land this but tweak a few things after the fact. I've been doing away with checks of both `sles_version` and `suse_version` as `suse_version` is set on both Leap and SLES.  And, a quick check seems to show that `sle_version` is also set on both. 

Also, slowly want to start using %{rhel} for the rhel check to avoid an OBS'ism of `rhel_version`.

PR #1169, opened 2020-03-19
EasyBuild 4.1.2

boegel:
EasyBuild v4.1.2 includes an important security fix, see https://lists.ugent.be/wws/arc/easybuild/2020-03/msg00030.html for more information.

### DIFF
--- a/components/dev-tools/easybuild/SPECS/easybuild.spec
+++ b/components/dev-tools/easybuild/SPECS/easybuild.spec
@@ -15,7 +15,7 @@
 
 Summary:   Software build and installation framework
 Name:      EasyBuild%{PROJ_DELIM}
-Version:   4.1.1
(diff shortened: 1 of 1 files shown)


adrianreber:
This fails to build because the URL does not work:

https://pypi.io/packages/source/e/easybuild-easyblocks/easybuild-easyblocks-4.1.2.tar.gz

Gives me a 404.

boegel:
Ah, right, I overlooked that part...

There was only a new release of `easybuild-framework` (4.1.2), the other two components `easybuild-easyblocks` and `easybuild-easyconfigs` are still at 4.1.1 (the overall EasyBuild version is determined by the max of the components).

I'll open a follow-up PR to fix this...

boegel:
should be fixed with #1173...

PR #1189, opened 2020-04-13
MPI setup scripts configure correct compiler environment

jcsiadal:
This updates the MPI setup so that compiler environment points to the correct MPI wrapper for the preferred compiler chain. It makes more sense to centralize the logic tree to one place instead of requiring each MPI-based component to repeat the same steps, especially for the Intel MPI Library.

Secondly, it updates the impi environment module.  OpenHPC should not be modifying the installation of Parallel Studio or changing binary file operation. I confirmed with our developer consulting team that setting the environment is correct method to accomplish the end goal. 

You can verify which compiler is called using "mpicc -compile-info".

### DIFF
--- a/components/OHPC_setup_mpi
+++ b/components/OHPC_setup_mpi
@@ -31,16 +31,47 @@ if [ -z "$MODULEPATH" ]; then
 fi
 
 if [ "$OHPC_MPI_FAMILY" = "openmpi" ]; then
+    export ICC=mpicc
+    export CXX=mpicxx
+    export FC=mpifort
+    export F77=mpif77   
     module load openmpi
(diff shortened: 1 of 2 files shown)


koomie:
As mentioned in previous TSC call, this is not what we want. The impi compatibility module does not prevent end users from using wrapper scripts like `mpiicc`, but the beauty is that they do not have to use it. Since all of our builds rely on modules we can continue as is.  Setting env vars like `CXX=mpicxx` unilaterally would also be problematic as some packages build with a mix of MPI and non-MPI and most parallel packages have build detection for MPI wrapper scripts. A subset of these also have no idea about the existence of `mpiicc` and assume that MPI C codes should always be built with `mpicc`.

PR #1194, opened 2020-04-20
Add Magpie 2.3

jcsiadal:
The Magpie 2.3 collection of scripts. I'm sure there are additional optimizations to do and adding user tests is in the works. I thought it would be a good idea to take a look at the current code, make sure it's on-track, and confirm its working in OBS.

Their is no formal build. Instead, the extracted files are cleaned up and then copied directly into the installation directory.

An LMOD file is provided. Java is a minor issue, as JAVA_HOME is required for some components. Since there's no OpenHPC Java standard, I set up the modulefile to detect it dynamically. I had originally done the detect on install, but it would miss Java updates or version switches. Java will need to be installed to each compute node by te cluster admin.


jcsiadal:
Updating to new version. Will submit in a new PR.

PR #1222, opened 2020-06-12
docs: Remove reference to nonexistent package

solj:
I don't claim to know much about this package or why it's not in centos8. I was just going through the documentation to install openhpc and this package doesn't exist anymore.

There is no branch for centos8: https://git.centos.org/rpms/infinipath-psm/branches


solj:
The travisci failure seems unrelated to my changes. Looks like a temporary blip in the build system. Unfortunately, I'm unable to retry the build.

solj:
Also, it looks like the command in section 3.6 may be unnecessary because the package is included in the "Infiniband Support" group.

```
# yum groupinfo "Infiniband Support" | grep opa-basic
   opa-basic-tools
```

PR #1226, opened 2020-06-19
docs: Clean up extra step from c6a3e98c

solj:
This updates the documentation to reflect the upstream CentOS switch to
using systemd for managing the tftp service.


koomie:
Thanks for this fix.

PR #1228, opened 2020-06-23
docs: Add documentation for using NFSv4

solj:
This adds instructions for using NFSv4 instead of NFSv3 for serving
shared network filesystems from the sms.


solj:
I'm not sure this is something you'll actually want to merge but we plan on doing this in our environment so that we can easily configure the firewall for NFS.

adrianreber:
At this point in time most mounts in the documentation are now using `nfsvers=4` in `/etc/fstab`. We had to do the change because version 3 was just not working any more in our tests. I am closing this for now.

Issue #520, opened 2017-08-04
warewulf-provision fails with restrictive umask


crbaird:
https://groups.io/g/OpenHPC-users/topic/centos_7_3_update/4773435?p=,,,20,0,0,0::recentpostdate%2Fsticky,,,20,2,0,4773435

crbaird:
Tests passing

PR #1231, opened 2020-06-27
components: Proper fix for #520

solj:
The fix in 8744b043 doesn't actually resolve the issue with a
restrictive umask. Quoting the [mkpath](https://perldoc.perl.org/5.10.0/File/Path.html) page:

    mode

    The numeric permissions mode to apply to each created directory
    (defaults to 0777), to be modified by the current umask. If the
    directory already exists (and thus does not need to be created),
    the permissions will not be modified.

This means that the mode settings that were applied previously still
respected the restrictive umask and resulted in a nonworking pxeboot
configuration.

This commit modifies the previous fix to use make_path instead since the
`chmod` parameter does precisely what we want.

    chmod => $num

    Takes a numeric mode to apply to each created directory (not modified
    by the current umask). If the directory already exists (and thus
    does not need to be created), the permissions will not be modified.

### DIFF
--- a/components/provisioning/warewulf-provision/SOURCES/warewulf-provision.pxe_file_modes.patch
+++ b/components/provisioning/warewulf-provision/SOURCES/warewulf-provision.pxe_file_modes.patch
@@ -1,28 +1,37 @@
---- a/lib/Warewulf/Bootstrap.pm	2019-06-17 14:01:40.000000000 -0700
-+++ b/lib/Warewulf/Bootstrap.pm	2019-06-25 14:42:31.410866655 -0700
-@@ -210,7 +210,9 @@
-             my $dirname = dirname($file);
+diff --git a/lib/Warewulf/Bootstrap.pm b/lib/Warewulf/Bootstrap.pm
+index ff2ad21..3b4ded0 100644
+--- a/lib/Warewulf/Bootstrap.pm
++++ b/lib/Warewulf/Bootstrap.pm
+@@ -16,6 +16,7 @@ use Warewulf::DataStore;
+ use Warewulf::Util;
+ use File::Basename;
+ use File::Path;
++use File::Path qw(make_path);
+ use Digest::MD5 qw(md5_hex);
+ use POSIX qw(uname);
+ 
+@@ -211,6 +212,9 @@ bootstrap_export()
  
              if (! -d $dirname) {
--                mkpath($dirname);
-+                mkpath($dirname, {
-+                    mode => 0755,
+                 mkpath($dirname);
++                make_path($dirname, {
++                    chmod => 0755,
 +                });
              }
          }
  
-@@ -350,7 +352,9 @@
+@@ -350,7 +354,9 @@ build_local_bootstrap()
              }
  
              mkpath($tmpdir);
 -            mkpath($bootstrapdir);
-+            mkpath($bootstrapdir, {
-+                mode => 0755,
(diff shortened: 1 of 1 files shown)


jmstover:
@solj ... If y'all are hitting a restrictive umask can you add this to upstream as well?

https://github.com/warewulf/warewulf3

Thanks.

components/provisioning/warewulf-provision/SOURCES/warewulf-provision.pxe_file_modes.patch
```diff
@@ -36,35 +45,64 @@
...
-@@ -94,8 +94,11 @@
+diff --git a/lib/Warewulf/Provision/Pxe.pm b/lib/Warewulf/Provision/Pxe.pm
+index c9f6a7b..c843206 100644
+--- a/lib/Warewulf/Provision/Pxe.pm
++++ b/lib/Warewulf/Provision/Pxe.pm
+@@ -19,6 +19,7 @@ use Warewulf::DataStore;
+ use Warewulf::Provision::Tftp;
+ use File::Basename;
+ use File::Path;
++use File::Path qw(make_path);
```
jmstover, line 58: just add `qw(make_path)` to the `File::Path` above?

components/provisioning/warewulf-provision/SOURCES/warewulf-provision.pxe_file_modes.patch
```diff
@@ -36,35 +45,64 @@
...
++            chmod => 0755,
++        });
+     }
+ 
+     foreach my $nodeobj (@nodeobjs) {
+@@ -322,6 +331,7 @@ update()
+                     if (! close IPXE) {
+                         &eprint("Could not write iPXE configuration file: $!\n");
+                     }
++		    chmod 0644, "$statedir/warewulf/ipxe/cfg/$config";
```
jmstover:
```suggestion
+		    chmod(0644, "$statedir/warewulf/ipxe/cfg/$config");
```

This fits closer too where we're using `chmod` in a few other places.

jmstover (commented):
Small nitpicks really...

components/provisioning/warewulf-provision/SOURCES/warewulf-provision.pxe_file_modes.patch
solj: Added your changes. Please forgive my ignorance of Perlisms.

components/provisioning/warewulf-provision/SOURCES/warewulf-provision.pxe_file_modes.patch
solj: Added this to all the chmod incantations.

solj:
> @solj ... If y'all are hitting a restrictive umask can you add this to upstream as well?
> 
> https://github.com/warewulf/warewulf3
> 
> Thanks.

Sure thing. Will put in a pull request shortly.


Thanks!

solj:
It looks like this never got merged and we're still hitting the issue with 2.x. Is there any way to get this included?

adrianreber:
Has your fix been merged upstream?

solj:
@adrianreber yes, but warewulf 3.8.1 was released in 2018 and this was merged in later.

adrianreber:
According to git OpenHPC comes with warewulf 3.9.0. So it should include your patch.

PR #1250, opened 2020-07-29
Jcsiadal boost

koomie:


koomie:
A few fixes on top of #1225

PR #1258, opened 2020-09-15
EasyBuild v4.3.0

boegel:
@koomie Is this still in time for the OpenHPC 2.0 release?

### DIFF
--- a/components/dev-tools/easybuild/SPECS/easybuild.spec
+++ b/components/dev-tools/easybuild/SPECS/easybuild.spec
@@ -15,7 +15,7 @@
 
 Summary:   Software build and installation framework
 Name:      EasyBuild%{PROJ_DELIM}
-Version:   4.2.2
+Version:   4.3.0
 Release:   1%{?dist}
 License:   GPLv2
 Group:     %{PROJ_NAME}/dev-tools


koomie:
Looks like we got a build failure on Leap 15.2...

https://obs.openhpc.community/package/live_build_log/OpenHPC:2.0:Factory/easybuild/Leap_15/x86_64

boegel:
@koomie `ssl.SSLError: [SSL] malloc failure`

I don't see how that can be caused by something specific to EasyBuild, looks like a much more general issue with SSL support in the Python 3.6.10 being used?

koomie:
Well, file this under the auspices of huge time sinks, but I finally got to the bottom of it. It seems that the newer version of `openssl` that is in Leap 15.2 has a change of behavior which assumes that the `getrandom()` system call is always available (and won't revert to using `/dev/urandom` like older versions).  I really should have clued in faster when EasyBuild built ok on aarch64 but not x86_64. The primary difference in our ohpc build environments between those two is the host kernel which supports the OBS chroot build environment (much newer for the arm builds).

Seems a kernel of v3.17 or newer is required to support this system call and our primary x86 build server was just under that threshold. Once I bumped it up, openssl worked ok and latest EasyBuild is compiling on Leap15.2.

boegel:
\o/

Thanks for hunting this down @koomie!

PR #1263, opened 2020-10-01
Update LLVM to 10.

jcsiadal:

Updates LLVM/Clang to version 10. Both distros have current versions of clang available; this build is different because it implements a 2-stage build with bootstrap. This should completely eliminate most GCC libs from the runtime toolchain.

I added a conflict to distro-provided libunwind-devel. The lld linker searches /usr/lib64 before its local llvm lib directory and it breaks if it finds libunwind.so. Not sure where to fix this other than forcing the user to implement command line options for lld, so I added a conflict. SCOREp requires the distro libunwind-devel, but only during build.

Tested on CentOS 8.2 and OpenSUSE 15.2. Built a few test C++ apps with the resulting RPMs. Successful builds on OBS.

Build takes approximately 90 minutes on a fast, dedicated build system.

Other notes:

- Moved to ninja build system; this is what the llvm developers use
- Included more  llvm tools
- Leaving some static libs in place as they seem to be required by clang/lld at runtime
- Fixed the Python ambiguity problem with a global search/replace; not sure if any of the Python 2 stuff is even used, but keeping it to be safe.

### DIFF
--- a/components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
+++ b/components/compiler-families/llvm-compilers/SPECS/llvm-compilers.spec
@@ -8,6 +8,11 @@
 #
 #----------------------------------------------------------------------------eh-
 
+# Simple check for other active LLVM versions; they will break this build
+%if 0%(which clang llvm-ar >/dev/null 2>&1; echo $?) == 0
+%{error: "LLVM found in PATH; cannot build LLVM with with another LLVM available"}
+%endif
+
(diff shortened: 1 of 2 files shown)


koomie:
I don't know why Travis is failing in this case. I suspect it is related to the branch switch under the covers (2.0 -> 2.x). The `check_spec.py` utility seems to run ok for me locally on this branch.

koomie:
Enabled build in OBS for 2.1:Factory

PR #1270, opened 2020-10-19
Update Warewulf initrd to include firmware symlinks

jcsiadal:

New Intel NIC drivers load firmware using a symlink under /lib/firmware pointing to the canonical file in the same directory. The wwbootstrap utility adds only true files to the initrd. This update also adds symlinks matching the bootstrap.conf list.

I plan to also upstream this to WW.


jcsiadal:
Accidently hit the delete button on the wrong branch during cleanup. Restored and reopening.

jcsiadal:
Also checked build on OBS. Looks OK.

PR #1294, opened 2021-02-03
Add Bright Cluster Manager 9.0 OpenHPC recipe

ahillbright:
This PR introduces a recipe for integrating Bright Cluster Manager 9.0/SLURM/CentOS 8 clusters with OpenHPC. The recipe is mainly based off the existing xCAT/Warewulf recipes with modifications to make it work in a Bright cluster.

Signed-off-by: Andrew Hill <andrew.hill@brightcomputing.com>


adrianreber:
@ahillbright Sorry that there hasn't been any feedback on this for over a year. @utdsimmons I think we should bring this to the TSC if we want to include this recipe.

PR #1325, opened 2021-06-11
Remove Singularity from future releases

jcsiadal:
With OpenSUSE 15.3 and the latest EPEL, distros are providing newer versions of Singularity that OpenHPC. I know this topic has come up in the past, but now seems like the right time to defer this package to the distros.

I believe I caught all of the instances related to OpenHPC 2.2+ There is an open question on how to configure testing for Charliecloud, since it's not obvious Singularity is needed to test Charliecloud. I modified the test to check for singularity so the log will show the problem.


koomie:
If we did this, I don't think we would want to remove singularity checks. We still want to have a basic smoke test, but they would need to rely on the distro provided version instead. We still need to validate that the install instructions correctly land a working version.

jcsiadal:
Closing. Replace with #1327.

PR #1327, opened 2021-06-24
Remove Singularity from future releases (updated)

jcsiadal:
Replaces #1325.

With OpenSUSE 15.3 and the latest EPEL, distros are providing newer versions of Singularity that OpenHPC. The decision was made at the TSC to remove the provided package in favor of using the distro packages.

Testing is updated. Singularity remains as a check using the distro package. It no longer uses modules as distros already add bash completion for Singularity.

Documentation is *not* updated. We need to keep in mind that OpenSUSE requires all Singularity users to be added to the "singularity" group or they will get permission denied errors.


ruixingw:
If this is removed from OHPC and distro packages are recommended, does it mean every compute nodes have to install Singularity to their own local disk? (assuming stateful installation)

jcsiadal:
@adrianreber take a look at this. The TSC decision last year was to remove Singularity from OpenHPC. I've rebased this to the current 2.x branch.

jcsiadal:
@adrianreber - I think changes to the CI are an issue here. The validation tests are looking for .spec file that shouldn't exist with the package removal.

adrianreber:
CI PR to only check added and modified files has been merged. Please rebase.

adrianreber:
@utdsimmons This looks good. I am happy to merge this. I remember this has been discussed in the TSC. Just wanted to double check with you.

ohpcsim:
I think this is good to push through. At some point we might have to revisit with the SingularityCE vs Apptainer schism but we need to let it settle a bit and figure out what RHEL and SUSE will do upstream before we concern ourselves with it again.

Thanks!

PR #1330, opened 2021-07-06
TAU update to 2.31.1

jcsiadal:
Updated TAU needed for LEAP/SLES 15U4 -- All Python2 scripts converted to Python3.

Rewrote most of the scripting. It's better aligned with RPM workflow now, so that all pre-build source modification is in the %prep section, and all code generation is in the %build section. The double-build is gone.

TAU is an odd package because of its insistence on absolute paths everywhere. Any remaining build path strings are similar to the current 2.29 RPM--they are strings in lib files; however, 'readelf -a' reports no leftover build path in libs or binaries. Test-suite results are identical to the 2.29 RPM results.


adrianreber:
@jcsiadal Can you rebase to have your changes run through local CI before pushing it to OBS? Please also squash the fixup commits.

jcsiadal:
@adrianreber Check out RHEL log line 3644:
grep: ./utils/FixMakefile.info: No such file or directory
grep: ./utils/FixMakefile.sed: No such file or directory

I'm not getting these errors on local Rocky8 builds. OBS SLES seems to be fine. Any idea about why OBS/rhel is failing to find these files?

adrianreber:
This is indeed a bit strange. At least you need to add the BR: `make`. The first rebuild on Rocky fails with: `/var/tmp/rpm-tmp.NDwDcZ: line 89: make: command not found`

I can locally fix the other error `configure: error: C compiler cannot create executables` with the following line: `sed -e "s,/usr/bin/gcc,gcc,g" -i utils/include/Makefile.skel`. There is already a similar line for `g++`. It seems if the OS gcc is not installed the build fails. So the whole TAU build might have never used our `gcc` but always `/usr/bin/gcc`. Seems to be hardcoded. With a buildroot that has no gcc installed it now fails.

This still leaves the message `grep: ./utils/FixMakefile.sed: No such file or directory`. That still happens on the Rocky builds locally but it does not seem to be a problem for finishing the build locally.

adrianreber:
Please avoid fixup commits in pull requests. Please squash this into a single commit. If there are logical independent commits in a PR keeping them separate makes sense, but in this case it is just fixups which can avoid having in the main repository.

CI is happy, after squashing, this is ready to be merged.

jcsiadal:
You may see an interim update or two, as I need to confirm there are no errors hiding in the successful build. Then I'll squash and push tonight.

PR #1334, opened 2021-07-08
Compatability layer for Arm 21 compiler release.

rhenwood-arm:
The current Arm compiler release is 21. Update the compatibility spec file to use this release.


koomie:
Just noticed this PR got funky with 101 file changes.  Do you want to update (or create a fresh PR) to isolate your change. Alternatively, if you can confirm that 4a31faefb487c9e5f8280db8575dc486cbe10876 is the only intended change, I can cherry-pick that directly.

PR #1337, opened 2021-07-12
CMake update to 3.20.5

jcsiadal:
This updates CMake to version 3.20.5. Also revised the build to align with rpmbuild stages and made the bootstrap parallel.

Quick background: Several packages that were building without error on SLES15SP3 using the distro cmake were moved to the OpenHPC 2.3 cmake and then encountered build errors. I updated the OpenHPC cmake to this version (3.20.5) and the packages again built without errors.

Will also post update to ohpc-filesystem. Individual packages shouldn't need to own the root utilities directory.


koomie:
Superseded by #1356

PR #1339, opened 2021-07-13
Updated filesystem RPM

jcsiadal:
Signed-off-by: Jeremy Siadal <jeremy.c.siadal@intel.com>

Need to add utils subdirectory to the filesystem RPM.

Looking at the current layout, there are two authoritative (hardcoded) sources for the OHPC FS hierarchy: OHPC_macros and ohpc-filesystem. They do not reference each other and they're synchronized manually. instead, we should maintain a single authoritative source and all other RPMs should refer to it. OHPC_macros makes the most sense.

I've added OHPC_macros and replaced hard-coded values with macros. The resulting RPMs are identical to the original, except for utils and doc being added to the FS.


jcsiadal:
Also added the URL. Not sure why that hasn't been picked up by our automated checks before.

PR #1340, opened 2021-07-20
NetCDF update to supprt OpenSUSE 15.3

jcsiadal:
Update needed for netcdf-fortran to support building on LEAP 15.3 with Intel PSXE Compilers. I updated all three packages at the same time.

1. The code authors prefer cmake/make over configure/make, so specfiles are updated to use that method. I also tested pure cmake and it also works. I had some build issues with the current cmake, but both the distro-provided and latest cmake versions worked. CMake update submitted as  PR#1337.
(I prefer we move exclusively to our cmake. I've see too many issues when the cmake version changes and this let's us ignore changes to the distro cmake)
2. Updated the specfiles to conform to most of the distro RPM spec standards.
3. Updated description with the latest text from Unidata.
4. Note that using cmake changes the resulting netcdf-cxx4 lib name. Not sure if they intended that and devs have not responded to my issue.


jcsiadal:
Closing. Will issue new update for OpenSUSE 15.4.

PR #1341, opened 2021-08-03
intel-compiler/mpi-devel package update to oneAPI

jcsiadal:
This updates the intel-compiler-devel and intel-mpi-devel packages to use oneAPI. I'm interested in having others try this out and getting feedback. Given that oneAPI has been out for half-a-year, I'd like migrate before the next release. That said, I'll need some help with OBS Testing. it will require modifying the base images to install the toolkit release package instead of PSXE and I don't have experience with that.

Reference issue #1317 

This was tested on CentOS 8.4 and OpenSUSE 15.3 with the latest PSXE and two oneAPI releases. Tested builds of various package using the latest oneAPI HPC TK. Packages seem to be working, but there are a lot of warnings about unsupported command line options.

This is a short summary of changes:
 - The compiler package includes a oneAPI release package that configures the YUM repo. After this is installed, the compiler-devel package can be installed and will install the oneAPI HPC Toolkit as a dependency.
 - Included oneAPI modulefiles are used. OHPC intel/impi module names remain the the same, but are greatly simplified, prepending the oneAPI module directory only when needed.
 - The legacy compilers are used, to simplify initial migration.
 - A minimum version is set. Older oneAPI builds don't work with LMOD without patching oneAPI modulefiles.
 - Install/removal of the RPMs won't clobber any generated files that the user has modified.
 -  The previous PSXE package is renamed and works the same way it used to. I tried to simplify a lot of it, since the 1.x stuff isn't relevant any more.

I would ultimately like move the module generation to a utility and call it in pre and post. That way an admin could update their configuration without reinstalling the RPM.

### DIFF
--- a/components/compiler-families/intel-compilers-devel/SOURCES/mod_generator.sh
+++ b/components/compiler-families/intel-compilers-devel/SOURCES/mod_generator.sh
@@ -77,7 +77,6 @@ done
 
 echo " "
 echo "# OpenHPC machine generated from `basename $inFile`"
-echo " "
 
 while read var ; do 
     varName=`echo $var | awk -F = '{print $1}'` || exit 1
@@ -99,7 +98,6 @@ while read var ; do
     fi
 done < $OUTFILE
 
-echo " "
 echo "# end of machine generated"
 echo " "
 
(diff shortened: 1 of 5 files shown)


jcsiadal:
Locally built all OpenHPC 2.x packages on OpenSUSE 15.3 with this PR and other applicable PRs applied. Everything builds except for dimemas, metis, mumps, petsc, plasma, scipy, slepc, and superlu_dist. I'm going to work on updates for these.

jcsiadal:
I pushed a minor fixup. All packages are building with oneAPI 2021.3 except for R and SciPy. Both of those packages should be updated to new versions. I'm able to build R using its latest release.

Trilinos also builds correctly with a one line patch; it was not building successfully before on Parallel Studio.

koomie:
Landed with some follow-on tweaks in #1390.

PR #1342, opened 2021-08-09
Remove Flex from OpenHPC

jcsiadal:
The goal here is to remove an unnecessary package, as distros provide current versions of flex.

It's currently used just for compiling dimemas, and is only needed for SUSE. The dimemas build requires the static flex libs which are not available in SUSE, while the rest of the BSC tools compile with the shared libs. Dimemas could likely be patched, but it would be a much larger task.

Instead of supporting a separate package just for dimemas build requirements, I in-lined a temp flex build to provide the static libs.


adrianreber:
@jcsiadal I see there are a couple of PR from you. Could you please rebase your PRs.

jcsiadal:
Need to reevaluate this PR. Will resubmit if needed.

PR #1365, opened 2021-10-12
Match spack-ohpc dependencies to published system requirements

mikerenfro:
From discussion with @adrianreber on Slack this morning, we noticed on a fresh install of OHPC with Spack, that we were unable to run some `spack install` commands out of the box, due to us not yet having `g++` installed, nor had we run `spack compiler find` after installing OHPC-packaged compilers to fix root's `compilers.yaml`.

This PR ensures all the [system prerequisites](https://spack.readthedocs.io/en/latest/getting_started.html#system-prerequisites) are dependencies of the Spack package. It may not provide optimized compilations, but should work better out of the box.

### DIFF
--- a/components/dev-tools/spack/SPECS/spack.spec
+++ b/components/dev-tools/spack/SPECS/spack.spec
@@ -39,6 +39,7 @@ Requires: tar
 Requires: gzip
 Requires: unzip
 Requires: bzip2
+Requires: xz
 Requires: zstd
 Requires: file
 Requires: gnupg2
--- a/components/dev-tools/spack/SPECS/spack.spec
+++ b/components/dev-tools/spack/SPECS/spack.spec
@@ -44,7 +44,6 @@ Requires: file
 Requires: gnupg2
 Requires: git
 
-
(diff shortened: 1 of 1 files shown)


mikerenfro:
Missed the signoff instructions at first, we'll get those fixed before marking ready for review.

sjpb:
I think this is missing `xz`?

`curl` is also not listed in the docs as a dependency, but [actually is](https://spack.readthedocs.io/en/latest/getting_started.html#curl-git-mercurial-etc) - I know the CentOS GenericCloud I use includes that so not sure if it should be explicitly listed here.

juleskerley:
> I think this is missing `xz`?
> 
> `curl` is also not listed in the docs as a dependency, but [actually is](https://spack.readthedocs.io/en/latest/getting_started.html#curl-git-mercurial-etc) - I know the CentOS GenericCloud I use includes that so not sure if it should be explicitly listed here.

The xz is supposed to be in there. I must have somehow looked over it. I will quickly correct that.

PR #1382, opened 2021-11-09
Warewulf 4 package

jcsiadal:

Addresses issue #1314.

This is the current working package for Warewulf 4. It does not alter any Warewulf 3 components from the distro, so no changes to existing recipes are needed if we want to make this available for early adopters.

It uses the latest release, patched to fix helpfiles and RPM build errors (from merged PR195) and to relocate shared files.

Explicit conflicts are set for the previous WW3. It will not upgrade from WW3, but will upgrade itself.


jcsiadal:
Will resubmit with a later version or commit. I want to capture some of the latest changes to the WW4 project.

PR #1390, opened 2021-11-12
Oneapi

koomie:


whophil:
I know this is cutting edge, but is there any guidance on how to _use_ this? I see oneAPI mentioned briefly in the latest install guides, but there is a broken reference, which I can only assume will answer all the questions I have :)

image

koomie:
That should be referring to the section entitled "Optional Development Tool Builds" which highlights installing oneAPI from public repositories on your head node using standard package manager semantics.  The packages with get installed into `/opt/intel` so that is why there is the optional suggestion in the bit you referenced to share that location over NFS.

Can you say which .pdf you see the broken link?  I've checked the Rocky SLURM variant and it looks ok.

whophil:
Thanks @koomie . I am seeing that in https://github.com/openhpc/ohpc/releases/download/v2.4.GA/Install_guide-rocky8-Warewulf-SLURM-2.4-aarch64.pdf, linked from the wiki page https://github.com/openhpc/ohpc/wiki/2.X.

That particular version of the install guide also makes no reference to oneAPI in the section **Optional Development Tool Builds**

koomie:
> Thanks @koomie . I am seeing that in https://github.com/openhpc/ohpc/releases/download/v2.4.GA/Install_guide-rocky8-Warewulf-SLURM-2.4-aarch64.pdf, linked from the wiki page https://github.com/openhpc/ohpc/wiki/2.X.
> 
> That particular version of the install guide also makes no reference to oneAPI in the section **Optional Development Tool Builds**

Ahh!  oneAPI only makes sense for `x86_64` so should not be in the `aarch64` recipes. I will fix that and update; thanks for pointing this out.

whophil:
I realized only after sending the link that I had (accidentally) clicked on the aarch64 PDF. Thanks for the help!

whophil:
One last question here - is there a sanctioned way to enable Intel MPI? Following the install guide, I have installed:
- intel-oneapi-toolkit-release-ohpc
- intel-compilers-devel-ohpc
- intel-mpi-devel-ohpc

This gives me an `intel` module, but it does not activate the paths for oneAPI's MPI.

Also, if there is a more appropriate place to ask these types of questions, please let me know! I have tried (unsuccessfully) to join both the OpenHPC mailing lists and Slack channels.

koomie:
> One last question here - is there a sanctioned way to enable Intel MPI? Following the install guide, I have installed:
> 
> * intel-oneapi-toolkit-release-ohpc
> * intel-compilers-devel-ohpc
> * intel-mpi-devel-ohpc
> 
> This gives me an `intel` module, but it does not activate the paths for oneAPI's MPI.

Correct. If you want intel mpi, then you need to load that module as well. So, assuming you have openmpi4 loaded currently you would: `module swap openmpi4 impi`
> 
> Also, if there is a more appropriate place to ask these types of questions, please let me know! I have tried (unsuccessfully) to join both the OpenHPC mailing lists and Slack channels.

Github is fine, although probably better to create a new ticket.  And, it would be nice to know what type of issues you are having with joining the mailing list and slack.

PR #1402, opened 2022-01-13
components/runtimes/singularity: Add build dependency

solj:
Singularity requires `make` in order to build.


adrianreber:
Sounds like a good idea. Thanks. Merged.

PR #1410, opened 2022-03-16
fftw.spec: fix typo & enable sse2/avx/avx2

### LINKED ISSUE #1409: Typo in fftw.spec: FFTW included in openhpc was never compiled with any SIMD intrinsics & enable avx2/avx512
This line in `fftw.spec` missed two closing `"`s that cause the `FFTW` was never compiled with any SIMD intrinsics:
https://github.com/openhpc/ohpc/blob/86eb69ff4dcfe0db176fefe0cf97d3bb9ff90075/components/parallel-libs/fftw/SPECS/fftw.spec#L75

While you look into this line, please have a look at the next line and consider adding more SIMD options. Currently, it compiles only with `sse2` and `avx` intrinsics. According to [FFTW documentation ](https://www.fftw.org/fftw3_doc/Installation-on-Unix.html), it does no harm to include recent SIMD options even when running on old machines. 
> "FFTW will try to detect at runtime whether the CPU supports these extensions. That is, you can compile with--enable-avx and the code will still run on a CPU without AVX support."

I suggest adding `--enable-avx2` and `--enable-avx512` for better performance on recent CPUs.

ruixingw:
This fix openhpc/ohpc#1409. Fix a typo to support `sse2`/`avx` and in addition adds `avx2`/`avx512` SIMD extensions.

Tested on a machine with Intel Xeon E5-2696 v3 (has avx2 but not avx512) by following `Rebuilding Packages from Source` in the Install Guide Rocky 8.5/x86 64, rebuild and reinstall `fftw-gnu9-openmpi4-ohpc`.
```
[root@master lib]# strings /opt/ohpc/pub/libs/gnu9/openmpi4/fftw/3.3.8/lib/libfftw3.so | grep have
fftw_have_simd_sse2
fftw_have_simd_avx
fftw_have_simd_avx2
fftw_have_simd_avx2_128
fftw_have_simd_avx512
fftw_have_simd_sse2
fftw_have_simd_avx2
fftw_have_simd_avx512
fftw_have_simd_avx
fftw_have_simd_avx2_128
```

comparing with original `fftw-gnu9-mpich-ohpc`:
```
[root@master lib]# strings /opt/ohpc/pub/libs/gnu9/mpich/fftw/3.3.8/lib/libfftw3.so | grep have
[root@master lib]#
```

This shows the modified `fftw.spec` adds SIMD supports. 

EDIT: avx512 is removed because the upstream isn't ready.

### DIFF
--- a/components/parallel-libs/fftw/SPECS/fftw.spec
+++ b/components/parallel-libs/fftw/SPECS/fftw.spec
@@ -72,10 +72,10 @@ rm -f config.cache
 
 for i in %{precision_list} ; do
 	LOOPBASEFLAGS=${BASEFLAGS}
-	if [[ "${i} == "single" || "${i} == "double" ]]; then
+	if [[ "${i}" == "single" || "${i}" == "double" ]]; then
 		# taken from https://src.fedoraproject.org/rpms/fftw/blob/master/f/fftw.spec
 %ifarch x86_64
-		LOOPBASEFLAGS="${LOOPBASEFLAGS} --enable-sse2 --enable-avx"
+		LOOPBASEFLAGS="${LOOPBASEFLAGS} --enable-sse2 --enable-avx --enable-avx2"
 %endif
 %ifarch aarch64
 		LOOPBASEFLAGS="${LOOPBASEFLAGS} --enable-neon"


adrianreber:
Thanks. Looks good. GitHub currently seems to have problems with GitHub Actions, so the CI is not triggered.

As this is your first PR here @utdsimmons @koomie need to approve the CI run manually (once GitHub is working again correctly).

boegel:
Be careful with building FFTW with AVX-512, see https://github.com/easybuilders/easybuild-easyblocks/pull/1416 for detailed performance analysis when doing so, and also upstream issue @ https://github.com/FFTW/fftw3/issues/143 (which states "We have never tested FFTW on Skylake")

adrianreber:
Please add a `Signed-off-by:` to your commit. This can easily be done using `git commit -s --amend` followed by a force push to the branch of this PR.

Have you seen @boegel's comment? If you cannot test AVX-512 support maybe it would be better to not enable it.

ruixingw:
> Please add a `Signed-off-by:` to your commit. This can easily be done using `git commit -s --amend` followed by a force push to the branch of this PR.
> 
> Have you seen @boegel's comment? If you cannot test AVX-512 support maybe it would be better to not enable it.

I have changed the commit and included the `Signed-off-by:`.

Thanks @boegel for pointing that out. I think it is better to add `avx512` after https://github.com/FFTW/fftw3/issues/143 is solved. I have removed `avx512` from the commit.

ruixingw:
I see that the broken CI was fixed in a recent commit so I merged it. (not very experienced with git, is this the correct way?)

adrianreber:
> I see that the broken CI was fixed in a recent commit so I merged it. (not very experienced with git, is this the correct way?)

I will not say that is incorrect, but I would do it differently. You now have a merge commit in this pull request which is unnecessary (from my point of view).

I would do a `git pull --rebase`. That way git should use the latest content from OpenHPC and put your commit on top of it. Then do a `git push --force`.

ruixingw:
> > I see that the broken CI was fixed in a recent commit so I merged it. (not very experienced with git, is this the correct way?)
> 
> I will not say that is incorrect, but I would do it differently. You now have a merge commit in this pull request which is unnecessary (from my point of view).
> 
> I would do a `git pull --rebase`. That way git should use the latest content from OpenHPC and put your commit on top of it. Then do a `git push --force`.

Thanks! I now made a rebase.

adrianreber:
Looks good. @utdsimmons can you approve the CI run. First time contributors do need approval before CI starts to run.

ohpcsim:
approved and ran successfully; going to rebase and merge

PR #1411, opened 2022-03-17
Avoid killing user sessions on slurm.epilog

viniciusferrao:
This PR fixes two issues and a cosmetic issue on
`slurm.epilog.clean` script:

* The cosmetic one is trivial, tab vs spaces.

* The script should not run for UIDs lower than 1000 instead of 100.
Every UID lower than 1000 is considered a system account, this PR
fixes the comparison on the guard clause.

* The last one is a special case when the HPC environment uses the
headnode also as a compute node, or on HPC systems where nodes
aren't available and a single big server takes care of everything.
To avoid killing legitimate user sessions, like graphical login or
ssh login, we check if `slurmctld` is running on the same machine
as the epilogue script. If it's not running them we just kill
everything as the old behavior. If `slurmctld` is running just exit
without killing the remaining processes.

Signed-off-by: Vinícius Ferrão <vinicius@ferrao.net.br>

### DIFF
--- a/components/rms/slurm/SOURCES/slurm.epilog.clean
+++ b/components/rms/slurm/SOURCES/slurm.epilog.clean
@@ -12,13 +12,13 @@ if [ x$SLURM_UID = "x" ] ; then
 	exit 0
 fi
 if [ x$SLURM_JOB_ID = "x" ] ; then 
-        exit 0
+	exit 0
 fi
 
 #
 # Don't try to kill user root or system daemon jobs
 #
-if [ $SLURM_UID -lt 100 ] ; then
+if [ $SLURM_UID -lt 1000 ] ; then
 	exit 0
 fi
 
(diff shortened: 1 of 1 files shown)


adrianreber:
Please remove the empty commit.

adrianreber:
This PR sounds like a good idea. I am curious if there is a better way to detect the head node besides looking for a running `slurmctld`?

If you remove the empty commit and do a force push CI should be running and this can be merged.

viniciusferrao:
I've removed the empty commit @adrianreber, and forced push. It's good now.

I was thinking about the same question when I made this modification. The real issue is in two cases: a big server, those 128 cores AMD EPYC ones, with only one machine that acts as compute, head and login node, and the other corner case is when it's a small cluster but the head and login node are also a compute node. In both cases the common denominator is the existence of `slurmctld` running on the same machine which is the login node.

There's another case where the commit does not apply, it's when you have a server acting as login and compute node, but not as a head node. In this case this commit will do nothing. But I can't see any valid usage in this specific case, it simply does seems to make any sense.

So looking for `slurmctld` looked as a good candidate. I don't know any mechanism available on SLURM to achieve the same functionality to be honest.

Thanks.

adrianreber:
Also merged. Thanks.

PR #1408, opened 2022-03-11
Added links to gcc and gfortran

viniciusferrao:
This commit fixes missing symbolic links and wrapper scripts for gcc
and gfortran. Fixes #328 and add new functionality.

cc -> gcc
c89 -> wrapper
c99 -> wrapper
f95 -> gfrotran

Those files are needed on some software that hardcode those binaries
on it's code during runtime. Since compute nodes don't usually have
"Development Tools" installed from default system packages, it's
execution fails with command not found.

As a real world example, ANSYS Fluent would fail to execute:

```
-I/opt/ohpc/pub/apps/ansys_inc/v211/fluent/include -I. -c udf_names.c
/bin/sh: cc: command not found
make[3]: *** [udf_names.o] Error 127
```

Providing `cc`, `c89`, `c99` and `f95` will solve those specific
issues during execution. Patching and testing was done in a running
production OpenHPC cluster and it worked as expected:

```
$ which {cc,c89,c99,f95}
/opt/ohpc/pub/compiler/gcc/9.4.0/bin/cc
/opt/ohpc/pub/compiler/gcc/9.4.0/bin/c89
/opt/ohpc/pub/compiler/gcc/9.4.0/bin/c99
/opt/ohpc/pub/compiler/gcc/9.4.0/bin/f95
```

Code is based heavily on EL8 gcc package:
https://git.centos.org/rpms/gcc/blob/c8/f/SPECS/gcc.spec

Signed-off-by: Vinícius Ferrão <vinicius@ferrao.net.br>

PR #1412, opened 2022-03-18
Update Dockerfile.check_spec

viniciusferrao:
CI is broken due to CentOS end-of-life, as observed on #1408.

Changing it to Rocky Linux would not only solve CI while also keeping the project aligned with it's supported distribution.


viniciusferrao:
Oops, it was duplicated to #1413. Sorry for that.

PR #1415, opened 2022-04-10
Increase reliability of chronyd for compute nodes

viniciusferrao:
This commit will enhance the reliability of time synchronization on compute
nodes if there's a preexisting clock difference.

By enabling the compute nodes to "iburst" on the server, it will synchronize
it's clock with headnode imediatelly after chrony (as a client) started.

To further enhance it's reliability we also mark the headnode chronyd server
as a local stratum. In an event of unproper time fetch from upstream time
servers the headnode will be the default timekeeper for it's compute nodes.

This changes modifies the behavior of software like munge, that refuses to
work if there's a significant clock skew between the server and the client.
Avoiding the cluster to be offline due to clock synchronization issues.

Signed-off-by: Vinícius Ferrão <vinicius@ferrao.net.br>

### DIFF
--- a/docs/recipes/install/centos8/x86_64/warewulf/slurm/steps.tex
+++ b/docs/recipes/install/centos8/x86_64/warewulf/slurm/steps.tex
@@ -149,7 +149,7 @@
 # Add Network Time Protocol (NTP) support
 [sms](*\#*) (*\chrootinstall*) chrony
 # Identify master host as local NTP server
-[sms](*\#*) echo "server ${sms_ip}" >> $CHROOT/etc/chrony.conf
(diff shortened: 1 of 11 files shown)


adrianreber:
Sounds good from my point of view. It is also mentioned in the documentation.

Do you mind updating the commit message to include a link to the documentation. I found this one:

https://chrony.tuxfamily.org/faq.html#_what_is_the_minimum_recommended_configuration_for_an_ntp_client

adrianreber:
CC: @utdsimmons

viniciusferrao:
Done @adrianreber.

Thank you for the quick pull review.

adrianreber:
Merged. Thanks!

PR #1418, opened 2022-04-30
Use dnf instead of yum and it's plugins

viniciusferrao:
This PR is aimed to replace `yum` and it's utilities: `yumdownloader`, `yum-config-manager` and `yum-builddep` in favor of `dnf` which is the default package manager command on EL8 and upwards.

### DIFF
--- a/docs/recipes/install/centos8/x86_64/warewulf/slurm/steps.tex
+++ b/docs/recipes/install/centos8/x86_64/warewulf/slurm/steps.tex
@@ -20,18 +20,18 @@
 \newcommand{\arch}{x86\_64}
 
 % Define package manager commands
-\newcommand{\pkgmgr}{yum}
+\newcommand{\pkgmgr}{dnf}
 \newcommand{\addrepo}{wget -P /etc/yum.repos.d}
 \newcommand{\chrootaddrepo}{wget -P \$CHROOT/etc/yum.repos.d}
-\newcommand{\clean}{yum clean expire-cache}
-\newcommand{\chrootclean}{yum --installroot=\$CHROOT clean expire-cache}
-\newcommand{\install}{yum -y install}
-\newcommand{\chrootinstall}{yum -y --installroot=\$CHROOT install}
-\newcommand{\groupinstall}{yum -y groupinstall}
-\newcommand{\groupchrootinstall}{yum -y --installroot=\$CHROOT groupinstall}
-\newcommand{\remove}{yum -y remove}
-\newcommand{\upgrade}{yum -y upgrade}
-\newcommand{\chrootupgrade}{yum -y --installroot=\$CHROOT upgrade}
+\newcommand{\clean}{dnf clean expire-cache}
+\newcommand{\chrootclean}{dnf --installroot=\$CHROOT clean expire-cache}
+\newcommand{\install}{dnf -y install}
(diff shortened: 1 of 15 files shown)


adrianreber:
I am undecided. Both commands seem valid. The RHEL 8 documentation uses yum almost everywhere. I am personally also using the dnf command. Not sure what the documentation should refer to.

viniciusferrao:
Hi @adrianreber I get your point. I was tricked by this line specifically:

https://github.com/openhpc/ohpc/blob/1866c541bbd3585ccb417043c2ecc82da880e877/docs/recipes/install/common/centos_repos.tex#L23

I came to a conclusion this was a first effort to move the documentation to `dnf`. The next line is actually `dnf` instead of `yum`:

https://github.com/openhpc/ohpc/blob/1866c541bbd3585ccb417043c2ecc82da880e877/docs/recipes/install/common/centos_repos.tex#L24

The proper `yum` version would be `yum-config-manager` with `yum-utils` being installed on line 23, and not `yum config-manager` (observe the space after `yum`).

Internally I also only use OpenHPC with `dnf` instead of `yum` so I made the PR with this premise, since we already tested and updated internal docs with `dnf` instead.

I agree that Red Hat's documentation for EL8 uses `yum` in many places, referring it as YUMv4, I haven't checked it before doing the PR.

I'm not sure if we should reject it or not, of course I want the PR merged since I've made it, but I also agree with your point. If we look at EL9 Beta documentation `yum` is now clearly deprecated: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/9-beta/html-single/managing_software_with_the_dnf_tool/index

So as a suggestion, if we can't merge it on the [2.x](https://github.com/openhpc/ohpc/tree/2.x) branch, maybe we target it for what would become 3.x?

If that sounds like a good idea, I may create another PR to implement the changes on [2.x](https://github.com/openhpc/ohpc/tree/2.x) docs to avoid the `dnf` confusion. Or if you want to merge this anyway I'm obviously fine with it. 👍 

Thank you.

adrianreber:
We discussed this in the TSC and it seems everyone agrees that this is the right thing to do.

As this heavily changes the recipes and thus may break CI: @koomie what do you think?

koomie:
I have mixed opinions here.  RHEL8 does use yum predominantly in examples and this is what most people know.  In fact, I suspect many folks do not even know that they are using wrappers to dnf and a switch will bring up more questions.  Anyway, if we did this, the provisioning should be updated so that necessary dnf plugins show up as part of a base image.  From the documentation/CI side, we do already abstract the package manager installs since we have to deal today with `yum` and `zypper`, so in theory, it will not be too bad, but I'm sure some small issues will arise.

Personally, I think switching to using `dnf` would make more sense with the introduction of a future rhel9 based recipe.

viniciusferrao:
Hi @koomie with the upcoming release of OpenHPC 3.0 is this PR still relevant? Do you want/need it to be updated?

martin-g:
I am +1 to use `dnf` instead of `yum`! 

> but I'm sure some small issues will arise.

Agreed! It might be a a bit late now when we are so close to the release of 3.0.

adrianreber:
Closing in favour of #1940. dnf should be the default in the 3.x branch from now on.

PR #1423, opened 2022-05-26
Updated .gitignore file

jcsiadal:

Updates the .gitignore file to include all directories created during the build process, and SOURCE files created or downloaded by get_source.sh.


adrianreber:
Thanks. I like that a lot. Especially looking at the output of `git status` in my checkout.

PR #1425, opened 2022-06-01
ci: use jitterbit/get-changed-files@v1

adrianreber:
jitterbit/get-changed-files@v1 provides an easier way to figure out the changed files in a pull request. Let's switch to this for better CI results.

### DIFF
--- a/components/dev-tools/spack/SPECS/spack.spec
+++ b/components/dev-tools/spack/SPECS/spack.spec
@@ -25,7 +25,6 @@ Source0:	https://github.com/LLNL/%{pname}/archive/v%{version}.tar.gz
 BuildRequires: rsync
 BuildRequires: python3
 Requires: bash
-Requires: curl
 Requires: coreutils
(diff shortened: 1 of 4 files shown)


components/dev-tools/spack/SPECS/spack.spec
```diff
@@ -25,7 +25,6 @@ Source0:	https://github.com/LLNL/%{pname}/archive/v%{version}.tar.gz
 BuildRequires: rsync
 BuildRequires: python3
 Requires: bash
-Requires: curl
```
koomie, line 28: Was this intentional for the CI processing update?

adrianreber: I was looking for something to test first. But this is an actual fix. That line exists twice.

koomie: Ahh. Makes it an uber requirement :-)

PR #1429, opened 2022-06-07
warewulf-provision build fix for SUSE 15.4+

jcsiadal:

Busybox source in WW3 will not build against newer glibc. This works around the issue by using the Busybox included with OpenSUSE.

Changes the specfile to use the SUSE release virtual package, allowing this package to also build on SLED and SLES.

### DIFF
--- a/components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
+++ b/components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
@@ -42,6 +42,7 @@ Conflicts: warewulf < 3
 BuildRequires: autoconf
 BuildRequires: automake, make
 BuildRequires: which
+BuildRequires: gcc
 BuildRequires: warewulf-common%{PROJ_DELIM}
 BuildRequires: libselinux-devel, libacl-devel, libattr-devel
 BuildRequires: libuuid-devel, device-mapper-devel, xz-devel
--- a/components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
+++ b/components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
@@ -41,14 +41,15 @@ Requires: %{name}-initramfs-%{_arch} = %{version}-%{release}
 Conflicts: warewulf < 3
 BuildRequires: autoconf
 BuildRequires: automake
+BuildRequires: which
 BuildRequires: warewulf-common%{PROJ_DELIM}
 BuildRequires: libselinux-devel, libacl-devel, libattr-devel
 BuildRequires: libuuid-devel, device-mapper-devel, xz-devel
 BuildRequires: libtirpc-devel
 
 # charles.r.baird@intel.com - required to determine where to stick warewulf-httpd.conf
-%if 0%{?suse_version}
-BuildRequires: openSUSE-release
+%if 0%{?suse_version} || 0%{?sle_version}
+BuildRequires: distribution-release
 %endif
 
 %if 0%{?rhel:1}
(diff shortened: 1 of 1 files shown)


components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
```diff
@@ -91,6 +93,12 @@ Requires: parted, autofs, e2fsprogs
 BuildRequires: libarchive.so.13()(64bit)
 Requires: libarchive.so.13()(64bit)
 %global CONF_FLAGS --with-local-e2fsprogs --with-local-libarchive --with-local-parted --with-local-partprobe
+# The included Busybox will not build on OpenSUSE 15.4
+%if %{?sle_version} >= 150400
```
adrianreber: CI is not happy about this line. I see a couple of `%if 0%{?sle_version} >= 150000` to make sure it works correctly if `sle_version` is undefined.

adrianreber:
@jcsiadal Are you motivated to extent the CI build step to also run on a SLES/LEAP container? It would require some changes in the scripts to call not call dnf/yum but the right tool on SLES/LEAP.

components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
jcsiadal: That was my oversight and it's fixed now. The CI is complaining that it cannot find the dependencies it needs, unless I'm misreading the error message. Any ideas about why the SRPM rebuild is failing?

jcsiadal: Dug into this more. The RPM is created successfully. The CI then installs dependencies using "dnf builddep <SRPM>" which installs *most* of the required dependencies. After rpmbuild correctly identifies that dependencies are missing, the test fails.
The question is: why does rpmbuild use the correct dependency list, but dnf-builddep is missing dependencies from its install list? Both are using the same SRPM.

adrianreber: Just saw your comment right now. Sorry for the delay. The script `misc/shell-functions` defines `rhel` to `7`. This is wrong here. I was running it on Fedora when I initially wrote it and needed something to build SRPMS for RHEL based systems.

I will just remove it and let it use the settings from the current system.

adrianreber:
Please rebase. I changed one of the scripts to no longer set 'rhel' to '7'.

jcsiadal:
I also removed the conditional for adding /usr/sbin tp the PATH. This was only for SUSE before, but looks like it's impacting the CI for RHEL builds now. If these builds are running as a regular user, is that user able to execute /usr/sbin/mkfs.ext4?

adrianreber:
> I also removed the conditional for adding /usr/sbin tp the PATH. This was only for SUSE before, but looks like it's impacting the CI for RHEL builds now. If these builds are running as a regular user, is that user able to execute /usr/sbin/mkfs.ext4?

I just tried it and it works locally as non-root, but not in a container. The BuildRequires `which` solves it however. warewulf tries to find mkfs.ext4 by using which and the container has which not installed.

About the commits in this PR, please squash all fix-up commits in the original commit and please also avoid merge commits. Using `rebase` should do that.

If you do another rebase, the LEAP based CI build should also be picked up and we can immediately see that your changes work on LEAP.

jcsiadal:
Finally rebased and working within the new CI.

adrianreber:
Thanks again for helping me figuring out the missing things in the new CI scripts. Good to see that we now get results for RHEL and LEAP.

PR #1430, opened 2022-06-08
EasyBuild v4.5.5

boegel:


boegel:
build test is failing because the EasyBuild bootstrap script (and EasyBuild itself) doesn't want to be run as `root`...

I think it's time to get rid of the bootstrap script entirely, a simple `pip install easybuild` should be a better approach now (and also, we deprecated the bootstrap script a while ago, see https://github.com/easybuilders/easybuild-framework/pull/3742)

adrianreber:
We shouldn't actually build packages in CI as root. That is a good idea. We need to be root to install the dependencies, but the actual `rpmbuild` should be as non-root. I will update the CI setup.

boegel:
@adrianreber Was it really the intention to close this via #1432?
Correct me if I'm wrong, but this implies that EasyBuild isn't updated to 4.5.5 yet?

adrianreber:
No, that was not the intention. Sorry for the wrong usage of Fixes. Could you rebase to see if the CI error is now fixed.

boegel:
@adrianreber done in #1433

PR #1436, opened 2022-06-21
Add Warewulf 4 to OpenHPC

jcsiadal:
Upgrades Warewulf to version 4.3. Because the build works differently, Warewulf 4.x can coexist in the repo with Warewulf 3.x.

The current Warewulf3 in OpenHPC is very old, but updating recipes to use the new version could take some time. I would like to see Warewulf4 offered as a preview/option without removing Warewulf3. In general, WW4 is smaller, simpler, and much easier to configure. Since node images can be imported from Docker containers, it's also easier to create nodes.

There is no simple upgrade path from 3 to 4, although it can be done manually with a little effort. Warewulf4 packages have conflicts configured for all previous versions. Warewulf packages are intended to upgrade cleanly from 4.3 to future versions.

Also needed to fix the build_srpm script. The 2.x version rebuilds every specfile it finds that contains the current specfile basename. That's plain wrong. The validation on this update was failing because of an error in a completely unrelated package.


adrianreber:
I think the build_srpm.sh changes are no longer necessary and looking at the changes it could be all squashed in one commit.

jcsiadal:
Squashed and rebased to 2.x.

ohpcsim:
Rebasing and merging and about to work on it on OBS.

adrianreber:
@jcsiadal https://obs.openhpc.community/package/show/OpenHPC:2.6:Factory/warewulf

The package doesn't build on LEAP. PTAL.

jcsiadal:
I see that. Missing a common package (we have a disconnect between GitHub validation and OBS build). I'll have a fix in tonight, as there's one more new change I want to add.

PR #1439, opened 2022-06-25
Upgrade gnu-compilers to GCC 12.2

### LINKED ISSUE #1437: GCC12
Update to GNU Compiler Collection 12.1 or later as the default toolchain.

My initial testing shows that most of the packages will build as-is or with minor modifications.

Would supersede #1209.

jcsiadal:

Resolves #1437.  I've been using this patch internally for my own builds on OpenSUSE 15.4 and Rocky 8.6.

Changes:
- Updates the GNU Compiler Collection from 9.5 to 12.2.0
- Includes patches to MPICH and MVAPICH, to allow the main toolchain components to build and load.
- Removed other references to Gnu compilers <= 8 and to the Python2 toolchain.


jcsiadal:
Looks like an issue with the CI. GCC12 builds successfully, but when it tries to build some of the other specs in this PR, the just-built GCC12 RPM is not available. Even if it was available, many of these specs need other dependent packages to be rebuilt against GCC12 first. Can we just review the logs manually to confirm that the gnu-12-compiler-ohpc RPM builds successfully?

RHEL log
2022-06-25T03:00:42.5517771Z Wrote: /home/ohpc/rpmbuild/RPMS/x86_64/gnu12-compilers-ohpc-12.1.0-1.ohpc.2.0.0.x86_64.rpm

LEAP Log
`2022-06-25T03:08:12.3018762Z Wrote: /home/ohpc/rpmbuild/RPMS/x86_64/gnu12-compilers-ohpc-12.1.0-1.ohpc.2.0.0.x86_64.rpm`

adrianreber:
Maybe it would make sense to split this into two PR. One just to update gcc to 12 and the second PR to switch everything else to use it. That way CI would be able to handle it once we to the build on the actual build system.

components/compiler-families/gnu-compilers/SPECS/gnu-compilers.spec
```diff
@@ -24,52 +23,55 @@ Summary:   The GNU C Compiler and Support Files
...
-%endif
-BuildRequires:  zlib-devel
-%if 0%{?sles_version} || 0%{?suse_version}
+BuildRequires:  binutils >= 2.30
+BuildRequires:  make >= 3.80
+BuildRequires:  gettext-devel >= 0.14.5
+BuildRequires:  flex >= 2.5.4
+BuildRequires:  texinfo >= 4.7
+BuildRequires:  m4 >= 1.4.6
+%if 0%{?sle_version} || 0%{?suse_version}
```
martin-g, line 42: Should line 79 below also be updated from `sles_version` to `sle_version` ?
Disclaimer: I don't know yet what is the difference between those but I noticed that it has changed here.

jcsiadal: Thanks, I've updated it to sle_version. sles_version is the older macro.

jcsiadal:
Rebased, updated to 12.2.0, and squashed commits.

tests/common/functions
```diff
@@ -45,11 +45,6 @@ check_exe() {
 # export python family variables
 get_python_vars() {
     local __python_family=$1
-    if [ "x$__python_family" == "xpython2" ]; then
-        export _python=python
-        export python_module_prefix=py2
-        export python_package_prefix=python
```
martin-g, line 51: Is Python2 no more needed/supported ?

components/OHPC_macros
```diff
@@ -75,21 +75,24 @@ BuildRequires: ohpc-buildroot
...
 %{!?python_family: %global python_family python3}
 
 # Compiler dependencies
 %if 0%{?ohpc_compiler_dependent} == 1
 
+%if "%{compiler_family}" == "gnu12"
+BuildRequires: gnu12-compilers%{PROJ_DELIM} >= 12.1.0
+Requires:      gnu12-compilers%{PROJ_DELIM} >= 12.1.0
+%global gnu_family gnu12
+%endif # gnu9
```
martin-g, line 89: `# gnu12`

tests/common/functions
jcsiadal: Python2 is no longer available in the latest SUSE 15 and deprecated in RHEL 8. Good time to drop it.

components/OHPC_macros
jcsiadal: Good catch. I also need to move it off the same line. The RPM developers have indicated they may turn end-of-line comments into errors.

PR #1441, opened 2022-06-28
Additional changes for GNU12 update

jcsiadal:
Follow-up to #1439

Updates MVAPICH, MPICH, and other components to compile with GCC 12.1. 


jcsiadal:
@adrianreber Looks like an issue with the --rebuild command in the CI. The "--define" argument is being interpreted by the "su" command instead of the "rpmbuild" command.

[UPDATE] I added another commit with fixes to run_build.py. It makes the "command" a single array element. I also quoted the --define argument; not doing this also causes it to fail.

adrianreber:
I applied all your changes. At this point the switch to gnu12 is almost done for the 2.6 branch. Thanks for your work preparing this switch.

PR #1442, opened 2022-06-30
PLASMA update to 21.8.29

jcsiadal:

Updates the PLASMA package from 2.8.0 to 21.8.9. All of the Python scripts are migrated to Python3.
For some reason the source tarball doesn't include the lua source in the plasma git tree. I removed it from the makefile and use the distro-provided Lua instead.
Here-doc module converted to native Lmod lua.

Arm builds need to be validated, as I don't have access to Arm systems to check the compiler options.


adrianreber:
Please rebase. I fixed the CI errors.

adrianreber:
Now it says: `doxygen: Command not found`. This seems a real error.

adrianreber:
To fix the RHEL build failure maybe something like CC=gcc is needed during `make install`. The Leap failure is the same problem as with boost. Somehow the PATH is broken and lmod cannot correctly handle it.

jcsiadal:
CC is already explicitly set in OHPC_setup_compilers. I suspect we're going to hit these same issues with a lot of component updates and I'd like to fix the core issue.

adrianreber:
If you squash the fixup commit in the first one this is ready from my point of view.

jcsiadal:
All commits squashed.

PR #1447, opened 2022-07-19
Boost update to 1.79.0

jcsiadal:

- Boost build update to 1.79.0
- Intel toolchain uses new clang-based compiler
- Build tested on Rocky 8.6 and OpenSUSE 15.4
- Build works with gcc 9.4 and 12.1
- Needs validation on ARM


adrianreber:
Please rebase. I fixed the CI errors.

adrianreber:
On leap it fails with:
```
Building B2 engine..

A C++11 capable compiler is required for building the B2 engine.
Toolset 'gcc' does not appear to support C++11.

> g++ -x c++ -std=c++11  check_cxx11.cpp
> g++ -x c++ -std=c++11 -D_GNU_SOURCE  check_cxx11.cpp
> g++ -x c++ -std=c++11 -pthread  check_cxx11.cpp

Failed to build B2 build engine
```
Strange error. Do you have any ideas how this happens?

jcsiadal:
Digging into the logs, I see: 
`g++: command not found`
That would also explain the C++11 requirement error. Maybe I'm misreading it, but the variable dump indicates that gnu9 isn't in $PATH, which should be done by ohpc_compiler_setup.

adrianreber:
> Digging into the logs, I see: `g++: command not found` That would also explain the C++11 requirement error. Maybe I'm misreading it, but the variable dump indicates that gnu9 isn't in $PATH, which should be done by ohpc_compiler_setup.

Yes, this is totally strange and I only see it on Leap. Somehow the module load result is not correctly propagated to the rpmbuild shell. If I do an `echo $PATH` after the setup compiler macro I see:
```
/usr/local/bin:/bin:/usr/bin ENV_ROOTPATH /usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin
```
which looks broken.

The `PATH` is set to `/usr/local/bin:/bin:/usr/bin ENV_ROOTPATH /usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin` at the start of the rpm build which seems to come from `/etc/login.defs`. That does not look correct.

Manually setting PATH to `/usr/local/sbin:/usr/local/bin:/sbin:/bin:/usr/sbin:/usr/bin` at the start of the `%build` section fixes this error for me. Not sure what happens on Leap and why.

jcsiadal:
You will see my test PR at #1449. I tried it with METIS (compiler dependent) and it validated without error. I tried it with fftw (MPI dependent) and it failed. So we're looking at a combination of OpenSUSE 15.4 and the MPI Compiler setup.

adrianreber:
This needs a rebase according to GitHub.

jcsiadal:
Rebased and all commits squashed.

adrianreber:
Thanks for fixing the CI setup.

adrianreber:
@jcsiadal I am playing around with adding aarch64 support to our PR CI and it seems like this PR broke boost on aarch64.

I think this change is problematic:
```diff
-./b2 -d2 -q %{?_smp_mflags} threading=multi link=shared variant=release \
-               --prefix=%{install_path} \
-               --user-config=./user-config.jam toolset=${toolset}
-
+./b2 -d2 -q %{?_smp_mflags} --user-config=./rpm-config.jam \
+     address-model="64" \
+     architecture="x86" \
+     threading="multi" \
+     link="shared" \
+     runtime-link="shared" \
+     variant="release" \
+     toolset=%{toolset}
```
Maybe you could provide a follow up once we aarch64 CI enabled in the OpenHPC repository.

PR #1451, opened 2022-07-26
Fix for OpenSUSE validation failures

jcsiadal:

The validation errors were a combination of running the su command with no login shell and running Lmod. This doesn't work under LEAP,.

This update does a few things:

- Updates validation to add login for the su command.
- Changes OpenSUSE to use 15.3, which is formally supported, instead of the latest version.
- Rewrote the compiler_setup macro. The rpm developers say that %if within a macro isn't supported (although it usually works). They also noted that line continuation doesn't do anything in specfile macros (they only work in macrofile macros), so I removed them.
- Added a newline in the FFTW spec, so I could validate this on GitHub.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -48,7 +48,7 @@ jobs:
   build_on_leap:
     runs-on: ubuntu-latest
     container:
-      image: registry.opensuse.org/opensuse/leap:latest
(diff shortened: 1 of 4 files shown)


adrianreber:
Thanks for the fix. Sounds correct.

PR #1453, opened 2022-07-29
ci: enable aarch64 CI runs

adrianreber:
This extends our CI to run builds also on AARCH64. The AARCH64 machine are available from Drone CI.

@utdsimmons For this to work we need to enable drone.io CI on the OpenHPC github repository.

See https://cloud.drone.io/adrianreber/ohpc/15/1/2 for a result building the plasma RPM on AARCH64 from a PR on my fork of OHPC.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -1,0 +1,24 @@
+---
+kind: pipeline
+type: docker
+name: aarch64 RHEL
+
+platform:
+  os: linux
+  arch: arm64
+
+steps:
+- name: build
+  image: docker.io/library/rockylinux:8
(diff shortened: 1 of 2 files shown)


ohpcsim:
looks good; rebasing and merging

adrianreber:
So it seems drone.io does not accept new registrations and I am not yet sure harness.io (the new owner) has the same features as drone.io.

@koomie Do we have an aarch64 system where I could install the github self hosted runner. Then we can run aarch64 CI jobs on our own aarch64 system.

PR #1454, opened 2022-08-01
ci: setup slurm on CI system for further tests

adrianreber:
This adds an additional step to our existing CI scripts to run some of the tests from the test suite in GitHub's CI setup.

For now I am only running 5 tests to see if it works. So far it looks good.

I have to rebuild slurm to allow multiple slurmd processes on one host. This step could be skipped if we include the multihost support in our regular slurm build.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -15,6 +15,7 @@ task:
     dnf install -y epel-release dnf-plugins-core git python3 http://repos.openhpc.community/OpenHPC/2/CentOS_8/aarch64/ohpc-release-2-1.el8.aarch64.rpm findutils rpm-build wget gawk jq which
     dnf config-manager --set-enabled powertools
     dnf config-manager --set-enabled devel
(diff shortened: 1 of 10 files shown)


tests/libs/fftw/tests/rm_execution
```diff
@@ -11,17 +11,23 @@ fi
 check_rms
 rm=$RESOURCE_MANAGER
 
-NODES=2
-TASKS=8
-ARGS=8
+if [ -z "$GITHUB_ACTIONS" ]; then
```
martin-g: What about CirrusCI (for Rocky on aarch64) ?

adrianreber: First goal was to get it working on GitHub Actions only. This can be easily changed later if we want to run it somewhere else.

That was my approach to this at least.

martin-g: According to https://cirrus-ci.org/guide/writing-tasks/#environment-variables you could use `CIRRUS_CI` to check whether it executes on Cirrus.

ohpcsim:
Let’s go ahead and give it a try. Nice work!

Issue #1403, opened 2022-01-28
hpc-workspace


koomie:
https://github.com/holgerBerger/hpc-workspace

georgiastuart:
@koomie , I work with @utdsimmons and I can build the SPEC for the hpc-workspace component if someone else isn't already assigned.

martin-g:
I think this issue could be closed.

https://github.com/openhpc/ohpc/issues/1691 - Fixed the build.
https://github.com/openhpc/ohpc/issues/1675 - issue for adding tests.

PR #1459, opened 2022-08-05
Add hpc-workspace to OpenHPC

georgiastuart:
This is an attempt to fulfill #1403 

[hpc-workspace](https://github.com/holgerBerger/hpc-workspace) is an admin tool for managing directories with associated expiration dates and share groups. 

This PR adds:
- hpc-workspace spec file
- hpc-workspace license
- hpc-workspace config template

Open questions:
- Is the admin category the correct spot for this component?
- I included the default daily cleanup script, should that be changed?

### DIFF
--- a/components/admin/hpc-workspace/SPECS/hpcworkspace.spec
+++ b/components/admin/hpc-workspace/SPECS/hpcworkspace.spec
@@ -54,7 +54,7 @@ working filesystems in an HPC environment.
 cmake \
   -DCMAKE_INSTALL_PREFIX=%{install_path} \
   .
-make -j 4
+make
 
 %install
 make DESTDIR=%{buildroot} install
--- a/components/admin/hpc-workspace/SPECS/hpcworkspace.spec
+++ b/components/admin/hpc-workspace/SPECS/hpcworkspace.spec
@@ -70,16 +70,8 @@ make
 
 %install
 make DESTDIR=%{buildroot} install
-%{__mkdir_p} %{buildroot}%{_sysconfdir}/cron.daily
 %{__mkdir_p} %{buildroot}%{_localstatedir}/log/%{pname}
 cp %{SOURCE1} %{buildroot}%{_sysconfdir}
-%{__cat} << EOF > %{buildroot}%{_sysconfdir}/cron.daily/%{pname}
-#!/bin/sh
-
-# Daily cleanup script for %{pname}
-%{install_path}/sbin/ws_expirer -c > %{_localstatedir}/log/%{pname}/expirer-`date +%y.%m.%d`
-find %{_localstatedir}/log/%{pname} -type f -ctime +90 -exec rm {} \;
-EOF
 
 %{__mkdir} -p %{buildroot}/%{OHPC_MODULES}/%{pname}
 %{__cat} << EOF > %{buildroot}/%{OHPC_MODULES}/%{pname}/%{version}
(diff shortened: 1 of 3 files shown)


components/admin/hpc-workspace/SPECS/hpcworkspace.spec
```diff
@@ -0,0 +1,122 @@
...
+working filesystems in an HPC environment.
+
+%prep
+%setup -q -n %{pname}-%{version}
+
+%build
+cmake \
+  -DCMAKE_INSTALL_PREFIX=%{install_path} \
+  .
+make -j 4
```
adrianreber: Please do not hard code the number of parallel tasks to 4. There is a macro for this.

components/admin/hpc-workspace/SPECS/hpcworkspace.spec
```diff
@@ -0,0 +1,122 @@
...
+%{_sysconfdir}/cron.daily/%{pname}
+%attr(4755, root, root) %{install_path}/bin/ws_allocate
+%attr(4755, root, root) %{install_path}/bin/ws_release
+%attr(4755, root, root) %{install_path}/bin/ws_restore
+
+
+%pre
+# provide specific uid/gid to ensure that it is the same across the cluster
+/usr/bin/getent group hpcws >/dev/null 2>&1 || \
+  /usr/sbin/groupadd -r hpcws -g 85
```
adrianreber: I am curious about the 85. Not sure we have a central place in openhpc where we decide on pre defined UIDs and GIDs. Just curious why the 85.

adrianreber:
Please have a look at the result of the CI run on Leap. There seems to be an error.

components/admin/hpc-workspace/SPECS/hpcworkspace.spec
georgiastuart: Looks like munge is 201 and slurm is 202. Would it be better to change it to 203?

adrianreber:
> This is an attempt to fulfill #1403 

Thanks!

> Open questions:
> - Is the admin category the correct spot for this component?

Sounds good to me.

> - I included the default daily cleanup script, should that be changed?

That is indeed a difficult question. On the one hand I think this is the correct solution, on the other hand it could be a site specific setup. I would leave it as it is. Make sure it is mentioned in the documentation. Also maybe the cron setup should be marked as a config file so that RPM does not overwrite local changes on upgrade.

components/admin/hpc-workspace/SPECS/hpcworkspace.spec
adrianreber: I have no idea. I was just curious. We probably need the input from @koomie. Maybe he remembers how we distribute static UIDs in Openhpc.

components/admin/hpc-workspace/SPECS/hpcworkspace.spec
georgiastuart: Resolved in ca83914

georgiastuart:
Open tasks:
- [ ] Mark cron job as config in spec
- [ ] Add hpc-workspace-ohpc to documentation files
- [x] Resolve UID/GID numbering

ohpcsim:
Hey @georgiastuart,

For the cron job, we've decided not to include it as part of the RPM. We will document what needs to be done in the recipes and let sysadmins decide when to run it themselves based on their local usage patterns.

For the UID/GID numbering, you can go ahead and remove the specific group choice (85) and we will let the order of operations decide the UID/GID numbering as we do for SLURM/PBS and Munge.

I'll go ahead and do the documentation as well. If you'd like to make the above changes, and resubmit great. If you don't have the time, let me know and I'll do it.

Thanks again for your contribution!

georgiastuart:
@mghpcsim Ready for the next review. 

- uid/gid is now 203 in spec and config template per our discussion 
- cron job removed

georgiastuart:
Investigating the test failure.

PR #1470, opened 2022-08-31
aarch64 fixes for Boost

### LINKED ISSUE #1456: boost broken on aarch64
The latest PR to update boost unfortunately broke the build on aarch64. Now that we have an `aarch64` build included in the CI run, it would be good to fix it.

@jcsiadal can you take a look at it? Hopefully with existing aarch64 CI it make it easier to test. Not sure if you have access to OBS again.

jcsiadal:

Resolves #1456.

A few edits so the Boost libraries will compile on Arm.

Test/build platform was Rocky 8 for aarch64, on a QEMU virtual machine (cpu=cortex-a57). Builds successful using gcc9/openmpi4 and arm1/openmpi4.

### DIFF
--- a/components/parallel-libs/boost/SPECS/boost.spec
+++ b/components/parallel-libs/boost/SPECS/boost.spec
@@ -86,8 +86,8 @@ for all users with minimal restrictions.
 %endif
 
 %if "%{compiler_family}" == "arm"
-which_armclang=`which armclang`
-where_armclang=`dirname ${which_armclang}`
+which_armclang=$(which armclang)
+where_armclang=$(dirname ${which_armclang})
 export PATH=${where_armclang}/../llvm-bin:$PATH
 %endif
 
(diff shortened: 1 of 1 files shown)


adrianreber:
aarch64 CI failed because it ran out of memory. Let me double the amount of memory to see if this helps your build to pass.

adrianreber:
I added more memory to the aarch64 build container (12GB instead of 4GB). Please rebase.

adrianreber:
It even OOMs with 12GB. So maybe we have to ignore CI for this PR.

jcsiadal:
FYI: I recommend at least 70GB. I have not been able to get Trillinos to compile successfully with anything under that. My VMs are set to 16GB minimum, with balloon to 128GB.

adrianreber:
> FYI: I recommend at least 70GB. I have not been able to get Trillinos to compile successfully with anything under that. My VMs are set to 16GB minimum, with balloon to 128GB.

The free CI service we are using has a maximum of 32GB and I do not want to request the maximum if we only need it for a couple of packages. So maybe we have to accept that some packages will not compile in CI, but still on the real build system.

PR #1475, opened 2022-09-15
Add buildrequires for lmod package

huangtianhua:


Yikun:
```
+ ./configure --prefix=/opt/ohpc/admin --with-redirect=yes --with-autoSwap=no
configure: error: in `/home/ohpc/rpmbuild/BUILD/Lmod-8.7.3':
configure: error: no acceptable C compiler found in $PATH
See `config.log' for more details
error: Bad exit status from /var/tmp/rpm-tmp.YxBjsb (%build)
```

Yes, otherwise get back the error like above, could you also fullfill the PR info?

Issue #1573, opened 2022-10-10
trilinos (13.4.0)


adrianreber:
Available at: https://github.com/trilinos/trilinos
Project home page: http://trilinos.org

koomie:
Tests passing in CI.

PR #1574, opened 2022-10-10
parallel-libs/trilinos: bump version to 13.4.0 (#1573)

adrianreber:


components/parallel-libs/trilinos/SPECS/trilinos.spec
```diff
@@ -46,7 +46,7 @@ BuildRequires:  phdf5-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 BuildRequires:  netcdf-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 BuildRequires:  python3
 
-%if "%{compiler_family}" != "intel" && "%{compiler_family}" != "arm"
+%if "%{compiler_family}" != "intel" && "%{compiler_family}" != "arm1"
```
martin-g, line 49: What is `arm1` ?
It looks like a temporary change to test how the new version behaves on `arm` but I might be wrong.

adrianreber: First of all, thanks for looking at the pull requests.

`arm1` is correct. That is the name of the compiler family as referenced here https://github.com/openhpc/ohpc/blob/2.x/components/OHPC_macros#L102 and here https://github.com/openhpc/ohpc/blob/2.x/components/OHPC_setup_compiler#L59.

So this part is now correct. Thanks for looking.

PR #1581, opened 2022-10-11
Adapt macros for new rpm version

huangtianhua:
Signed-off-by: huangtianhua <huangtianhua@huawei.com>


adrianreber:
Thanks. This is one of the packages which does not work in CI. I will merge it anyway. I probably should add an exclude list for CI as there are a couple of packages which just do not work in CI.

PR #1582, opened 2022-10-12
.gitignore update

jcsiadal:

Minor update to components/.gitignore to reflect changes to downloaded SOURCE files.


adrianreber:
If this file is correctly updated could be a check in CI I think. Maybe run `misc/get_source.sh` and do a git status to see if it shows no untracked files or something like this. Just as a new additional CI check.

PR #1584, opened 2022-10-14
Warewulf 4 update for SUSE builds

jcsiadal:
Should resolve missing dependency in OBS builds. Also adds overlay change for SUSE hosts.

Co-authored-by: mslacken <cgoll@suse.de>

### DIFF
--- a/components/provisioning/warewulf/SPECS/warewulf.spec
+++ b/components/provisioning/warewulf/SPECS/warewulf.spec
@@ -39,6 +39,7 @@ Conflicts: warewulf-provision
 Conflicts: warewulf-ipmi
 
 BuildRequires: make
+BuildRequires: /etc/os-release
 
 %if 0%{?suse_version} || 0%{?sle_version}
 BuildRequires: systemd-rpm-macros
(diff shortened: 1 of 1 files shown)


adrianreber:
OBS does not like the new BuildRequire: https://obs.openhpc.community/package/show/OpenHPC:2.6:Factory/warewulf

The error I see is `nothing provides /etc/os-release`.

jcsiadal:
Are the repos not configured correctly, or is this a bug in OBS? For SUSE, that file is provided by one of many different release-RPMs, but each release-RPM is in a function-specific repo (I'm using the HPC Module repo). Can you verify that we have a repo configured to provide this?

For RHEL, /etc/os-release is a symlink to /usr/lib/os-release, but both are installed with the same RPM. WW4 compiled on Rocky just fine without the "requires". I'm thinking that OBS is bugged, since requiring an absolute filename is a valid spec directive.

I have a workaround, but it's less efficient and very OpenSUSE/RHEL-centric.

adrianreber:
Unfortunately I cannot say much about how OBS is configured and why this fails. It looks reasonable from my point of view. I will merge your new PR and trigger a build to see if OBS likes that better.

PR #1585, opened 2022-10-15
Fix WW4 builds on OBS

jcsiadal:

Replaces the filename requires with distro-specific virtual package names. It still builds without error on local Rocky8, OpenSUSE15.3, and SLE15.4.

Reorganized a few lines in the spec and replaced "define" with "global" for global constants.

### DIFF
--- a/components/provisioning/warewulf/SPECS/warewulf.spec
+++ b/components/provisioning/warewulf/SPECS/warewulf.spec
@@ -10,13 +10,18 @@
 
 %include %{_sourcedir}/OHPC_macros
 
-%define debug_package %{nil}
+%global debug_package %{nil}
 
 # Base package name
-%define pname warewulf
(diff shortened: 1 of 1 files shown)


adrianreber:
It seems like OBS pulls in a package named dummy-release. But it still fails on Leap 15.3.

adrianreber:
Maybe you can try to build it in your OBS home project. The CI in GitHub is just the default Leap 15.3 container. Not sure how OBS is configured differently.

jcsiadal:
Good news: It's not an RPM dependency anymore.
Bad news: It's a WW4 Makefile issue. One of the Makefile targets never runs and I think it's because of a missing .PHONY statement. I can patch and then issue an upstream fix. I'll see if my private OBS is working to test.

PR #1589, opened 2022-10-16
Warewulf 4 fixes for OBS

jcsiadal:

Summary: OBS doesn't fully support golang builds, since golang downloads modules as needed and the OBS virtual machine doesn't allow that. There is a new _service module that would do this during package source staging, but it isn't implemented yet.

Today, one must add all vendor modules to the source tree. That's added here as a tarball in SOURCES.

Successfully builds locally on SLES15.4 and on OBS Rocky and OpenSUSE15.3.


adrianreber:
Sorry, I have not seen this. Will merge now.

It is interesting, however. Most modern Go projects include a complete vendor tree and do not require a download of dependencies. At least all of the container related tools I have seen like containerd, Kubernetes, runc, Podman and so on.

But I guess, this is the wrong place for this discussion. If it fixes the build it should be good.

adrianreber:
Thanks for the fix. It works now for all four combinations.

PR #1593, opened 2022-10-21
EasyBuild v4.6.2

boegel:
https://github.com/easybuilders/easybuild/releases/tag/easybuild-v4.6.2

### DIFF
--- a/components/dev-tools/easybuild/SPECS/easybuild.spec
+++ b/components/dev-tools/easybuild/SPECS/easybuild.spec
@@ -15,7 +15,7 @@
 
 Summary:   Software build and installation framework
 Name:      EasyBuild%{PROJ_DELIM}
-Version:   4.6.0
+Version:   4.6.2
 Release:   1%{?dist}
 License:   GPLv2
 Group:     %{PROJ_NAME}/dev-tools


adrianreber:
Thanks. Just in time for the next release.

If I already have you around. I am currently looking into building OpenHPC on CentOS Stream 9 and EasyBuild is one of the few things that fail to build:
```
[  183s] [[ERROR]] Running 'easy_install --quiet --upgrade --prefix=/tmp/tmps98f59_d/eb_stage1 /home/abuild/rpmbuild/SOURCES/easybuild-framework-4.6.0.tar.gz /home/abuild/rpmbuild/SOURCES/easybuild-easyblocks-4.6.0.tar.gz /home/abuild/rpmbuild/SOURCES/easybuild-easyconfigs-4.6.0.tar.gz' failed: module 'setuptools.command.easy_install' has no attribute 'main'
[  183s] Traceback (most recent call last):
[  183s]   File "/home/abuild/rpmbuild/BUILDROOT/EasyBuild-ohpc-4.6.0-4.1.aarch64/./bootstrap_eb.py", line 380, in run_easy_install
[  183s]     easy_install.main(args)
[  183s] AttributeError: module 'setuptools.command.easy_install' has no attribute 'main'
[  183s] 
[  183s] error: Bad exit status from /var/tmp/rpm-tmp.US9fYM (%build)
```

I saw in the documentation that bootsteap_eb.py is not really supported any more. Do you know if there is anywhere a SPEC file to build an EasyBuild RPM for CentOS Stream 9?

boegel:
@adrianreber I'm not aware of an existing `.spec` file for building an RPM for EasyBuild, but I would really like to get that sorted out...

EasyBuild is now a pretty standard Python package, just install with `pip install`, so the `.spec` should be pretty standard too I guess.
The only "special" aspect is that it still consists of 3 separate Python packages: `easybuild-framework`, `easybuild-easyblocks` and `easybuild-easyconfigs`. The `easybuild` package is a meta-package, which just pulls in the other 3 as dependencies.

Maybe we should follow-up on this in a dedicated issue?

boegel:
I've opened a dedicated issue on this: #1597

PR #1594, opened 2022-10-24
Upgrade ipmiutils to 3.1.8

huangtianhua:
Signed-off-by: huangtianhua <huangtianhua@huawei.com>

### DIFF
--- a/components/distro-packages/ipmiutil/SPECS/ipmiutil.spec
+++ b/components/distro-packages/ipmiutil/SPECS/ipmiutil.spec
@@ -14,7 +14,7 @@
 %define pname ipmiutil
 
 Name:    %{pname}%{PROJ_DELIM}
-Version: 2.9.6
+Version: 3.1.8
 Release: 1%{?dist}
 Summary:   Easy-to-use IPMI server management utilities
 License:   BSD 3-clause
@@ -41,6 +41,7 @@ BuildRequires: gcc gcc-c++ libtool
 %endif
 %define init_dir  %{_initrddir}
 
+BuildRequires: make wget
 %description
 The ipmiutil package provides easy-to-use utilities to view the SEL,
 perform an IPMI chassis reset, set up the IPMI LAN and Platform Event Filter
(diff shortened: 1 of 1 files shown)


adrianreber:
Thanks for you PR, but I am closing this as OpenHPC no longer ships ipmiutils. It is still in our repository but it was never built for the 2.x branch.

huangtianhua:
I proposed this because ipmiutils built error on enviromment which openssl is 1.1.1m, if community has no plan to upgrade, I will make a patch to fix this, thank you.

adrianreber:
> I proposed this because ipmiutils built error on enviromment which openssl is 1.1.1m, if community has no plan to upgrade, I will make a patch to fix this, thank you.

OpenHPC does not built ipmiutils at all. It is an old package. Why do you build it? Do you need it for openeuler?

huangtianhua:
@adrianreber ,
OK, thanks! 
And so we just have to adapt the packages for whole scope as https://github.com/openhpc/ohpc/blob/2.x/misc/obs/config.2.x#L35  right?

adrianreber:
> @adrianreber , OK, thanks! And so we just have to adapt the packages for whole scope as https://github.com/openhpc/ohpc/blob/2.x/misc/obs/config.2.x#L35 right?

Right. That file lists the packages which are actually built. You can also look at https://obs.openhpc.community/

huangtianhua:
Sorry, I distrub you again, https://github.com/openhpc/ohpc/blob/2.x/misc/obs/config.2.x#L35 only show the packages which are updated in 2.6? But IIUC we have to adapt all ohpc packages(the spec files on 2.x) on openEuler, not only updated in 2.6 ? So if there is a list for all packages we have to adapt?

adrianreber:
You have to look at the 2.6 section and at all previous sections. Maybe something like this would also help: `dnf --available --disablerepo='*' --enablerepo='OpenHPC' --enablerepo='OpenHPC-updates' list`

That should give you an overview of all packages available for the 2.x release branch.

Maybe also this command: ` dnf --disablerepo='*' --enablerepo='OpenHPC' --enablerepo='OpenHPC-updates' repoquery --srpm --qf  '%{NAME}'`.

This would be without 2.6. Once 2.6 is released those commands should also include all packages for 2.6.

PR #1595, opened 2022-10-24
Add new oneAPI module gen tool

jcsiadal:

 - Replaces install-only module generator with dedicated utility which can be run any time. Supports non-default and non-RPM installs
 - Updates oneAPI versions
 - Moved a common global macro to OHPC_macros
 - Removes PSXE module generation


components/compiler-families/intel-compilers-devel/SOURCES/oneapi-modgen.lua
```diff
@@ -0,0 +1,937 @@
...
+-- Remove leading directories from file path
+function BaseName(path)
+    return tostring(path:match(".*/([^/]*)$"))
+end
+
+function DirName(path)
+    return tostring(path:match("(.*)/[^/]+$"))
+end
+
+-- Concatonate all arguments into a path
```
martin-g:
```suggestion
-- Concatenate all arguments into a path
```

components/compiler-families/intel-compilers-devel/SOURCES/oneapi-modgen.lua
```diff
@@ -0,0 +1,937 @@
...
+    local handle = assert(io.popen(cmd, 'r'),
+        "Process open error: '" .. cmd .. "'")
+    handle:flush()
+    local cmdoutput = assert(handle:read('*a'),
+        "Process read eoor: '" .. cmd .. "'")
+    -- Command returns STDOUT (minus newline) and if no error
+    return cmdoutput:gsub("\n+$", ""), not handle:close()
+end
+
+-- Convert mult-line string to array
```
martin-g:
```suggestion
-- Convert multi-line string to array
```

components/compiler-families/intel-compilers-devel/SOURCES/oneapi-modgen.lua
```diff
@@ -0,0 +1,937 @@
...
+    local header = [[
+local version = "]] .. version .. [["
+        
+]]
+
+    -- Template for Intel(R) oneAPI compilers
+    if modtype == "oneapi" then
+        return ([[
+help("\nThis module makes all modules for oneAPI Toolkits\n" ..
+  "available to load. It does not local any oneAPI modules\n"
```
martin-g:
```suggestion
  "available to load. It does not load any oneAPI modules\n"
```
?

components/compiler-families/intel-compilers-devel/SOURCES/oneapi-modgen.lua
```diff
@@ -0,0 +1,937 @@
...
+        end
+    end
+    print()
+end
+
+-- Run md5sum against file; return the (first) checksum value
+local function md5sum(filename)
+    local output, iserror = RunOSCommand("md5sum -z " .. filename)
+    if iserror then
+        printformat("Unable to create md5sum for" .. filename, "error")
```
martin-g:
```suggestion
        printformat("Unable to create md5sum for " .. filename, "error")
```

components/compiler-families/intel-compilers-devel/SOURCES/oneapi-modgen.lua
```diff
@@ -0,0 +1,937 @@
...
+                printformat("Bad format: line " .. i, "substep")
+            else
+                if not checklist[c.file] then -- Filename not found
+                    printformat("md5sum read error: line " .. i, "substep")
+                else
+                    local moddir = DirName(c.file)
+                    -- Include only valid versions
+                    if checklist[c.file] == "OK" or
+                        checklist[c.file] == "MOD" then
+                        -- Check that module type is recognozed
```
martin-g:
```suggestion
                        -- Check that module type is recognized
```

jcsiadal:
@martin-g Thanks for the review and corrections. I've added them to the PR.

components/compiler-families/intel-compilers-devel/SPECS/intel-compilers-devel.spec
```diff
@@ -68,145 +65,25 @@ mkdir -p %{buildroot}/%{OHPC_MODULEDEPS}/gnu/mkl
...
-else
-    # Don't touch any generated files if there's no manifest
-    # On upgrade, expect lots of modulefiles created as .rpmnew
-    echo "WARNING: Manifest not found. Previously generated"
-    echo "         modulefiles will not be removed or moved."
-fi
-# Remove the generated oneAPI module links, remove directories if empty
-find %{OHPC_MODULEDEPS}/oneapi -type l -delete
-find %{OHPC_MODULEDEPS}/oneapi/* -empty -type d -delete
+# Remove generated compilers/mkl modules
```
opoplawski, line 75: This should be bracketed with:
```
if [ $1 -eq 0 ]
then
  ...
fi
```
per https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/ to avoid running on upgrades/reinstalls.

opoplawski (changes requested):

components/compiler-families/intel-compilers-devel/SPECS/intel-compilers-devel.spec
opoplawski: This is still an issue.

components/compiler-families/intel-compilers-devel/SOURCES/oneapi-modgen.lua
```diff
@@ -0,0 +1,937 @@
...
+    return output:sub(1, 32)
+end
+
+-- Find the default version
+local function find_default(path, isfullpath)
+    local linkinfo = lfs.symlinkattributes(PathCat(path, "default"))
+    if not linkinfo then
+        return nil
+    end
+    if not lfs.attributes(PathCat(path, linkinfo.target), "mode") then
```
opoplawski, line 513: Apparently the "target" attribute was added in lfs 1.7.0, but EL8 only has lua-filesystem 1.6.3, so this fails for me.

components/compiler-families/intel-compilers-devel/SOURCES/oneapi-modgen.lua
```diff
@@ -0,0 +1,937 @@
...
+    /opt/intel/oneapi is used to locate installed versions. Other
+    installations can be added manually using the root path for that
+    installation.
+    ]]
+end
+
+require "io"
+require "os"
+require "lfs" -- lua-filesystem RPM required
+
```
adrianreber, line 41: I do not see any code directly interacting with the lmod lua code base. This just seems to use three lua modules which are not connected to lmod. I see it problematic introducing a new language into the OpenHPC code base.

The modules `os`, `io` and `lfs` are probably also implementable in something which fits the current repository better.

jcsiadal: Are you offering to convert and maintain in a different language? The reason for LUA is that its only purpose is to support LMOD. It makes even less sense to have Python/Perl support applets for a main script written entirely in LUA.
One other option I can support is to use C and change this to a binary executable.

jcsiadal:
Resubmitting a modified version.

opoplawski:
@jcsiadal has this been submitted yet?  I'm not finding one.

PR #1599, opened 2022-10-29
Change `Source` to `Source0` in mrsh and warewulf

Yikun:
Change `Source` to `Source0` in mrsh.spec otherwise the `No source number 0` error raise

According to https://rpm-packaging-guide.github.io:

> Path or URL to the compressed archive of the upstream source code (unpatched, patches are handled elsewhere). This should point to an accessible and reliable storage of the archive, for example, the upstream page and not the packager’s local storage. If needed, more SourceX directives can be added, incrementing the number each time, for example: Source1, Source2, Source3, and so on.

### DIFF
--- a/components/admin/mrsh/SPECS/mrsh.spec
+++ b/components/admin/mrsh/SPECS/mrsh.spec
@@ -16,21 +16,22 @@
 Name:    %{pname}%{PROJ_DELIM}
 Version: 2.12
 Release: 1%{?dist}
-Epoch: 3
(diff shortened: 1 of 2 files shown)


adrianreber:
Please rebase and this should be good to be merged.

Yikun:
Replaced by:
https://github.com/openhpc/ohpc/pull/1610
https://github.com/openhpc/ohpc/pull/1609

PR #1607, opened 2022-11-18
ci: introduce spellcheck for all spec files

adrianreber:
This also fixes all errors found by spellcheck.

### DIFF
--- a/tests/ci/Makefile
+++ b/tests/ci/Makefile
@@ -1,0 +1,17 @@
+lint: codespell-lint flake8-lint whitespace-lint
+
+codespell-lint:
+	@echo "Running 'codespell' on all spec files"
+	cd ../..; codespell $$(find components -name "*.spec")
+
+flake8-lint:
+	@echo "Running 'flake8' on selected Python files"
+	flake8 ../../tests/ci/check_spec.py
+
+whitespace-lint:
+	@echo "Checking spec files for trailing whitespaces"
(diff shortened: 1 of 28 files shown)


tests/ci/Makefile
```diff
@@ -0,0 +1,17 @@
+lint: codespell-lint flake8-lint whitespace-lint
+
+codespell-lint:
+	@echo "Running 'codespell' on all spec files"
+	cd ../..; codespell $$(find components -name "*spec")
```
martin-g:
```suggestion
	cd ../..; codespell $$(find components -name "*.spec")
```

tests/ci/check_spec.py
```diff
@@ -46,11 +44,19 @@
...
+skip_ci_specs_env = os.getenv('SKIP_CI_SPECS')
+if skip_ci_specs_env:
+    skip_ci_specs = skip_ci_specs_env.rstrip().split(' ')
+
 for spec in sys.argv[1:]:
     if not spec.endswith('.spec'):
         continue
     spec_found = True
-    print("\n--> Scanning spec file %s" % spec)
+    print("--> Scanning spec file %s" % spec)
```
martin-g: Would it be better if line 56 is moved after line 59 ?
I.e. print either `skipping` or `scanning`, not both.

adrianreber: good idea

PR #1636, opened 2022-12-13
ci: also test against CentOS Stream 9

adrianreber:


.github/workflows/validate.yml
```diff
@@ -125,3 +125,23 @@ jobs:
...
+
+  build_on_centos_stream_9:
+    runs-on: ubuntu-latest
+    name: Build on CentOS Stream 9
+    container:
+      image: quay.io/centos/centos:stream9
+    steps:
+    - name: Install git
+      run: dnf -y install git
+    - uses: actions/checkout@v2
```
martin-g: v3 is [available](https://github.com/actions/checkout/tree/v3)
```suggestion
    - uses: actions/checkout@v3
```

.github/workflows/validate.yml
```diff
@@ -125,3 +125,23 @@ jobs:
...
+      with:
+        fetch-depth: 0
```
martin-g: Does it really need all history and all tags to be checked out ?
https://github.com/actions/checkout/blob/main/action.yml#L57

.github/workflows/validate.yml
adrianreber: changed

.github/workflows/validate.yml
adrianreber: not sure, probably just some historic copy and paste. also removed

PR #1640, opened 2022-12-20
Use Ana06/get-changed-files instead of jitterbit/get-changed-files

martin-g:
jitterbit/get-changed-files@v1 often fails with errors like:

```
Base commit: c06f2fe4456f933cafcde33a8a764a2d7bf47635 
Head commit: 55342d52e87b98e9d2d06d2b1c54152d7d4acc02 
Error: The head commit for this pull_request event is not ahead of the base commit. Please submit an issue on this action's GitHub repo. All: components/perf-tools/msr-safe/SPECS/msr-safe.spec Added:
Modified: components/perf-tools/msr-safe/SPECS/msr-safe.spec Removed:
Renamed:
Added or modified: components/perf-tools/msr-safe/SPECS/msr-safe.spec
```

This issue is known since a long time but it seems the action is not maintained anymore:
https://github.com/jitterbit/get-changed-files/issues?q=The+head+commit+for+this+pull_request+event+is+not+ahead

https://github.com/Ana06/get-changed-files is a maintained fork that claims to have fixed this issue: https://github.com/jitterbit/get-changed-files/issues/11#issuecomment-854657301

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -30,7 +30,7 @@ jobs:
       with:
         fetch-depth: 0
     - id: files
-      uses: jitterbit/get-changed-files@v1
+      uses: Ana06/get-changed-files@v2.2.0
     - name: Validate Changes
       run: |
         tests/ci/check_spec.py ${{ steps.files.outputs.added_modified }}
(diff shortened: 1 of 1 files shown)


adrianreber:
Good idea. One thought I had about this is if we should use the custom script for Cirrus CI just everywhere or use the fixed GitHub Actions script like in this change.

Extend `tests/ci/cirrus_get_changed_files.sh` to handle GitHub Actions and probably pushes or do it like in this PR. Extending our own script sounds more work and also unnecessary. So maybe just we should just use this PR.

martin-g:
The new Github action fixed our issues at https://github.com/ohpc-openeuler/ohpc/actions

https://github.com/ohpc-openeuler/ohpc/pull/90

Yikun:
@martin-g Thanks, I'm +1 on this to get GA work, if we still facing many error, we can consider the custom script. : )

Issue #1639, opened 2022-12-19
Upgrade ADIOS component to ADIOS2 (2.8.3)


martin-g:
There are some problems in the build of ADIOS (v1) related to Python2, see https://github.com/openhpc/ohpc/issues/872.

The ADIOS developers [suggest](https://github.com/ornladios/ADIOS/issues/216#issuecomment-1347768036) to upgrade/move to ADIOS2 instead.

martin-g:
I am going to work on a PR for this!

adrianreber:
@mghpcsim do we want to go through the usual submission process for adios2 or do we say it is just an upgrade. On the one hand it is a new component, but it could also be seen as an upgrade.

PR #1642, opened 2022-12-21
Issue #1639: Add initial version of adios2.spec

martin-g:
The rpmlintrc file is 1:1 copy from the `adios` component.
The spec file is a copy of the adios(1) component with the needed changes.
The tests are copied from ADIOS2 test suite and slightly adapted for OpenHPC needs.


martin-g:
The PR is a Draft because I need to test it on `aarch64` too!

Update: The build on `aarch64` passed! Now I will work on adding tests.

components/io-libs/adios2/SPECS/adios2.spec
```diff
@@ -0,0 +1,159 @@
...
+    -DADIOS2_USE_HDF5=ON -DHDF5_ROOT="${HDF5_DIR}" \
+    -DADIOS2_USE_ZeroMQ=OFF \
+    -DADIOS2_USE_Fortran=ON \
+    -DADIOS2_USE_Python=ON \
+    -DADIOS2_USE_SST=ON \
+    -DADIOS2_USE_BZip2=ON \
+    -DBUILD_TESTING=ON \
+    ..
+make
+# make test
```
martin-g, line 100: Should we run the tests here ?
There are 2 issues with them:
1) some tests fail
2) the tests execution takes about an hour

components/io-libs/adios2/SPECS/adios2.spec
```diff
@@ -0,0 +1,159 @@
...
+EOF
+
+%{__mkdir_p} ${RPM_BUILD_ROOT}/%{_docdir}
+
+%files
+%{OHPC_HOME}
+%doc Copyright.txt
+%doc Contributing.md
+%doc LICENSE
+%doc ReadMe.md
```
martin-g, line 165: This is the content of `adios2-build` folder after successful `make`:

```
$ ls -1F
adios2-config.cmake
adios2-config-common.cmake
adios2-config-install.cmake
adios2-config-version.cmake
adios2-c-targets.cmake
adios2-cxx11-targets.cmake
adios2-fortran-targets.cmake
adios2_reorganize_wrapper*
adios2-targets.cmake
bin/
bindings/
bpls.cmake.gen
cmake/
CMakeCache.txt
CMakeFiles/
cmake_install.cmake
compile_commands.json
CTestCustom.cmake
CTestTestfile.cmake
DartConfiguration.tcl
examples/
install_manifest.txt
lib/
lib64/
Makefile
plugins/
Release/
source/
testing/
Testing/
thirdparty/
verinfo.h
```

Do I need to add more entries to `%files` ?

tests/libs/adios2/tests/test_module
```diff
@@ -0,0 +1,111 @@
...
+    if [ -z ${!LIB} ];then
+        flunk "${PKG}_LIB directory not defined"
+    fi
+ 
+    if [ ! -d ${!LIB} ];then
+        flunk "directory ${!LIB} does not exist"
+    fi 
+}
+
+@test "[$testname] Verify (static) library available in ${PKG}_LIB ($LMOD_FAMILY_COMPILER)" {
```
martin-g: Currently this test fails because the [spec](https://github.com/openhpc/ohpc/pull/1642/files#diff-8f0a3c1ac4c3e7830aa6271ebf70c9d8877829c36f21b7243596715be66488eeR87) builds shared/dynamic libraries.
This test is copied from ADIOS1 and adapted to ADIOS2.

Should I update the spec the build static libraries or the test to check for dynamic libraries?

components/io-libs/adios2/SPECS/adios2.spec
martin-g: I decided not to run ADIOS2's tests but to copy few of them and use them as smoke tests in OpenHPC tests suite.

tests/libs/adios2/tests/test_module
martin-g: I've changed the test to check for dynamic libraries.

There is some issue with the build of static libraries. The build passes for OpenMPI4 but fails for MPICH complaining that PHDF5_LIB could not be found. If static libraries are preferred then we will have to debug what is the cause.
Needless to say the build passes fine with both OpenMPI and MPICH on my developer machine.

martin-g:
@adrianreber Could you please trigger the CirrusCI jobs ? Just to have the full build
I've rebased and re-targeted the PR to 3.x

adrianreber:
```
Traceback (most recent call last):
  File "/__w/ohpc/ohpc/tests/ci/spec_to_test_mapping.py", line 261, in <module>
    pkgs += test_map[i][2]
IndexError: list index out of range
```

adrianreber:
This looks almost ready.  Can you reduce the number of commits? Maybe one for the new package and one for the tests? Something like this?

adrianreber:
There are a couple of changes to adios2.spec in the test commit. Can you move those changes to the original adios2.spec commit?

martin-g:
I am not sure how to do this.
The only way that comes to my mind is to open a new PR with just the .spec changes. And then rebase this PR, so that it includes only the test related changes.

adrianreber:
I would probably do an interactive rebase, break between the commits and change both commits to look like I want to.

martin-g:
Learnt something new! Thanks!

Issue #1644, opened 2023-01-01
Can not install ohpc-gnu12-runtimes on Rocky 8


devilkingsatan666:
After issuing:
   dnf install ohpc-gnu12-runtimes.x86_64
I got the following error:
Error: 
  Problem: conflicting requests
     - nothing provides singularity needed by ohpc-gnu12-runtimes-2.6-7.1.ohpc.2.6.x86_64

Interestingly, I have apptainer version 1.1.4-2.el8 installed on my system.

No idea what's going wrong.

Best,
Wu-Cheng Chiang

adrianreber:
The problem is that the package `apptainer` from EPEL used to provide the dependency `singularity` although it is not really singularity. EPEL decided to remove this packaging error https://pagure.io/epel/issue/214. So now, nothing provides `singularity` in EPEL.

Either EPEL needs to introduce the real `singularity` package or OpenHPC needs to do a small update to replace the require on `singularity` with `apptainer`.

devilkingsatan666:
EPEL has singularity-ce-3.10.4-1 available for installlation. Which one will the community choose to be the default one?

Best,
Wu-Cheng Chiang

sjpb:
This is probably way off base but is `singularity-ohpc` suitable to be marked as `provides singularity`? I guess the use of environment modules by that package means maybe not?

---

A temporary fix seems to be to do:
```
dnf install -y https://github.com/apptainer/apptainer/releases/download/v1.1.3/apptainer-1.1.3-1.x86_64.rpm
```

sjpb:
Also - the title of this suggests the problem is narrower than it is - `ohpc-base-compute` also depends on `singularity` in v2.6, which means openhpc can't currently be used to create compute nodes without a workaround.

adrianreber:
> This is probably way off base but is `singularity-ohpc` suitable to be marked as `provides singularity`? I guess the use of environment modules by that package means maybe not?

We decided to not ship a copy of singularity because EPEL and LEAP do already ship one. The apptainer/singularity situation seems to be complicated and I do not understand what is happening.

There are a couple of fixes possible. I would just change the meta package to require apptainer on RHEL based clones and it should be fixed.

> A temporary fix seems to be to do:
> 
> ```
> dnf install -y https://github.com/apptainer/apptainer/releases/download/v1.1.3/apptainer-1.1.3-1.x86_64.rpm
> ```

Thanks.

koomie:
[...]

PR #1645, opened 2023-01-06
admin/meta-packages: use singularity-ce on rhel clones

adrianreber:
Previously EPEL had the apptainer package which had a 'Provides: singularity'. This was removed: https://pagure.io/epel/issue/214

The changes the OpenHPC meta-package to use the package singularity-ce on RHEL clones.

### DIFF
--- a/components/admin/meta-packages/SPECS/meta-packages.spec
+++ b/components/admin/meta-packages/SPECS/meta-packages.spec
@@ -86,7 +86,6 @@ Requires:  libicu
 Requires:  libunwind
 Requires:  numactl
 Requires:  python3
-Requires:  singularity
 %if 0%{?rhel}
 Requires:  cairo-devel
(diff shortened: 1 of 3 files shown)


adrianreber:
This tries to solve #1644 

@mghpcsim @koomie Do we want a fix like this for the current problems as reported via the mailing list, slack and GitHub?

martin-g (approved):
LGTM

adrianreber:
I changed the `Requires:`  to a `Recommends:` of either `singularity-ce` or `apptainer` or `singularity`.

This way, if either `singularity-ce` or `apptainer` or `singularity` is available it will be installed. If it is not available `dnf` will just continue. If the user has already one of the package pre-installed it should also just work.

Using `Recommends:` also offers the possibility that the user can uninstall the package later if needed.

PR #1648, opened 2023-01-10
Enable openEuler 22.03 based CI on aarch64

adrianreber:


adrianreber:
@martin-g I was thinking about something like this.

The next step would be to open a PR with the lmod fix you had in your original PR and see if that fixes the OBS build error we currently see for lmod.

adrianreber:
With this PR merged the potential lmod fix would run through CI before being merged.

Yikun:
> The next step would be to open a PR with the lmod fix

@martin-g FYI, according to current lmod failure in OBS, it  should be due to missing `lua-posix`, there are two solution:

- add it to lmod.spec buildrequire.
- or add to openEuler_22.03 preinstall


@adrianreber BTW, a question about preinstall package which packages should be added in:

https://obs.openhpc.community/projects/openEuler:22.03/prjconf

is there any rule of OHPC recommended?

adrianreber:
> > The next step would be to open a PR with the lmod fix
> 
> @martin-g FYI, according to current lmod failure in OBS, it should be due to missing `lua-posix`, there are two solution:
> 
>     * add it to lmod.spec buildrequire.

The big PR from you and @martin-g already have a fix, but instead of merging everything I would prefer to do it step by step.

```diff
 BuildRequires: tcl-devel
 BuildRequires: gcc make bc
 
-%if 0%{?rhel}
+%if 0%{?rhel} || 0%{?openEuler}
 BuildRequires: lua-libs
 BuildRequires: lua-filesystem
 BuildRequires: lua-posix
```

>     * or add to openEuler_22.03 preinstall

No, see above.

> @adrianreber BTW, a question about preinstall package which packages should be added in:
> 
> https://obs.openhpc.community/projects/openEuler:22.03/prjconf
> 
> is there any rule of OHPC recommended?

No, I tried to start with a minimal working list of packages. I prefer to not add anything that is not needed.

martin-g (approved):

Yikun:
@adrianreber Thanks! sounds good to me! And much thanks for your help!

PR #1649, opened 2023-01-10
Fix broken compute node builds on Rocky

jcsiadal:

Remove singularity from the metafiles. Rocky Linux doesn't provide singularity (it provides apptainer), so Rocky builds are broken unless additional repos are added.

Main issue here is the compute node metafile should not be forcing a specific container system on the sysadmin.

### DIFF
--- a/components/admin/meta-packages/SPECS/meta-packages.spec
+++ b/components/admin/meta-packages/SPECS/meta-packages.spec
@@ -74,7 +74,6 @@ Requires:  NetworkManager
 %if 0%{?suse_version}
 Requires:  glibc-locale
 Requires:  nfs-kernel-server
-Requires:  singularity
 %endif
 %description -n %{PROJ_NAME}-base
 Collection of base packages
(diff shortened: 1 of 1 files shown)


martin-g:
This PR seems related to https://github.com/openhpc/ohpc/pull/1645

adrianreber:
> Remove singularity from the metafiles. Rocky Linux doesn't provide singularity (it provides apptainer), so Rocky builds are broken unless additional repos are added.

See #1645 for details why this error happened. This has nothing to do with Rocky Linux but is coming from EPEL and enabling EPEL is required for the use of OpenHPC. So no additional repositories are required.

> Main issue here is the compute node metafile should not be forcing a specific container system on the sysadmin.

I changed it to use `Recommends: (singularity-ce or singularity or apptainer)` for greater flexibility which implementation the user wants to install and to also not fail if it is not available.

adrianreber:
Closing as #1645 has been merged.

PR #1652, opened 2023-01-12
Add support for openEuler for Slurm

martin-g:

### DIFF
--- a/components/admin/genders/SPECS/genders.spec
+++ b/components/admin/genders/SPECS/genders.spec
@@ -27,8 +27,12 @@ BuildRequires: perl(ExtUtils::MakeMaker)
 BuildRequires: python3
 BuildRequires: python3-devel
 
-%if 0%{?rhel_version}
+%if 0%{?rhel_version} || 0%{?openEuler}
 BuildRequires: byacc
+%endif
+
+%if 0%{?openEuler}
+BuildRequires: perl-devel
 %endif
 
 Provides: %{pname} = %{version}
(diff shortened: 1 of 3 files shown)


components/rms/slurm/SPECS/slurm.spec
```diff
@@ -152,6 +152,11 @@ BuildRequires: pkgconfig
 %endif
 %endif
 
+%if 0%{?openEuler}
+BuildRequires: perl perl-devel
+Requires: perl perl-devel
```
adrianreber: This line looks unnecessary. The is usually picked up automatically

martin-g: Removed the `Requires`!

components/rms/slurm/SPECS/slurm.spec
```diff
@@ -152,6 +152,10 @@ BuildRequires: pkgconfig
 %endif
 %endif
 
+%if 0%{?openEuler}
+BuildRequires: perl perl-devel
```
adrianreber: perl-devel already requires perl, so perl is also unnecessary

components/admin/genders/SPECS/genders.spec
```diff
@@ -31,6 +31,10 @@ BuildRequires: python3-devel
 BuildRequires: byacc
```
Yikun, line 31: `%if 0%{?rhel_version} || 0%{?openEuler}` for byacc

martin-g: Improved it!

components/rms/slurm/SPECS/slurm.spec
martin-g: Removed it!

PR #1653, opened 2023-01-13
Execute the tests for openEuler on CirrusCI

martin-g:
Cat the log files in case of failure.

### DIFF
--- a/tests/ci/setup_slurm_and_run_tests.sh
+++ b/tests/ci/setup_slurm_and_run_tests.sh
@@ -90,6 +90,24 @@ fi
 
 export SIMPLE_CI=1
 
+set +e
+
+sudo --user="${USER}" --login bash -c 'cd ${PWD}/tests; find ./ -name "*.log" -delete'
+
 # Always running at least with '--enable-modules'. No need to check for
 # an empty TESTS array.
-sudo --user="${USER}" --preserve-env=SIMPLE_CI --login bash -c "cd ${PWD}/tests; ./bootstrap; ./configure --disable-all --enable-modules --enable-rms-harness --with-mpi-families='openmpi4 mpich' ${TESTS[*]}; make check"
+if sudo --user="${USER}" --preserve-env=SIMPLE_CI --login bash -c "cd ${PWD}/tests; ./bootstrap; ./configure --disable-all --enable-modules --enable-rms-harness --with-mpi-families='openmpi4 mpich' ${TESTS[*]}; make check";
+then
+    exit 0
+fi
+
+set -e
+
+# If we are here, the tests failed. Print the logs.
+echo -e "\nThe tests execution failed. Printing the logs.\n"
+find ./ -name "*.log" -print0 | while IFS= read -r -d '' log_file
+do
+	echo "================================================"
+	echo "Log file: ${log_file}";
+	cat "${log_file}";
+done
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -33,4 +33,8 @@ openeuler_task:
   script: uname -a
   setup_script: tests/ci/prepare-ci-environment.sh
   build_script: |
+    . /etc/profile.d/lmod.sh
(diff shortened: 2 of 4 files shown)


tests/ci/setup_slurm_and_run_tests.sh
```diff
@@ -90,6 +90,21 @@ fi
 
 export SIMPLE_CI=1
 
+sudo --user="${USER}" --login bash -c 'cd ${PWD}/tests; find ./ -name "*.log" -delete'
+
 # Always running at least with '--enable-modules'. No need to check for
 # an empty TESTS array.
 sudo --user="${USER}" --preserve-env=SIMPLE_CI --login bash -c "cd ${PWD}/tests; ./bootstrap; ./configure --disable-all --enable-modules --enable-rms-harness --with-mpi-families='openmpi4 mpich' ${TESTS[*]}; make check"
+
+tests_status=$?
```
adrianreber: Not sure that this will work. The script is running with `set -e` which will result in an exit already here if the return code is not zero. So maybe a `set +e` is necessary before running the test?

tests/ci/setup_slurm_and_run_tests.sh
```diff
@@ -90,6 +90,21 @@ fi
...
+if [ "${tests_status}" -ne 0 ]; then
```
adrianreber: Maybe do a
```shell
if [ "${tests_status}" -eq 0 ]; then
        exit 0
fi
```
here and the log dumping outside of the `if`.

martin-g: Good suggestion! Done!

martin-g:
Latest [error](https://github.com/openhpc/ohpc/pull/1653/checks?check_run_id=10667174610):
```
Error: Unable to find a match: openmpi4-gnu12-ohpc mpich-gnu12-ohpc
```
`prun-ohpc` is available with https://github.com/openhpc/ohpc/pull/1654!

adrianreber:
> Latest [error](https://github.com/openhpc/ohpc/pull/1653/checks?check_run_id=10667174610):
> 
> ```
> Error: Unable to find a match: openmpi4-gnu12-ohpc mpich-gnu12-ohpc
> ```
> 
> `prun-ohpc` is available with #1654!

Now another error happens:
```
+ scontrol update 'nodename=c[0-1]' state=idle
+ break
+ srun -N2 hostname
srun: error: c0: task 0: Exited with exit code 2
srun: error: c1: task 1: Exited with exit code 2
slurmstepd-c1: error: execve(): hostname: No such file or directory
slurmstepd-c0: error: execve(): hostname: No such file or directory
Exit status: 2
```

martin-g:
Yes, I noticed `hostname` related warning/error in the builds of previous specs but until now it was not really a stopper.
Now the bigger problem with `hostname`:
- if we pre-install it with `dnf install hostname` then the next problem is that `hostname` command returns auto-generated/Docker value like `9507b7e1f620` but again Slurm complains that it cannot get its address.
- until now for our testing we were using a hack: 
```
cat > /usr/bin/hostname
#!/usr/bin/env bach
echo "openEuler-arm64"
Ctrl+d
chmod +x /usr/bin/hostname
```
This works just fine!

Any other ideas how to solve it properly ?

adrianreber:
> Yes, I noticed `hostname` related warning/error in the builds of previous specs but until now it was not really a stopper. Now the bigger problem with `hostname`:
> 
>     * if we pre-install it with `dnf install hostname` then the next problem is that `hostname` command returns auto-generated/Docker value like `9507b7e1f620` but again Slurm complains that it cannot get its address.

Hmm, strange. I do not see that. `/etc/hosts` is populated correctly and ping also seems to work in the container with the result of `hostname`

Can you install it so that I can see the error message.

martin-g:
All seems to be fine with `dnf install hostname`! No need of hacks!

adrianreber:
> All seems to be fine with `dnf install hostname`! No need of hacks!

Nice. First tests finally running!

PR #1654, opened 2023-01-13
Prepare macros to handle 3.x branch

adrianreber:


components/OHPC_macros
```diff
@@ -28,7 +28,11 @@
 %global OHPC_UTILS      %{OHPC_PUB}/utils
 %global debug_package   %{nil}
 
+%if 0%{?rhel} >= 9 || || 0%{?openEuler}
```
martin-g: double `||`

adrianreber: thanks, fixed

martin-g (approved):

PR #1656, opened 2023-01-16
admin/powerman: adapt admin pkg for openEuler

Yikun:


adrianreber:
powerman has not been built in OpenHPC 2.x and I do not think we will include it in 3.x. I can still merge it if you want, but it is not necessary.

Yikun:
@adrianreber Emm, it's better to have before we drop it completely. Maybe also a good reference to downstream. Thanks!

PR #1659, opened 2023-01-16
mpi-families: Adapt mpi related pkgs for openEuler

Yikun:
- Adapt mvapich2/openmpi for openEuler
- Add perl build requires for mpich (mpich-ucx-gnu12) to solve `*** No rule to make target 'include/mplconfig.h', needed by 'all-am'.  Stop.`

### DIFF
--- a/components/mpi-families/libfabric/SPECS/libfabric.spec
+++ b/components/mpi-families/libfabric/SPECS/libfabric.spec
@@ -28,7 +28,7 @@ BuildRequires: gcc make
 Buildrequires: ofed
 BuildRequires: rdma-core-devel infiniband-diags-devel
 %endif
-%if 0%{?rhel}
+%if 0%{?rhel} || 0%{?openEuler}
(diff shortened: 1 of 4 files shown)


Yikun:
```
No matching package to install: 'gnu12-compilers-ohpc >= 12.1.0'
No matching package to install: 'libfabric-ohpc'
Package rdma-core-devel-41.0-3.el9.x86_64 is already installed.
Package libpsm2-devel-11.2.229-2.el9.x86_64 is already installed.
Package make-1:4.3-7.el9.x86_64 is already installed.
Package ohpc-buildroot-2.7-300.ohpc.1.7.noarch is already installed.
No matching package to install: 'openpbs-devel-ohpc'
No matching package to install: 'openpbs-server-ohpc'
No matching package to install: 'ucx-ohpc'
Not all dependencies satisfied
Error: Some packages could not be found.
```

CentOS 9 error seems come from OBS x86 infra problems.

PR #1661, opened 2023-01-26
Update Warewulf to 4.4 (3.x)

jcsiadal:

Adds many additional usability features and fixes.

This updated spec resolves an issue with newer Rocky builds, where tftp directory is hardcoded, but keeps layout the same as older versions and SUSE builds.


jcsiadal:
Looks like openEuler CI failed on a repo timeout.

adrianreber:
I do not think we should include another compressed tar archive in our git repository.

I did not look closely enough the last time, but we should host the vendor archive somewhere else. Not sure where. Maybe on the repository server.

It could be copied to the OBS server, but that makes it hard to build the package somewhere else. So something with public reachable URL would be nice.

@koomie @mghpcsim can we use a directory on the repository server for tar archives which have no canonical upstream?

jcsiadal:
I don't like it either, specifically because of possible license issues. I was looking at solutions this morning, and it seems that JIT dependencies are a growing issue. It's not just our build system, but Mock and others.
Kube and Docker are both written in Go. Looking at Fedora's spec for those: they're pulling a Git repo commit# (which includes a populated vendor directory) instead of the source package. I honestly don't see that as any better than a vendor tarball--the SRPMs end up with the same vendor-licensed code.
We can implement the Golang OBS module, which would solve our builds. Local builds with Internet access work fine without including any vendor code.

jcsiadal:
There was also a thought about using "BuildRequires: go(somemodule)" for this, similar to Perl/Python. A poster noted that Google has plans for Go that will break that model.

adrianreber:
@jcsiadal I had a look at this and the `go_vendor` tarball seems unnecessary. All files are already part of the original tarball as downloaded from warewulf 4.4.0? I can build 4.4.0 without network access and without `go_vendor`.

Looking at the upstream 4.4.0 tarball I see all files which are also part of the `go_vendor` tarball.

Also, the BuildRequires on git is also not necessary. At least it works without it.

Please drop the `BR: git`, drop the `go_vendor` tarball and rebase and target the 3.x branch

mslacken:
As there are no use cases for `wwapi[crd]` these should go to a separate package.  Also there is no upstream documentation how to set these services up in a secure way.

jcsiadal:
@adrianreber - I'll get this updated.
@mslacken - OpenHPC normally combines everything into one package, client/server stuff excepted. Would your recommendation be to just drop the wwapi from the RPM for now?

mslacken:
@jcsiadal Yes, as long as there is no use case and no proper documentation, I would drop it.

jcsiadal:
Rebased to 3.x. Removed wwapi and vendor tarball.

adrianreber:
Looks good. CI is happy.

@mslacken any additional comments from your side. I would merge this now.

mslacken:
@adrianreber looks good.

adrianreber:
Builds available at: https://obs.openhpc.community/package/show/OpenHPC:3.0:Factory/warewulf

PR #1664, opened 2023-01-27
Upgrade spack and add admin tests cases to CI

adrianreber:


adrianreber:
@martin-g this enables admin tests like spack. PTAL.

martin-g (approved):
Awesome!

tests/ci/spec_to_test_mapping.py
```diff
@@ -2,153 +2,197 @@
...
     ],
     'components/mpi-families/mpich/SPECS/mpich.spec': [
         'mpich',
+        '',
         'mpich-gnu12-ohpc'
     ],
+    'components/dev-tools/spack/SPECS/spack.spec': [
+        '',
+        'spack',
+        ''
```
Yikun, line 164:
```suggestion
        'gnu12-compilers-ohpc'
```

Maybe add gnu12? because:
https://github.com/openhpc/ohpc/blob/ba053afc4eed0a71e31f9b40ab3b6e7949108a8d/tests/admin/spack/spack#L24

Yikun (commented):
LGTM! Thanks

tests/ci/spec_to_test_mapping.py
adrianreber: I did not include `gnu12-compilers-ohpc` because the openmpi4 and mpich packages are always installed in the test environment. No need to list it twice. Not sure. I would be totally wrong to list it here, but also not necessary.

PR #1665, opened 2023-01-30
Change upload-artifact action retention-days to 1

Yikun:
Change upload-artifact action retention-days to `1` to avoid over the spending limit.


Yikun:
cc @adrianreber

adrianreber:
So, CI is working again. At this point it is not clear why. I deleted a couple of artifacts manually. Maybe that has helped.

This PR makes sense, so I am merging this and hope that this helps with the resource problem we have been seeing in CI.

PR #1676, opened 2023-02-03
boost: Adapt distro pkgs for openEuler

Yikun:
- Add make build requires
- Adapt distro pkgs for openEuler
- Add test mapping for boost

### DIFF
--- a/tests/ci/run_build.py
+++ b/tests/ci/run_build.py
@@ -172,6 +172,10 @@ def build_srpm_and_rpm(command, family=None):
     if family is not None:
         rebuild_command[-1] += " --define 'mpi_family %s'" % family
 
+    # Disable parallel builds for boost on aarch64 to avoid OOM
+    if "boost-" in src_rpm and os.uname().machine == "aarch64":
+        rebuild_command[-1] += " --define '_smp_mflags -j1'"
+
     success, _ = run_command(rebuild_command)
     if not success:
         logging.error("Running 'rpmbuild --rebuild' failed")
--- a/tests/ci/spec_to_test_mapping.py
+++ b/tests/ci/spec_to_test_mapping.py
@@ -178,6 +178,11 @@ test_map = {
         '',
         'cmake-ohpc'
     ],
+    'components/parallel-libs/boost/SPECS/boost.spec': [
+        'boost',
+        '',
+        ''
+    ],
 }
 
 
--- a/components/parallel-libs/boost/SPECS/boost.spec
+++ b/components/parallel-libs/boost/SPECS/boost.spec
@@ -43,6 +43,7 @@ BuildRequires:  libquadmath-devel
 %endif
 %endif
 
+BuildRequires:  make
 BuildRequires:  fdupes
 BuildRequires:  dos2unix
 BuildRequires:  gmp-devel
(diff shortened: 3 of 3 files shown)


Yikun:
aarch64 error seems due to [Container errored with 'OOMKilled'](https://cirrus-ci.com/task/5234443007295488).

tests/ci/spec_to_test_mapping.py
```diff
@@ -178,6 +178,11 @@
         '',
         'cmake-ohpc'
     ],
+    'components/dev-tools/cmake/SPECS/boost.spec': [
+        'boost',
+        '',
+        'boost-gnu12-openmpi4-ohpc'
```
martin-g: don't we need similar for `mpich` ?

adrianreber: Yes, also, but the mapping is wrong. It contains cmake in the path of the spec file.

The packages tested should not be listed here, because that breaks the test. So in this case it should be empty. This is is only for additional packages which are not pulled in by RPM requires.

If you list the package here, the script will first install the results from the CI build and then for some reason DNF will downgrade the packages here to the version from the repository.

If you look at the log file you should see that the tests are not running and that boost has been downgraded.

adrianreber: The downgrading cannot be seen because of the wrong mapping.

Yikun: Thanks for expanation and review, it should be:

```suggestion
        'openmpi4-gnu12-ohpc mpich-gnu12-ohpc'
```

adrianreber: No need to specify those packages. They are anyway installed https://github.com/openhpc/ohpc/blob/2.x/tests/ci/setup_slurm_and_run_tests.sh#L18 and the packages should be installed as a dependency anyway.

adrianreber:
I guess boost CI must be disabled on aarch64 for now. There is never enough memory to have it work with being OOM killed. We could try to disable parallel builds for boost on aarch64 my somehow setting `%{?_smp_mflags}` to one just for boost. Not sure it is worth the effort. At this point we could try with something like if uname -m equals aarch64 and spec file is boost.spec we could add `--define _smp_mflags 1`. Would be worth a try. Probably in the Python script that handles the rebuilds.

tests/ci/run_build.py
```diff
@@ -172,6 +172,10 @@ def build_srpm_and_rpm(command, family=None):
     if family is not None:
         rebuild_command[-1] += " --define 'mpi_family %s'" % family
 
+    # Disable parallel builds for boost on aarch64 to avoid OOM
```
Yikun, line 175: > At this point we could try with something like if uname -m equals aarch64 and spec file is boost.spec we could add --define _smp_mflags 1. Would be worth a try. Probably in the Python script that handles the rebuilds.

@adrianreber  Is that so? Although I didn't quite understand the meaning of _smp_mflags, I tried it.

But looks like build failure due to `_smp_mflags` related changes.

tests/ci/spec_to_test_mapping.py
Yikun: IIUC, L154/L159 should be empty?
https://github.com/openhpc/ohpc/blob/fb703baa96b061dcce989ff6fd284fa78bea2414/tests/ci/spec_to_test_mapping.py#L151-L160

and all deps like L149 should also be removed, because it's redundant:

https://github.com/openhpc/ohpc/blob/fb703baa96b061dcce989ff6fd284fa78bea2414/tests/ci/spec_to_test_mapping.py#L149

right? If so, I'd like to submit a separate PR to address them.

tests/ci/run_build.py
adrianreber: You need to set it to `-j1` and not just `1`.

Yikun: Then I understand! lol!

adrianreber:
Nice, that worked!

PR #1678, opened 2023-02-03
ptscotch: Adapt distro pkgs for openEuler

Yikun:


martin-g (commented):
We should add test mapping for ptscotch

Yikun:
Already added before:
https://github.com/openhpc/ohpc/blob/59030dc7be4b3ac1a0a5d0906ac89873df65d292/tests/ci/spec_to_test_mapping.py#L46

martin-g (approved):
Now I remember that I added it few weeks ago.

PR #1683, opened 2023-02-08
dimemas: Adapt pkg for openEuler

Yikun:

### DIFF
--- a/tests/ci/spec_to_test_mapping.py
+++ b/tests/ci/spec_to_test_mapping.py
@@ -198,6 +198,11 @@ test_map = {
         '',
         ''
     ],
+    'components/perf-tools/dimemas/SPECS/dimemas.spec': [
+        'dimemas',
+        '',
+        'lmod-defaults-gnu12-openmpi4-ohpc'
+    ],
 }
 
 
--- a/components/perf-tools/dimemas/SPECS/dimemas.spec
+++ b/components/perf-tools/dimemas/SPECS/dimemas.spec
@@ -28,7 +28,7 @@ Source0:	https://ftp.tools.bsc.es/dimemas/dimemas-%{version}-src.tar.bz2
 BuildRequires: boost-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 Requires:      boost-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 BuildRequires: bison make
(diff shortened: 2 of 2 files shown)


Yikun:
```
not ok 1 [Dimemas] Verify dimemas module is loaded and matches rpm version (gnu12/mpich)
# (from function `module' in file /opt/ohpc/admin/lmod/lmod/init/bash, line 156,
#  from function `setup' in test file test_module, line 19)
#   `module load ohpc' failed
# Lmod has detected the following error: The following module(s) are unknown:
# "ohpc"
# 
# Please check the spelling or version number. Also try "module spider ..."
# It is also possible your cache file is out-of-date; it may help to try:
#   $ module --ignore_cache load "ohpc"
# 
# Also make sure that all modulefiles written in TCL start with the string
# #%Module
```

Any idea about above error? @adrianreber

tests/ci/spec_to_test_mapping.py
```diff
@@ -198,6 +198,11 @@
         '',
         ''
     ],
+    'components/perf-tools/dimemas/SPECS/dimemas.spec': [
+        'dimemas',
+        '',
+        'lmod-defaults-gnu12-openmpi4-ohpc'
```
martin-g, line 204: Why only the `openmpi4` flavor is needed here ?
The tests are executed also for `mpich`. And they pass without the dependency.

Yikun: Frankly, I referenced:

https://github.com/openhpc/ohpc/blob/222d7931e7df61b1aeacc4e975e2b859dca3af82/tests/ci/spec_to_test_mapping.py#L149

I also remember I resolved the  `The following module(s) are unknown: "ohpc"` in my local env by installing lmod-defaults-gnu12-openmpi4-ohpc. (Give a installs deafult)

adrianreber: > Why only the `openmpi4` flavor is needed here ?
> The tests are executed also for `mpich`. And they pass without the dependency.

I think this is necessary if there is a `module load ohpc` somewhere. There is the possibility to set a default compiler MPI combination with the help of the defaults package.

I am not sure why it is loaded in the test, but to solve the errors during `module load ohpc` one of the defaults needs to be installed.

That is also how I worked around this.

PR #1684, opened 2023-02-08
paraver: Adapt pkg for openEuler

Yikun:
Adapt paraver pkg for openEuler and disable parallel builds on aarch64 to avoid OOM.


Yikun:
```
Container errored with 'OOMKilled'
```

The previous OOM error link: https://cirrus-ci.com/task/5065161803825152

PR #1685, opened 2023-02-08
munge: Adapt pkg for openEuler

Yikun:


Yikun:
```
No match for argument: pdsh-mod-slurm-ohpc
No match for argument: pdsh-ohpc
```

adrianreber:
We use munge from the distro. You can close this.

PR #1687, opened 2023-02-08
R: Adapt pkg to openEuler

Yikun:

### DIFF
--- a/components/serial-libs/R/SPECS/R.spec
+++ b/components/serial-libs/R/SPECS/R.spec
@@ -31,6 +31,7 @@ Group:          %{PROJ_NAME}/serial-libs
 # Default library install path
 %define         install_path %{OHPC_LIBS}/%{compiler_family}/%{pname}%{OHPC_CUSTOM_PKG_DELIM}/%version
 
+BuildRequires:  make
 BuildRequires:  cairo-devel
 BuildRequires:  libjpeg-devel
 BuildRequires:  libpng-devel
@@ -44,14 +45,14 @@ BuildRequires:  xz-devel
 BuildRequires:  pcre2-devel
 BuildRequires:  libcurl-devel
 BuildRequires:  tk-devel
-%if 0%{?rhel}
+%if 0%{?rhel} || 0%{?openEuler}
 BuildRequires:  libXt-devel
 BuildRequires:  texinfo >= 5.1
 BuildRequires:  bzip2-devel
 %endif
 Requires:       cairo >= 1.2
 Requires:       fontconfig
-Requires:       freetype2
+Requires:       freetype
 Requires:       make
 Requires:       readline
 Requires:       xdg-utils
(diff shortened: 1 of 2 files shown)


Yikun:
`No matching package to install: 'openblas-gnu12-ohpc'`

adrianreber:
`make: command not found`

Yikun:
> make: command not found

@adrianreber Done! Thanks

adrianreber:
Please add an entry to the test mapping file.

components/serial-libs/R/SPECS/R.spec
```diff
@@ -44,14 +45,18 @@ BuildRequires:  xz-devel
...
-%if 0%{?rhel}
+%if 0%{?rhel} || 0%{?openEuler}
 BuildRequires:  libXt-devel
 BuildRequires:  texinfo >= 5.1
 BuildRequires:  bzip2-devel
 %endif
 Requires:       cairo >= 1.2
 Requires:       fontconfig
+%if 0%{?openEuler}
+Requires:       freetype
```
adrianreber, line 55: I had a quick look and it seems the package is called `freetype` everywhere. This is really strange that it worked before. Please remove the `%if` and just requires `freetype.

Yikun: Yes, just a double confirm in openEuler/Rocky only `freetype` now, but I'm not sure `freetype2` is required by SUSE:

```
$ docker run -ti  registry.opensuse.org/opensuse/leap:15.3 bash
$ zypper  search freetype
S | Name                    | Summary                                                             | Type
--+-------------------------+---------------------------------------------------------------------+-----------
  | freetype                | TrueType Font Engine                                                | package
  | freetype-devel          | Development files for the TrueType Font Engine                      | package
  | freetype-tools          | Bundled Tests, Demos and Tools for FreeType  (Needed for CJK-LaTeX) | package
  | freetype2               | A TrueType Font Library                                             | srcpackage
  | freetype2-devel         | Development environment for the freetype2 TrueType font library     | package
  | freetype2-devel-32bit   | Development environment for the freetype2 TrueType font library     | package
  | freetype2-profile-tti35 | Set TrueType interpreter to version 35                              | package
```

So, your mean was just change `Requires: freetype2` to `Requires: freetype`?

adrianreber: > Yes, just a double confirm in openEuler/Rocky only `freetype` now, but I'm not sure `freetype2` is required by SUSE:
> 

`freetype2` in Leap seems to be some kind of meta package. If you try to install it the same libraries are installed as if you install `freetype`.

> So, your mean was just change `Requires: freetype2` to `Requires: freetype`?

Yes, that sounds correct. I am not sure if `Requires: freetype` is needed at all, but let's just leave it in for now. `freetype2` is definitely wrong.

PR #1697, opened 2023-02-13
parallel-libs/trilinos: add make

Yikun:
- Github Action splits multipleline ENV with `\n` now
- Add make
- Skip trilinos build/test

### DIFF
--- a/tests/ci/run_build.py
+++ b/tests/ci/run_build.py
@@ -28,13 +28,7 @@ reader = csv.DictReader(open('/etc/os-release'), delimiter="=")
 skip_ci_specs = []
 skip_ci_specs_env = os.getenv('SKIP_CI_SPECS')
 if skip_ci_specs_env:
-    if os.getenv('CIRRUS_CI'):
-        # CIRRUS CI splits multi-line evinronment variables
-        # with a '\n'.
-        skip_ci_specs = skip_ci_specs_env.rstrip().split('\n')
-    else:
-        # GitHub Actions with a space.
-        skip_ci_specs = skip_ci_specs_env.rstrip().split(' ')
+    skip_ci_specs = skip_ci_specs_env.rstrip().split('\n')
 
 for row in reader:
     key = row.pop('NAME')
--- a/tests/ci/spec_to_test_mapping.py
+++ b/tests/ci/spec_to_test_mapping.py
@@ -4,6 +4,7 @@
 # and which tests to enable in the test suite.
 # This script will return three shell arrays (TESTS, ADMIN_TESTS and PKGS).
 
+import os
 import sys
 
 # This dictionary defines the mapping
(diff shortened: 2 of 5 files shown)


Yikun:
@adrianreber It is timeout in all GA CI, could we only test it in OBS?

martin-g:
The build at OBS passed successfully: https://obs.openhpc.community/package/live_build_log/home:mgrigorov/trilinos-gnu12-mpich/OpenHPC_3.0_Factory_openEuler_22.03/aarch64

adrianreber:
> @adrianreber It is timeout in all GA CI, could we only test it in OBS?

The package takes forever to build. That is true. There are exclude lists in different places for the different CIs. You can list the spec file there and then it should be skipped.

tests/ci/spec_to_test_mapping.py
```diff
@@ -118,11 +118,6 @@
         '',
         ''
     ],
-    'components/parallel-libs/trilinos/SPECS/trilinos.spec': [
```
martin-g: Isn't this mapping still useful for local runs ?

Yikun: Good point!

martin-g (approved):

tests/ci/run_build.py
```diff
@@ -28,13 +28,7 @@
...
 skip_ci_specs_env = os.getenv('SKIP_CI_SPECS')
 if skip_ci_specs_env:
-    if os.getenv('CIRRUS_CI'):
-        # CIRRUS CI splits multi-line evinronment variables
-        # with a '\n'.
-        skip_ci_specs = skip_ci_specs_env.rstrip().split('\n')
-    else:
-        # GitHub Actions with a space.
-        skip_ci_specs = skip_ci_specs_env.rstrip().split(' ')
+    skip_ci_specs = skip_ci_specs_env.rstrip().split('\n')
```
adrianreber, line 31: I am curious about this change. Has the way CIRRUS_CI splits multi-line environment variables changed.

Yikun: Actually, it should be the change of github action, CIRRUS_CI always use the `\n`:

I [printed](https://github.com/openhpc/ohpc/commit/aa200b56c2ee9113d090b7edb71a67dc7aa22788) the skip_ci_specs_env: 

```
print(skip_ci_specs_env)
print(skip_ci_specs_env.rstrip().split(' '))
print(skip_ci_specs_env.rstrip().split('\n'))
```

https://github.com/openhpc/ohpc/actions/runs/4239256449/jobs/7367128993#step:8:20

```
components/fs/lustre-client/SPECS/lustre.spec
components/parallel-libs/trilinos/SPECS/trilinos.spec

['components/fs/lustre-client/SPECS/lustre.spec\ncomponents/parallel-libs/trilinos/SPECS/trilinos.spec']
['components/fs/lustre-client/SPECS/lustre.spec', 'components/parallel-libs/trilinos/SPECS/trilinos.spec']
```

`skip_ci_specs_env` is `components/fs/lustre-client/SPECS/lustre.spec\ncomponents/parallel-libs/trilinos/SPECS/trilinos.spec`, so the space style seems outdate or it was always wrong.

adrianreber: okay, thanks for the clarification

PR #1698, opened 2023-02-13
serial-libs/superlu: Add make and openblas

Yikun:


Yikun:
```
-------------------------------------------------------
Libraries: SuperLU tests: gnu12
-------------------------------------------------------
Lmod has detected the following error: The following module(s) are unknown:
"openblas"

Please check the spelling or version number. Also try "module spider ..."
It is also possible your cache file is out-of-date; it may help to try:
  $ module --ignore_cache load "openblas"

Also make sure that all modulefiles written in TCL start with the string
#%Module



FAIL libs/superlu/ohpc-tests/test_compiler_families (exit status: 1)
```

adrianreber:
The change in the mapping file helps but unfortunately it is not totally correct. The right way to do this is to fix the spec file. You need to add something like this:
```
%if "%{compiler_family}" != "intel" && "%{compiler_family}" != "arm1"
Requires:      openblas-%{compiler_family}%{PROJ_DELIM}
%endif
```
The superlu module always loads openblas and so openblas needs to be a dependency.

PR #1706, opened 2023-02-22
runtimes/charliecloud: add a "BuildRequires: make"

martin-g:
Add a test mapping too

Successful OBS builds:
* https://obs.openhpc.community/package/live_build_log/home:mgrigorov/charliecloud-gnu12-openmpi4/OpenHPC_3.0_Factory_openEuler_22.03/x86_64
* https://obs.openhpc.community/package/live_build_log/home:mgrigorov/charliecloud-gnu12-openmpi4/OpenHPC_3.0_Factory_openEuler_22.03/aarch64

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -38,5 +38,6 @@ openeuler_task:
     . /etc/profile.d/lmod.sh
     tests/ci/run_build.py ohpc $(tests/ci/cirrus_get_changed_files.sh)
   test_script: |
+    export JOB_SKIP_CI_SPECS=components/runtimes/charliecloud/SPECS/charliecloud.spec
     chown ohpc -R tests
     tests/ci/setup_slurm_and_run_tests.sh ohpc $(tests/ci/cirrus_get_changed_files.sh)
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -37,7 +37,10 @@ openeuler_task:
   build_script: |
     . /etc/profile.d/lmod.sh
     tests/ci/run_build.py ohpc $(tests/ci/cirrus_get_changed_files.sh)
+  env:
+    JOB_SKIP_CI_SPECS: |
+      components/runtimes/charliecloud/SPECS/charliecloud.spec
   test_script: |
-    export JOB_SKIP_CI_SPECS=components/runtimes/charliecloud/SPECS/charliecloud.spec
+    export SKIP_CI_SPECS="${SKIP_CI_SPECS}${JOB_SKIP_CI_SPECS}"
     chown ohpc -R tests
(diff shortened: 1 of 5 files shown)


martin-g:
The test fails on Rocky with:
```
$ ch-run alpine -- cat /etc/os-release
ch-run[13]: error: can't init user+mount namespaces: Operation not permitted (ch_core.c:400 1)
```

Google suggests https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux_atomic_host/7/html-single/getting_started_with_containers/index#user_namespaces_options but I cannot find how to pass kernel args (`namespace.unpriv_enable=1`) and I cannot pass ` user.max_user_namespaces=15076` neither as `docker run --sysctl ...` (see https://docs.docker.com/engine/reference/commandline/run/#sysctl) nor with `sysctl -w  user.max_user_namespaces=15076` inside the container (because of: `read-only file system`)

Ideas ?

adrianreber:
I think your analysis is not correct. User namespaces where not enabled in RHEL 7 and clones. We are running, however, not on a  RHEL 7 kernel. This is a Ubuntu kernel.

My first try would be to see if running a privileged container enables us to use user namespaces in the containers set up by GitHub Actions.

martin-g:
I've tried the following on my dev machine (Ubuntu 22.04 x86_64):
```
$ docker run -it --rm --privileged  rocky-full:8 bash
[root@4b3237235bad /]# USER=root ch-run alpine -- cat /etc/os-release
ch-run[13]: error: Invalid argument (ch_core.c:226 22)
```

rocky-full is rockylinux:8 with singularity-ce and charliecloud installed, and `singularity build --sandbox alpine docker://alpine:3` already done

martin-g:
Same without privileged:
```
$ docker run -it --rm  rocky-full:8 bash
[root@16ddd10c2dc6 /]# USER=root ch-run alpine -- cat /etc/os-release
ch-run[13]: error: can't init user+mount namespaces: Operation not permitted (ch_core.c:400 1)
```

adrianreber:
So running a container inside of a container is described for example in this article: https://www.redhat.com/sysadmin/podman-inside-container

Using podman I can run the charliecloud test locally inside of a container.

Your error messages are different if running with and without `--privileged`.

martin-g:
> Your error messages are different if running with and without `--privileged`.

Yes. This is what I expect to see at GHA with 9d6977c66 too.

martin-g:
Nice! RockyLinux passed!
I'll see how to exclude charliecloud for openEuler because it does not provide singularity/apptainer at the moment.

martin-g:
All checks passed!
I think we can simplify the `[JOB_]SKIP_CI_SPECS` env vars. IMO there is no need of two vars. A job could just append specs to the global one.
Let me try it!

martin-g:
The PR is ready for review!

PR #1711, opened 2023-02-24
Update warewulf-ipmi to 3.10.0

martin-g:
Add a patch for openEuler because its `ldd` returns paths starting with `/usr/lib**` while all other distros (e.g. Rocky and Ubuntu) return `/lib**`, i.e. without the leading `/usr`. Those are symlinks anyway.

Set wwsrvdir to "/var/lib" to be able to pass the build. The problem is that if we use a custom value then it is ignored. The configure line looks like:

```
[   22s] + ./configure --build=aarch64-openEuler-linux-gnu --host=aarch64-openEuler-linux-gnu \
 --program-prefix=  --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin \
 --sbindir=/usr/sbin  --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 \
 --libexecdir=/usr/libexec  --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man \
 --infodir=/usr/share/info --localstatedir=/var/lib --sharedstatedir=/var/lib --with-local-ipmitool=yes
```
Note that it contains `--localstatedir` and `--sharedstatedir` twice. The second ones are from `%configure --localstatedir=%{wwsrvdir} --sharedstatedir=%{wwsrvdir} %{?CONF_FLAGS}` from the .spec but they seem to be always ignored.
Without this change the `%files initramfs-%{_arch}` at the common of the spec fail to find the directories and the file

### DIFF
--- a/components/provisioning/warewulf-ipmi/SOURCES/openEuler.initramfs.patch
+++ b/components/provisioning/warewulf-ipmi/SOURCES/openEuler.initramfs.patch
@@ -1,0 +1,13 @@
+diff --git i/ipmi/initramfs/Makefile.am w/ipmi/initramfs/Makefile.am
+index 65635a6..b1661ce 100644
+--- i/ipmi/initramfs/Makefile.am
++++ w/ipmi/initramfs/Makefile.am
+@@ -41,6 +41,8 @@ rootfs: ipmitool
+ 	mkdir rootfs/sbin
+ 	mkdir rootfs/lib
+ 	mkdir rootfs/lib64
++	mkdir -p rootfs/usr/lib
++	mkdir -p rootfs/usr/lib64
+ 	mkdir -p rootfs/warewulf/provision/
+ 	chmod -R u+w rootfs/
+ 	cp -a ipmitool rootfs/sbin/
--- a/components/provisioning/warewulf-ipmi/SPECS/warewulf-ipmi.spec
+++ b/components/provisioning/warewulf-ipmi/SPECS/warewulf-ipmi.spec
@@ -36,6 +36,10 @@ Requires: %{name}-initramfs-%{_arch} = %{version}-%{release}
 %global localipmi 1
 BuildRequires: ipmitool
 Requires: ipmitool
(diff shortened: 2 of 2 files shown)


martin-g:
Fails due to `No matching package to install: 'warewulf-common-ohpc'`

martin-g:
BTW where do the default configure options come from ? I don't find them defined neither in openhpc/components nor in warewulf sources.

```
--build=aarch64-openEuler-linux-gnu --host=aarch64-openEuler-linux-gnu \
 --program-prefix=  --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin \
 --sbindir=/usr/sbin  --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 \
 --libexecdir=/usr/libexec  --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man \
 --infodir=/usr/share/info
```

martin-g:
@adrianreber Do you any idea about the problem with `wwsrvdir` ?

adrianreber:
> BTW where do the default configure options come from ? I don't find them defined neither in openhpc/components nor in warewulf sources.
> 
> ```
> --build=aarch64-openEuler-linux-gnu --host=aarch64-openEuler-linux-gnu \
>  --program-prefix=  --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin \
>  --sbindir=/usr/sbin  --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 \
>  --libexecdir=/usr/libexec  --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man \
>  --infodir=/usr/share/info
> ```

This is usually coming from the %{configure} macro: `rpm -E '%{configure}'`

PR #1715, opened 2023-02-27
Change --localstatedir to --sharedstatedir for all ww3 3.10.0 modules

martin-g:
I believe this is the reason for the problem in https://github.com/openhpc/ohpc/pull/1711
warewulf-common should use `--sharedstatedir=...`, so that later https://github.com/warewulf/warewulf3/blob/f13a37796c333069a7c9e08aa0b00c0643cf3ee0/ipmi/configure.ac#L44 would properly resolve it.

3.10.0 uses `--sharedstatedir=...`:
https://github.com/warewulf/warewulf3/blob/c6de604fc76eabfaef2cb99f4c6ae5ed44eff1e0/ipmi/warewulf-ipmi.spec.in#L43

Latest `master` does not set local/shared statedir at all: https://github.com/warewulf/warewulf3/blob/master/ipmi/warewulf-ipmi.spec.in#L33

### DIFF
--- a/components/provisioning/warewulf-cluster/SPECS/warewulf-cluster.spec
+++ b/components/provisioning/warewulf-cluster/SPECS/warewulf-cluster.spec
@@ -56,7 +56,7 @@ cd %{_builddir}
 
 %build
 ./autogen.sh
-%configure --localstatedir=%{wwsrvdir}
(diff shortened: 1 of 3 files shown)


adrianreber:
Sounds good. Also see https://github.com/warewulf/warewulf3/pull/287

PR #1326, opened 2021-06-18
Update LIKWID to version 5.2.0 and enable aarch64 builds

TomTheBear:
This PR adds version 5.2.0 of LIKWID with Intel IcelakeSP and AMD Zen3 support. The spec file was updated to work also for aarch64 builds.

PR #1718, opened 2023-02-28
Add support for building Likwid on Linux ARM64

martin-g:
Fixes https://github.com/openhpc/ohpc/pull/1326

OBS builds on x86_64 and aarch64: https://obs.openhpc.community/package/show/home:mgrigorov/likwid-gnu12-openmpi4

### DIFF
--- a/components/perf-tools/likwid/SPECS/likwid.spec
+++ b/components/perf-tools/likwid/SPECS/likwid.spec
@@ -83,7 +83,14 @@ It consists of:
           FC="ifort" \
           FCFLAGS="-module ./" \
 %else
+    %ifarch aarch64
+          COMPILER="GCCARMv8" \
+          BUILDDAEMON="true" \
+          BUILDFREQ="true" \
+          ACCESSMODE="perf_event" \
+    %else
           COMPILER="GCC" \
+    %endif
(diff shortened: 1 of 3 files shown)


adrianreber:
Just saw it now. With your changes to the skip setup the package is not even built on any architecture. Just the testing should be skipped, we should still try to do a build.

martin-g:
Oops! Working on it!

components/perf-tools/likwid/SPECS/likwid.spec
```diff
@@ -83,7 +83,14 @@ It consists of:
           FC="ifort" \
           FCFLAGS="-module ./" \
 %else
+    %ifarch aarch64
+          COMPILER="GCCARMv8" \
+          BUILDDAEMON="true" \
```
martin-g, line 88: I am re-testing at OBS without the `BUILD**` options. Their values are `true` by default anyway.

martin-g: No. It seems those are needed for ARM64.

adrianreber:
Is this ready? Not sure because of the open conversation.

martin-g:
It is ready!

PR #1724, opened 2023-03-13
Update RHEL to 9 and Leap to 15.4 at GHA and CirrusCI for branch 3.x

martin-g:
Requested-by: Adrian Reber at Slack: https://openhpc.slack.com/archives/C04JN2Y20J0/p1678714471974889?thread_ts=1678692429.621319&cid=C04JN2Y20J0

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -13,13 +13,7 @@ task:
     cpu: 4
     memory: 12G
   script: uname -a
-  setup_script: |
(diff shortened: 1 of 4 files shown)


adrianreber:
I am not sure Slack links are helpful in git history. Especially as the OpenHPC slack only provides the messages of the last 90 days.

I think this should also enable CI on RHEL 9 and LEAP 15.4.

martin-g:
OK! 
I've updated the commit message!
Instead of disabling RHEL 8 and Leap 15.3 I updated them to RHEL 9 and Leap 15.4. Is this what you meant ?

adrianreber:
> Instead of disabling RHEL 8 and Leap 15.3 I updated them to RHEL 9 and Leap 15.4. Is this what you meant ?

The idea was to just test those distributions we are targeting with each branch.

PR #1731, opened 2023-03-15
Update openmpi from 4.1.4 to 4.1.5

martin-g:
https://www.mail-archive.com/announce@lists.open-mpi.org/msg00155.html

OBS builds: https://obs.openhpc.community/package/show/home:mgrigorov/openmpi-gnu12-openmpi4

### DIFF
--- a/components/mpi-families/openmpi/SPECS/openmpi.spec
+++ b/components/mpi-families/openmpi/SPECS/openmpi.spec
@@ -40,7 +40,7 @@ Summary:   A powerful implementation of MPI/SHMEM
 
 Name:      %{pname}%{RMS_DELIM}-%{compiler_family}%{PROJ_DELIM}
 
-Version:   4.1.4
+Version:   4.1.5
 Release:   1%{?dist}
 License:   BSD-3-Clause
 Group:     %{PROJ_NAME}/mpi-families
--- a/tests/ci/setup_slurm_and_run_tests.sh
+++ b/tests/ci/setup_slurm_and_run_tests.sh
@@ -25,6 +25,10 @@ dnf -y install \
 # shellcheck disable=SC2046 # (we want the words to be split)
 dnf -y install $(find /home/"${USER}"/rpmbuild/RPMS/ -name "*rpm") || true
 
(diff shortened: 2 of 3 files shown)


adrianreber:
This test error is interesting. The test suite fails to detect slurm but only on aarch64 and not on x86_64.

martin-g:
I see two noticeable differences in the logs:

1) the hostname
1.1) on x86_64:  `06d3d095cbcd`
1.2) on aarch64: `cirrus-ci-task-6672924006940672`

Later in the failures on aarch64 we see:
```
not ok 1 [modules] env variable passes through ()
# (from function `assert_success' in file ../common/test_helper_functions.bash, line 58,
#  in test file rm_execution, line 27)
#   `assert_success' failed
# 
# -- command failed --
# status : 162
# output (3 lines):
#   EXE = ./test_env
#   Connection refused
#   qsub: cannot connect to server cirrus- (errno=15010)
```
It seems there is a problem with the hostname. 


2) in the output of `configure`:
2.1) x86_64:   checking for /etc/pbs.conf... **no**
2.2) aarch64: checking for /etc/pbs.conf... **yes**

.cirrus.yml
```diff
@@ -21,6 +21,9 @@ task:
     JOB_SKIP_CI_SPECS: |
       components/perf-tools/likwid/SPECS/likwid.spec
   test_script: |
+    # Remove OpenPBS, if it is installed as a dependency. The tests use Slurm Resource Manager
+    dnf -y remove openpbs-*-ohpc || true
+    rm -f /etc/pbs.conf || true
```
adrianreber: What about doing it only once in `tests/ci/setup_slurm_and_run_tests.sh`?

`rm -f /etc/pbs.conf` does not need `|| true`. It will always return `0`.

adrianreber:
Please also rebase.

.cirrus.yml
martin-g: Done!

adrianreber:
Thanks. Can you also open the same PR against the 2.x branch?

martin-g:
Done: https://github.com/openhpc/ohpc/pull/1737

PR #1735, opened 2023-03-16
Upload GitHub Actions event file for result reporting

adrianreber:
This PR should have test results added as a comment.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -222,3 +222,14 @@ jobs:
       run: |
         . /etc/profile.d/lmod.sh
         tests/ci/run_build.py ohpc ${{ steps.files.outputs.added_modified }}
+
+  event_file:
+    name: "Event File"
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/validate.yml
```diff
@@ -222,3 +222,14 @@ jobs:
       run: |
         . /etc/profile.d/lmod.sh
         tests/ci/run_build.py ohpc ${{ steps.files.outputs.added_modified }}
+
+  event_file:
+    name: "Event File"
+    runs-on: ubuntu-latest
+    steps:
+    - name: Upload
+      uses: actions/upload-artifact@v2
```
martin-g, line 231: Please use v3 because this action logs:

```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/upload-artifact@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
```

PR #1736, opened 2023-03-17
admin/meta-packages: Exclude arm1/intel compiler for openEuler

Yikun:
This patch exclude the arm1/intel compiler based package for openEuler.


martin-g (approved):

adrianreber:
Can you remove the following comments:
```
 warning: extra tokens at the end of %endif directive in line 722:  %endif  # <-- end non-aarch64 only

warning: extra tokens at the end of %endif directive in line 813:  %endif  # <-- end aarch64 only

warning: extra tokens at the end of %endif directive in line 814:  %endif  # <-- end non openEuler

warning: extra tokens at the end of %endif directive in line 883:  %endif  # <-- end non openEuler
```

Newer versions of RPM do not like comments at the end.

PR #1742, opened 2023-03-20
EasyBuild v4.7.1

boegel:
Not sure if this should go to the `2.x` or `3.x` branch...

### DIFF
--- a/components/dev-tools/easybuild/SPECS/easybuild.spec
+++ b/components/dev-tools/easybuild/SPECS/easybuild.spec
@@ -15,7 +15,7 @@
 
 Summary:   Software build and installation framework
 Name:      EasyBuild%{PROJ_DELIM}
-Version:   4.7.0
+Version:   4.7.1
 Release:   %{?dist}.1
 License:   GPLv2
 Group:     %{PROJ_NAME}/dev-tools


martin-g:
Since it is bump of a patch version I think it is fine to apply it both in `3.x` and `2.x`.
First let's fix the tests! I'll take a look!

martin-g:
Please try with this change:
```diff
diff --git i/tests/ci/spec_to_test_mapping.py w/tests/ci/spec_to_test_mapping.py
index 2cfeaa7a1..2ca82f9cc 100755
--- i/tests/ci/spec_to_test_mapping.py
+++ w/tests/ci/spec_to_test_mapping.py
@@ -37,7 +37,7 @@ test_map = {
     'components/dev-tools/easybuild/SPECS/easybuild.spec': [
         'easybuild',
         '',
-        'gcc-c++',
+        'gcc-c++ python3-Cython-ohpc',
     ],
```

adrianreber:
@martin-g I think this can be solved by `BuildRequires: python3-devel`

martin-g:
@adrianreber I thought about it but then I remembered the recent problem with OpenPBS - `BuildRequires` installs it on the build VM but later it is missing in the test VM, for GitHub Actions.

adrianreber:
> @adrianreber I thought about it but then I remembered the recent problem with OpenPBS - `BuildRequires` installs it on the build VM but later it is missing in the test VM, for GitHub Actions.

Sorry, not sure what you are trying to say.

martin-g:
I mean that adding `BuildRequires: xyz` will make `xyz` available on the **build** Github Action runner. But `xyz` won't be available on the **test** runner.
This was the reason why the update of OpenMPI 4.1.5 was passing on Github Actions and failing on CirrusCI - because at Cirrus the same VM/runner is used for both build and test, and availability of OpenPBS was breaking the test.

Here the test fails with `no python in the PATH`.

martin-g:
I think we might be looking in the wrong direction.
I see `python3` is installed for the tests.
Maybe the test needs to look for `python3` instead of `python` ?!

adrianreber:
If you run EasyBuild from OBS or from CI you will see the following:
```
+ verbose ''\''python3'\'' version: 3.9.14, which matches Python 3 version requirement (>= 3.5)'
+ '[' -n '' ']'
+ PYTHON=python3
+ '[' -n python3 ']'
+ python3 -c 'import easybuild.framework'
+ verbose ''\''python3'\'' is NOT able to import '\''easybuild.framework'\'' so NOT retaining it'
+ '[' -n '' ']'
+ unset PYTHON
+ '[' -n '' ']'
```
This is because in the module file there is following PYTHONPATH `/opt/ohpc/pub/libs/easybuild/4.7.0/lib/python%{python3_version}/site-packages`

So the macro `%{python3_version}` is not correctly converted during `rpmbuild`. `BuildRequires: python3-devel` fixes that.

adrianreber:
Replaced with #1784

PR #1758, opened 2023-03-29
Update test-suite

jcsiadal:
Update test package
- Add config support for Rocky Linux 8
- Remove Python 2 tests if they belong to removed packages
- Update all remaining Py2 tests to Python3, using 2to3
- Update specfile (remove Py2 deps and clean up).
- Added generation of doc and test tarballs to get_sources. (also turned on error printing for build_srpm; log files aren't that useful if errors are being sent to null)

More work to do on individual tests, but this allows it to build and configure on SLES15.4.


jcsiadal:
@adrianreber Does the CI have a routine for generating the doc and test-suite tarballs needed to validate these packages?

components/admin/test-suite/SPECS/tests.spec
```diff
@@ -8,73 +8,77 @@
...
-    -s /bin/bash -m -b /home %{testuser}
+    -s /bin/bash -m -b %{homedir} %{testuser}
 %if 0%{?suse_version}
 if [ $(getent group singularity) ]; then
     usermod -a -G singularity ohpc-test
 fi
 %endif
-exit 0
+if [ $(getent group geopm) ]; then
+    usermod -a -G geopm ohpc-test
```
martin-g:
```suggestion
    usermod -a -G geopm %{testuser}
```

components/admin/test-suite/SPECS/tests.spec
```diff
@@ -8,73 +8,77 @@
...
 
 BuildRequires:  autoconf%{PROJ_DELIM}
 BuildRequires:  automake%{PROJ_DELIM}
+BuildRequires:  sed, findutils
 
-%global __brp_mangle_shebangs_exclude bats
-%undefine __brp_mangle_shebangs
-
-%if 0%{?suse_version}
+%if 0%{?suse_version} || 0%{?sle_version}
```
adrianreber, line 34: From what I saw in `rpm --showrc` checking for `%{?sle_version}` should be enough I think.

components/admin/test-suite/SPECS/tests.spec
```diff
@@ -8,73 +8,77 @@
...
 Requires(pre):  shadow
-Requires: python-base
+Requires: python3-base
 %endif
 
 %if 0%{?rhel_version}
```
adrianreber: Can you also change this to `0%{?rhel}`. This is one of the OBS specific macros which are not available outside of OBS.

adrianreber:
This should at least be three commits. One for the spec file changes. One for the removal of unused tests and one for the 2to3 conversion (thanks for looking at that).

> Does the CI have a routine for generating the doc and test-suite tarballs needed to validate these packages?

Unfortunately not yet. I thought about having something for those cases, but currently we do not. Would be good to have. We would need a hook in the scripts somewhere which detects the documentation spec file and test-suite spec file and then create the tarballs during CI runs. Do you want to try to implement something which could help us here?

components/admin/test-suite/SPECS/tests.spec
jcsiadal: I think you're correct, but I'm going by the chart here: 
https://en.opensuse.org/openSUSE:Build_Service_cross_distribution_howto
It implies that someone running Tumbleweed requires a suse_version check. I know it's not formally supported, but I still want it to work.
I've been kicking around adding %{suse} to OHPC_MACROS as the counterpart to %{rhel}.

components/admin/test-suite/SPECS/tests.spec
jcsiadal: Good catch. I'll do that.

jcsiadal:
I think it's funny because I usually get told to squash my commits in a PR (even if they're not just to fix a previous commit). Since it's all in a single commit, undoing and splitting should be simple. I'll also look at the tarball generation.

One other thing I'd like to do is move the whole test suite to /opt/ohpc and not create a user. I'd like to leave the test-user setup to the cluster admin and also allow other users to access to tests. I won't make that change in this PR.

adrianreber:
> I think it's funny because I usually get told to squash my commits in a PR (even if they're not just to fix a previous commit). Since it's all in a single commit, undoing and splitting should be simple. I'll also look at the tarball generation.

I am usually asking to squash fix up commits. If you are fixing things in a PR then we do not need that in the git history if we found it during the review.

If you have multiple logical things in your PR like here, then it should be in separate commits. Especially if we later need to look at the history it makes it easier to understand the change and possibly revert it.

adrianreber:
> One other thing I'd like to do is move the whole test suite to /opt/ohpc and not create a user. I'd like to leave the test-user setup to the cluster admin and also allow other users to access to tests. I won't make that change in this PR.

Not sure if this doable when looking at Jenkins. Not sure how Jenkins ingests the test suite.

jcsiadal:
It's all working (at least the build/install). The failures on RHEL/CentOS are happening because our autoconf package is missing dependencies. The autoconf build dependencies are set, but the install dependencies are missing. I'll issue a separate PR.

PR #1759, opened 2023-03-29
Add check for license file

jcsiadal:
Check that spec adds a non-overridable license file.

### DIFF
--- a/tests/ci/check_spec.py
+++ b/tests/ci/check_spec.py
@@ -29,6 +29,9 @@ regex_required = [
     'provisioning|rms|runtimes|serial-libs)$',
     # Need a URL
     '(^URL:.*$|Url:.*$)',
+    # RPMs should include their license files; don't use %doc for licenses
+    # Add a comment if no license file is needed
+    '(^%License:.*$|^# No license.*$)',
     ]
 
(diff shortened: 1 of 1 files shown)


martin-g (approved):

tests/ci/check_spec.py
```diff
@@ -29,6 +29,8 @@
     'provisioning|rms|runtimes|serial-libs)$',
     # Need a URL
     '(^URL:.*$|Url:.*$)',
+    # OSS redistribution requires a license file
```
adrianreber: The comment does not seem entirely correct. I think, this is the RPM `License` tag and has not direct connection to a file. Besides that, good check.

jcsiadal:
Fixed comment.
Also provided the ability to meet the check by adding a comment like:
`# No license file -- already added in main package` [or]
`# No license file needed for meta-RPM`

adrianreber:
Sorry, I had to revert this. It is not clear what this should do.

```
    '(^%License:.*$|^# No license.*$)',
```

Do you want to check for the spec file tag `License:` without `%` in the front or do you want to check for the `%license` designator in the `%files` section. It seems the combination of `%` and `:` makes no sense.

PR #1762, opened 2023-03-29
Remove ADIOS1 spec and tests in favour of ADIOS2

martin-g:
As discussed with @adrianreber in Slack.


martin-g:
Should I try to update the docs too ?
There are occurrences like `/docs/recipes/install/rocky8/aarch64/warewulf/slurm/manifest/io-libs.tex:15:adios-arm1-openmpi4-ohpc &`

misc/obs/config.2.x
```diff
@@ -19,7 +19,7 @@ dev-tools     = ["autoconf","automake","cmake","easybuild","hwloc","libtool","mp
 	         "scipy","spack","valgrind"]
 distro-packages = ["python-Cython","flex"]
 fs            = ["lustre-client"]
-io-libs       = ["adios","hdf5","netcdf","netcdf-cxx","netcdf-fortran","phdf5","pnetcdf","sionlib",
```
adrianreber: Do not remove it here. There is still the OBS config for the 2.x branch in that file and that would probably break if you remove it.

martin-g: Undone!

adrianreber:
> Should I try to update the docs too ? There are occurrences like `/docs/recipes/install/rocky8/aarch64/warewulf/slurm/manifest/io-libs.tex:15:adios-arm1-openmpi4-ohpc &`

Not yet.

misc/obs/config.2.x
```diff
@@ -19,7 +19,7 @@ dev-tools     = ["autoconf","automake","cmake","easybuild","hwloc","libtool","mp
...
+io-libs       = ["adios", "hdf5","netcdf","netcdf-cxx","netcdf-fortran","phdf5","pnetcdf","sionlib",
```
adrianreber: sorry, now there is a space (not really a problem, but also an unnecessary change)

PR #1763, opened 2023-03-29
perf-tools/extrae: fix arm compiler errors

adrianreber:


components/perf-tools/extrae/SPECS/extrae.spec
```diff
@@ -75,11 +73,13 @@ export LDFLAGS="$LDFLAGS -lz"
...
 %if "%{compiler_family}" == "arm1"
-    CFLAGS="-O3 -fsimdmath -fPIC" CXXFLAGS="-O3 -fsimdmath -fPIC" FCFLAGS="-O3 -fsimdmath -fPIC" \
+    CFLAGS="${CFLAGS} -fsimdmath -fPIC -Wno-implicit-function-declaration" \
+    CXXFLAGS="${CXXFLAGS} -fsimdmath -fPIC" \
+    FCFLAGS="${FCFLAGS} -fsimdmath -fPIC" \
 %endif
     || { cat config.log && exit 1; }
 
-make %{?_smp_mflags}
+make %{?_smp_mflags} V=1
```
martin-g, line 82: What is `V=1` ? Looks like a debug leftover (verbose output ?!)

adrianreber: That is for verbose output, yes. It is for debug, but not really a leftover. I think it is easier to debug failed builds if the complete compiler command-line in the build-log is visible instead of just:
```
 CC path/to/file
```

PR #1764, opened 2023-03-30
autoconf Dependency Fix

jcsiadal:
Fixes missing install dependencies.
Also made some formatting changes to the specfile.

### DIFF
--- a/components/dev-tools/autoconf/SPECS/autoconf.spec
+++ b/components/dev-tools/autoconf/SPECS/autoconf.spec
@@ -24,11 +24,18 @@ Source0:   https://ftp.gnu.org/gnu/autoconf/autoconf-%{version}.tar.gz
 BuildRequires: m4 make
 Requires: m4
 
+BuildRequires: perl
 BuildRequires: perl-macros
 BuildRequires: perl(File::Compare)
 BuildRequires: perl(File::Copy)
 BuildRequires: perl(Data::Dumper)
 BuildRequires: perl(Text::ParseWords)
+%if 0%{?suse_version} || 0%{?sle_version}
+Requires: perl-base
+%else
+Requires: perl-interpreter
+%endif
+Requires: perl(File::Compare)
 
 %define install_path %{OHPC_UTILS}/autotools
 
@@ -48,15 +55,20 @@ may be configuring software with an Autoconf-generated script;
 Autoconf is only required for the generation of the scripts, not
 their use.
 
+
 %prep
 %setup -q -n autoconf-%{version}
+
 
 %build
 %ifarch ppc64le
 cp /usr/lib/rpm/config.guess build-aux
 %endif
 
-./configure --prefix=%{install_path} || { cat config.log && exit 1; }
+./configure --prefix=%{install_path} \
+            --libdir=%{install_path}/lib \
+            || { cat config.log && exit 1; }
+
 
 %install
 make %{?_smp_mflags} DESTDIR=$RPM_BUILD_ROOT install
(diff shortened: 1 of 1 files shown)


components/dev-tools/autoconf/SPECS/autoconf.spec
```diff
@@ -24,11 +24,18 @@ Source0:   https://ftp.gnu.org/gnu/autoconf/autoconf-%{version}.tar.gz
...
+BuildRequires: perl
 BuildRequires: perl-macros
 BuildRequires: perl(File::Compare)
 BuildRequires: perl(File::Copy)
 BuildRequires: perl(Data::Dumper)
 BuildRequires: perl(Text::ParseWords)
+%if 0%{?suse_version} || 0%{?sle_version}
+Requires: perl-base
+%else
+Requires: perl-interpreter
```
adrianreber, line 36: The perl-base or perl-interpreter explicit dependency seems unnecessary. `perl(File::Compare)` pulls in those packages on each platform already.

components/dev-tools/autoconf/SPECS/autoconf.spec
```diff
@@ -48,37 +55,42 @@ may be configuring software with an Autoconf-generated script;
...
+
 
 %install
 make %{?_smp_mflags} DESTDIR=$RPM_BUILD_ROOT install
 
 # remove share/info/dir to avoid conflict with other package installs
 rm -f $RPM_BUILD_ROOT/%{install_path}/share/info/dir
 
-%{__mkdir_p} ${RPM_BUILD_ROOT}/%{_docdir}
+mkdir -p ${RPM_BUILD_ROOT}/%{_docdir}
```
adrianreber, line 79: Why this change?

adrianreber:
> Also made some formatting changes to the specfile.

If you want to make formatting changes please add them to separate commits and do not mix functional changes with formatting changes.

components/dev-tools/autoconf/SPECS/autoconf.spec
jcsiadal: I looked through the dependency chain and it doesn't appear that RHEL's perl-file-compare will install the perl binary. I also looked at the Fedora autoconf specfile and they explicitly add perl-interpreter. The SUSE one isn't needed, but I wanted to keep things parallel.

components/dev-tools/autoconf/SPECS/autoconf.spec
jcsiadal: I'm surprised you would ask. Red Hat's RPM guide (at least one of them) says not to use (LSB) shell command macros. They were put in to make RPM portable across OSs (like Windows or BSD) and that never happened. They're just extra work with no benefit. SUSE says to never use them.
When I touch an RPM spec, I try and clean it up to current RPM standards, i.e. both RH and SUSE mandate %license now.

components/dev-tools/autoconf/SPECS/autoconf.spec
adrianreber: If I do `dnf -y install "perl(File::Compare)"` `perl-interpreter` is pulled in on 9.1.

jcsiadal: > "you SHOULD NOT assume any other packages are present as RPM dependencies and anything brought into the buildroot by the build system can change over time." 

That's from Red Hat's packaging guidelines. They do say that one can usually drop Requires: if we're running autorequires, but I thought we're only doing manual deps. The other question to ask: if it's not needed, why does Fedora include it in their autoconf spec?

PR #1766, opened 2023-03-30
io-libs/adios2: disable rpath support

adrianreber:
The adios2 build system inserts a strange rpath:

Set runtime path of "...." to "$ORIGIN/../lib64"

Which breaks rpath verification at the end of the build.

This disables rpath (like found in trilinos)

@martin-g I think your initial idea was correct. This seems to be as discussed on Slack.


martin-g (approved):
Hopefully this won't break someone's usage!

components/io-libs/adios2/SPECS/adios2.spec
```diff
@@ -93,6 +93,8 @@ cmake \
     -DCMAKE_INSTALL_PREFIX=%{install_path} \
     -DCMAKE_BUILD_TYPE=Release \
     -DBUILD_SHARED_LIBS=ON \
+    -DCMAKE_SKIP_INSTALL_RPATH:BOOL=ON \
+    -DCMAKE_SKIP_RPATH:BOOL=ON \
```
martin-g, line 97: I wonder would it better if those two lines are wrapped in `%if "%{compiler_family}" == "arm1"` ?!
It was failing only for `arm1` and you said that RPATH is quite common in HPC world.

PR #1773, opened 2023-04-02
OpenPMIx Update

jcsiadal:
This is a prototype package I'm placing here for review, or to move into the development tree.

One major change is moving it under its proper name.

This update also adds PRRTE to the build, but I needed to add as a second spec. The full OpenPMIx package is required to build PRRTE, but I was not able to merge the two into a single build. There are a lot of hardcoded paths added during build and I was only able to get a successful build with PMIx in its final location. Even doing a temp build and install of PMIX introduced different errors in PRRTE.

If our build system can handle both specs in one directory, great; otherwise, we'll need to have a new package under RMS for PRRTE.

These both are using RCs, as they will likely be full releases before 3.0 and these specs also work with the current production releases.


jcsiadal:
@mghpcsim Prototyping the PMIx/PRRTE update. It seems to work, but see my notes in the summary.

jcsiadal:
@adrianreber Is the devel repo not active for CI on RHEL-type builds?

adrianreber:
> @adrianreber Is the devel repo not active for CI on RHEL-type builds?

No, so far it was not necessary. You can enable it.

CI can also not handle packages that dependent on each other. You would need to split this in two PRs.

PR #1774, opened 2023-04-03
Latest GEOPM Update

jcsiadal:
Updates GEOPM to version 2.0.1. There are two issues that will need to be addressed:

1. GEOPM uses separate service and policy components. These are both easily built in one RPM spec to generate two RPMs. The service RPM is installed directly into root fs (/) as a systemd service and the policy RPM is built using a specific toolchain. The issue is that the service RPM is rebuilt for each toolchain (it's a very fast build). Our other option is to split into two specfiles or implement OBS-checking logic and only build under one toolchain.
2. I've added a python-dasbus component under distro-packages. The el8 and el9 (not SUSE) distro packages are too old to support GEOPM 2. Dasbus is a python3 module required by the GEOPM service at runtime.


components/perf-tools/geopm/SPECS/geopm.spec
```diff
@@ -9,140 +9,238 @@
...
-BuildRequires: python3-devel
 BuildRequires: unzip
-BuildRequires: zlib-devel
-BuildRequires: openssh
+BuildRequires: sqlite-devel
 
-%if 0%{?suse_version}
-Buildrequires: libelf1
+%if 0%{?suse_version} || 0%{?sle_version}
+Buildrequires: libelf-devel
```
martin-g, line 48:
```suggestion
BuildRequires: libelf-devel
```

PR #1775, opened 2023-04-06
Fix the syntax for usages of %systemd_** macros

martin-g:
They accept only a list of service names. It is not expected to redirects (like ">/some/where") or piping (like "|| true")

Fix a warning about a missing group by creating it if it does not exist.

### DIFF
--- a/components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
+++ b/components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
@@ -230,6 +230,9 @@ This package contains the CGI scripts and event components to
 provision systems.  Systems used solely for administration of Warewulf
 do not require this package.
 
+%pre server
+/usr/bin/getent group %{httpgrp} 1>/dev/null 2>&1 || \
+  /usr/sbin/groupadd -r %{httpgrp}
 
 %post server
 # Update users and services on first time installation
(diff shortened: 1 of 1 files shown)


martin-g:
OBS build results: https://obs.openhpc.community/package/live_build_log/home:mgrigorov/warewulf-provision-gnu12-openmpi4/Leap15/aarch64

adrianreber:
I do not think this is the right way to fix this. All these steps in the RPM scripts should be able to fail without aborting the installation. I would "fix" this with entries to `rpmlintrc` to not run the checks triggering this error.

martin-g:
```
$ rpm -E '%systemd_post abc'

 
if [ $1 -eq 1 ] && [ -x /usr/bin/systemctl ]; then 
    # Initial installation 
    /usr/bin/systemctl --no-reload preset abc || : 
fi 

$ rpm -E '%systemd_postun_with_restart abc'
 
if [ $1 -ge 1 ] && [ -x /usr/bin/systemctl ]; then 
    # Package upgrade, not uninstall 
    /usr/bin/systemctl try-restart abc || : 
fi 

```

The scriptlet impl itself makes sure it does not fail.

I am not an expert in RPM as you are, but I think the current syntax in the spec is wrong!
It should be either fixed (as in this PR) or replaced to use `/bin/systemctl ... >/dev/null 2>&1 || true` instead.
Using an rpmlint rule will just suppress/hide the problem in our build.

adrianreber:
That is a very good argument. Let me look closer at it.

PR #1777, opened 2023-04-19
Fix wrong opensuse version in warewulf-vnfs.spec

Yikun:


koomie:
Unfortunately this PR swaps the Leap 15.3 and 15.4 install templates. I will commit a fix directly.

Yikun:
@koomie Oh, good catch, thanks for update!

PR #1781, opened 2023-04-19
admin/meta-packages: update version to 3.0

adrianreber:


martin-g (approved):

components/admin/meta-packages/SPECS/meta-packages.spec
```diff
@@ -17,7 +17,7 @@
 
 Summary: Meta-packages to ease installation
 Name:    meta-packages
-Version: 2.6.1
+Version: 3.0
```
Yikun, line 20: Not sure it should be `3.0.0` or not

PR #1787, opened 2023-04-20
Increase optimization baseline to x86-64-v2

adrianreber:


adrianreber:
@mslacken Do you know if LEAP 15.4 is also using `x86-64-v2` as baseline for optimization or is 15.4 expected to work on older CPUs?

adrianreber:
Looking at openEuler and LEAP it seems like only RHEL 9 is using `x86-64-v2`. For now I will only enable it for RHEL 9 and clones in OpenHPC.

Yikun:
I have confirmed with openEuler Intel Arch SIG.

- openEuler 22.03 series only support x86-64-v1
- Since openEuler 23.09 will start to support x86-64-v2 experimentally

mslacken:
Leap has just v1, tumbleweed moves now to v2.

adrianreber:
> Leap has just v1, tumbleweed moves now to v2.

Thanks for the confirmation.

adrianreber:
> I have confirmed with openEuler Intel Arch SIG.
> 
>     * openEuler 22.03 series only support x86-64-v1
> 
>     * Since openEuler 23.09 will start to support x86-64-v2 experimentally

Also thanks for the confirmation.

The current PR, only on RHEL 9, is the right way to go then.

PR #1788, opened 2023-04-20
Update the OBS project name from "OpenHPC" to "OpenHPC3"

martin-g:
Drop using the legacy crypto policies.


misc/obs/obs_config.py
```diff
@@ -79,7 +79,7 @@ def __init__(self, version):
         logging.info("--> Micro release   = %s" % self.microVer)
 
         if self.microVer == '0':
-            self.obsProject = "OpenHPC:" + self.branchVer + ":Factory"
+            self.obsProject = "OpenHPC3:" + self.branchVer + ":Factory"
```
adrianreber: Do not change this file, please.

martin-g: Reverted!

adrianreber:
Errors are expected currently because the repository moved.

adrianreber:
The logs show that the packages signed with the new key are installed without dnf/rpm complaining. Looks good. Thanks.

PR #1812, opened 2023-04-24
compiler-families/intel-compilers-devel: require exact version

adrianreber:
The current setup builds against the latest compiler available during build time, but our packages are requiring a minimum version which might not have all the symbols the compiler has we are using during build time.

With this change we are requiring once specific version during build time and during installation. This reduces the flexibility of our users but ensures that that version of the packages are installed which were used during build time.


adrianreber:
@mghpcsim PTAL

adrianreber:
@jcsiadal This seems to work locally for me. Any comments from your side? Would this work?

adrianreber:
Ah, I accidentally merged this. This was not the plan. Sorry. If any of you has a chance to look at this please let me know if you think this is correct or not correct.

PR #1813, opened 2023-04-25
Update oneAPI to fixed build release

jcsiadal:
Updates oneAPI to use a fixed release -- this keeps a common version used for all OpenHPC builds and validation. Additional versions of oneAPI can be installed in parallel; this should happen automatically if recommended packages are enabled.

### DIFF
--- a/components/compiler-families/intel-compilers-devel/SPECS/intel-compilers-devel.spec
+++ b/components/compiler-families/intel-compilers-devel/SPECS/intel-compilers-devel.spec
@@ -18,9 +18,8 @@
 # install newer versions during build time. If the user has the minimum
 # version, but the build system was already using a newer version, then
 # the resulting binaries might rely on symbols which are not present
-# in the minimum version.
+# in the minimum version.  Newer versions may still be installed in parallel.
 %define exact_intel_ver 2023.1.0
-
 
 Summary:   OpenHPC compatibility package for Intel(R) oneAPI HPC Toolkit
 Name:      %{pname}%{PROJ_DELIM}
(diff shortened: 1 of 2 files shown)


jcsiadal:
@adrianreber - Take a look at this. Currently, new version updates would be blocked. Each major package is version-in-name, so DNF can install in parallel. Above that is a no-version-in-name metafile so updates happen automatically, and above that is the top level toolkit metafile.

martin-g (approved):
LGTM

adrianreber:
Thanks. Looks good. The thing we need to remember is to always increase the version requirement to match the latest version available from Intel's yum repository.

adrianreber:
@jcsiadal I think the same change is needed on 3.x, right?

adrianreber:
You forgot to replace one of the macros. I did a quick fix up.

jcsiadal:
Thanks for catching that. I did a woefully incomplete level of validation to get this in ASAP.
Yes - the same changes should in 3.x, with some additional modifications. Let me know if I need to submit, or you can apply this commit to 3.x and I'll submit the extras separately.

adrianreber:
> Thanks for catching that. I did a woefully incomplete level of validation to get this in ASAP. Yes - the same changes should in 3.x, with some additional modifications. Let me know if I need to submit, or you can apply this commit to 3.x and I'll submit the extras separately.

Please create a PR with your changes and we can review it.

adrianreber:
@jcsiadal One more thing I was thinking about. The Intel compiler supports `-march=x86-64-v2` just like GCC does. I am in favour enabling it for RHEL 9 and clones (it is already active for all `gnu12` based builds). Do you see that as problematic to enable it also for the Intel compiler or also something you think would make sense?

PR #1816, opened 2023-05-04
compiler-families/gnu-compilers: add gcc 13 in parallel

adrianreber:
The changes of the spec file are a bit more complicated than expected due to the way OBS parses the spec file to download the sources.

This does not switch the default compiler to gcc 13, yet.


adrianreber:
https://obs.openhpc.community/package/show/home:adrianr/gnu-compilers

https://obs.openhpc.community/package/show/home:adrianr/gnu-compilers-gnu13

PR #1817, opened 2023-05-05
Updated oneAPI for 3.x branch

jcsiadal:
Copy of commit for 2.x branch, plus removal of optional PSXE modules.
Module generation pre-check uses new compiler RPM.


adrianreber:
I do not think introducing this much changes less than three weeks before the release is a good idea. Can you undo all the deletions and just bring over the version changes from the 2.x branch?

jcsiadal:
Actually, I think we should roll this into 2.x. PSXE is well past EOL. Do you want to schedule a short meeting to discuss?

adrianreber:
> Actually, I think we should roll this into 2.x. PSXE is well past EOL. Do you want to schedule a short meeting to discuss?

For the releases targeted for SC 23 that sounds like a good idea, but not now.

Issue #1818, opened 2023-05-05
hwloc-ohpc libraries hidden on node


middelkoopt:
Issue: Compute needs hwloc-ohpc libraries to start slurmd, but these libraries are hidden with the software mount off the head node.

After Fix:
```
[auser@c1 ~]$ df
Filesystem             1K-blocks    Used Available Use% Mounted on
tmpfs                    2006168 1279560    726608  64% /
devtmpfs                 1981968       0   1981968   0% /dev
tmpfs                    2006168       0   2006168   0% /dev/shm
tmpfs                    2006168    8532   1997636   1% /run
tmpfs                    2006168       0   2006168   0% /sys/fs/cgroup
10.4.0.2:/opt/ohpc/pub  17197056 6408960  10788096  38% /opt/ohpc/pub
10.4.0.2:/home          17197056 6408960  10788096  38% /home

[auser@c1 ~]$ find /opt/ohpc/pub -type d |grep hwloc/lib
/opt/ohpc/pub/libs/hwloc/lib
/opt/ohpc/pub/libs/hwloc/lib/pkgconfig
```

Error
```
May 05 12:30:44 c1 systemd[1]: Started Slurm node daemon.
May 05 12:30:44 c1 slurmd[1080]: /usr/sbin/slurmd: error while loading shared libraries: libhwloc.so.15: cannot open shared object file: No such file or directory
May 05 12:30:44 c1 systemd[1]: slurmd.service: Main process exited, code=exited, status=127/n/a
May 05 12:30:44 c1 systemd[1]: slurmd.service: Failed with result 'exit-code'.
```

Flavor: x86_64, Rocky 8.7, warewulf 3
Platform: qemu-system-x64_x64 on Ubuntu 22.04 on Jetstream2 w/  tun/tap bridge networking using kvm and virtio drivers

adrianreber:
I guess this does not happen in testing because we install openmpi4 on the head node which pulls in hwloc-ohpc.

Do you skip installing openmpi4 on the head node? Just curious why you are seeing this.

middelkoopt:
Correct, openmpi4 is not installed on the head node. This is a minimal install for testing and leaning, so I did not include any extra packages to reduce complexity. I followed the quickstart instructions for the build.

middelkoopt:
Applied against 3.x and closing issue.

PR #1819, opened 2023-05-05
TM 1818 hwloc

middelkoopt:
Patch for issue #1818,  hwlock not on compute node.


middelkoopt:
Targeted wrong branch!

PR #1820, opened 2023-05-05
Add hwloc-ohpc to head node to fix issue #1818

middelkoopt:
Patch for issue #1818,  hwlock not on compute node.

### DIFF
--- a/docs/recipes/install/common/install_provisioning_intro.tex
+++ b/docs/recipes/install/common/install_provisioning_intro.tex
@@ -14,6 +14,7 @@ of a common base package followed by the Warewulf provisioning system.
 # Install base meta-packages
 [sms](*\#*) (*\install*) ohpc-base
 [sms](*\#*) (*\install*) ohpc-warewulf
+[sms](*\#*) (*\install*) hwloc-ohpc
 \end{lstlisting}
 % end_ohpc_run
 


adrianreber:
Thanks for the PR.

@koomie what do you think

docs/recipes/install/common/install_provisioning_intro.tex
```diff
@@ -12,7 +12,7 @@
 % ohpc_comment_header Add baseline OpenHPC and provisioning services \ref{sec:add_provisioning}
 \begin{lstlisting}[language=bash,keywords={}]
 # Install base meta-packages
-[sms](*\#*) (*\install*) ohpc-base
+[sms](*\#*) (*\install*) ohpc-base hwloc-ohpc
```
martin-g: nit: I am not sure whether it is intentional or not but currently all packages are installed one by one, e.g. 
```
dnf install p1
dnf install p2
```

I am not 100% certain but I think this change would introduce an inconsistency (`dnf install p1 p2`).

middelkoopt: Updated in new push as requested.

adrianreber:
No fixup commits in pull requests please.

middelkoopt:
Thanks for the suggestion.  Commits squashed and re-pushed.

martin-g:
@adrianreber CI job `build_on_centos_stream_9` should be removed from 2.x, right ?

adrianreber:
> @adrianreber CI job `build_on_centos_stream_9` should be removed from 2.x, right ?

Correct

martin-g:
Please rebase to latest 2.x to fix the CI failure.

middelkoopt:
Rebased to latest 2.x

adrianreber:
Thanks for the fix. Can you also create a PR against the 3.x branch?

middelkoopt:
I'm working on testing this against 3.x to see if the issue is there as well but not sure how much luck I will have, I have yet to figure out the build process for testing.  Would you want me to assume this issue also is in the 3.x branch and create a new PR? (perhaps a base package could be updated to require the hwloc package).

adrianreber:
Good point. We could also solve this with changes to the packaging. That would not require documentation changes.

Maybe using  soft dependencies in RPM would work.

middelkoopt:
I was able to build off the 3.x documentation and the build-farm and confirmed the issue exists.  Should I update the 3.x branch or add a dependency in the RPM?  If the dep is the case I will need to figure out how to access and trigger the CI/CD/build-farm pipelines.

martin-g:
If you change any of the `components/**/*.spec` files then the OpenHPC CI will run the relevant builds and tests.

middelkoopt:
Will do - I will just do the 3.x branch - hope to get to it in a few weeks.

PR #1828, opened 2023-05-15
Allow for the conditional building of serial versions fo the netcdf p…

opoplawski:
…ackages

This replaces #1752 as requested with conditional versions.

I'm also going to try to work up a version with some additional macros...

PR #1829, opened 2023-05-15
2.x serial conditional macros

opoplawski:
Version of #1828 with some additional macros.


adrianreber:
Do you want to rework this on top of 3.x or should we close it. The changes are in both branches, but not with your nice macros.

PR #1830, opened 2023-05-18
Filter out Slurm settings which have custom/specific values for OHPC builds

martin-g:
Currently Slurm startup prints:
```
[root@c3 ~]# systemctl status slurmd
● slurmd.service - Slurm node daemon
     Loaded: loaded (/usr/lib/systemd/system/slurmd.service; disabled; vendor preset: disabled)
     Active: active (running) since Thu 2023-05-18 15:40:52 CST; 48min ago
   Main PID: 3397 (slurmd)
      Tasks: 1
     Memory: 3.6M
     CGroup: /system.slice/slurmd.service
             └─3397 /usr/sbin/slurmd -D -s --conf-server 192.168.1.220

May 18 15:40:52 c3 systemd[1]: Started Slurm node daemon.
May 18 15:40:52 c3 slurmd[3397]: slurmd: error: TaskPlugin 1 specified more than once, latest value used
May 18 15:40:52 c3 slurmd[3397]: slurmd: error: JobCompType 1 specified more than once, latest value used
May 18 15:40:52 c3 slurmd[3397]: error: TaskPlugin 1 specified more than once, latest value used
May 18 15:40:52 c3 slurmd[3397]: error: JobCompType 1 specified more than once, latest value used
```

### DIFF
--- a/components/rms/slurm/SPECS/slurm.spec
+++ b/components/rms/slurm/SPECS/slurm.spec
@@ -484,7 +484,7 @@ install -D -m755 contribs/sjstat %{buildroot}/%{_bindir}/sjstat
 
 # 9/8/14 karl.w.schulz@intel.com - provide starting config file
 %if 0%{?OHPC_BUILD}
-head -n -2 $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.example | grep -v ReturnToService > $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.ohpc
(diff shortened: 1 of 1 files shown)


adrianreber:
Thanks for the PR, but the original line doesn't make much sense:
```
$ head -n 2 /etc/slurm/slurm.conf.example | grep -v ReturnToService 
#
# Example slurm.conf file. Please run configurator.html
```

So this change doesn't do anything if I understand it correctly. But it would be nice to fix the errors you are seeing.

martin-g:
Hi @adrianreber !

Actually the parameter is `-2`, not `2`, i.e. it ignores the last two lines, not the first two.

The idea is to ignore all Slurm settings which are later set explicitly by OpenHPC:
https://github.com/openhpc/ohpc/blob/15dab81c0d831e768c50d9e5e1e7d98ea202057b/components/rms/slurm/SPECS/slurm.spec#L490-L504

adrianreber:
Thanks for clearing that up. Makes much more sense. At this point this is really hard to read. What do you think using `patch` or maybe `sed` to edit the file in one go?

martin-g:
Using `sed` looks the same as what this PR does.
I will redo it with a .patch !

martin-g:
Closing in favour of https://github.com/openhpc/ohpc/pull/1909

PR #1831, opened 2023-06-03
EasyBuild v4.7.2

boegel:
https://pypi.org/project/easybuild/4.7.2 + https://docs.easybuild.io/release-notes/#release_notes_eb472


adrianreber:
If you are motivated you could also open the same PR against the 2.x branch.

PR #1832, opened 2023-06-03
EasyBuild v4.7.2

boegel:


boegel:
@adrianreber Should I also add `python3-devel` to `BuildRequires` here, as was done in #1784 for `3.x` branch with the EasyBuild 4.7.1 update?

adrianreber:
Probably a good idea to keep it in sync. We needed it in the 3.x branch, not sure if it is necessary for 2.x but it shouldn't hurt.

PR #1844, opened 2023-07-11
warewulf-provision: apply fix from upstream

adrianreber:


components/provisioning/warewulf-provision/SPECS/warewulf-provision.spec
```diff
@@ -32,6 +32,7 @@ Patch4:  warewulf-provision.sle_tftpboot.patch
 Patch5:  warewulf-provision.wwgetfiles.patch
 Patch6:  warewulf-provision.update_ipxe_to_09e8a15.patch
 Patch7:  warewulf-provision.zstd.patch
+Patch8:  https://github.com/warewulf/warewulf3/commit/ecb8ece02892caf2cbd3b8d74a1f3e01052b9844.patch
```
martin-g, line 35: Is the new file (components/provisioning/warewulf-provision/SOURCES/ecb8ece02892caf2cbd3b8d74a1f3e01052b9844.patch) needed if the url here is absolute & remote ?

adrianreber: I think it is necessary for OBS and also our CI. I use it mainly as documentation from where we have the patch. If it is a patch form upstream without changes it helps in the future.

PR #1848, opened 2023-07-24
incremeting pmix to latest release (4.2.4) and enabled pmix in openmpi

ohpcsim:
This is a simplified PR for re-enabling PMIX for OpenMPI as well as updating pmix to the latest release.

### DIFF
--- a/components/mpi-families/openmpi/SPECS/openmpi.spec
+++ b/components/mpi-families/openmpi/SPECS/openmpi.spec
@@ -32,7 +32,7 @@
 %{!?with_lustre: %define with_lustre 0}
 %{!?with_slurm: %define with_slurm 0}
 %{!?with_tm: %global with_tm 1}
-%{!?with_pmix: %define with_pmix 0}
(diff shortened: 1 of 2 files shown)


adrianreber:
The slurm tests are failing with:
```
# srun: error: Couldn't find the specified plugin name for mpi/pmix looking at all files
# srun: error: cannot find mpi plugin for mpi/pmix
# srun: error: MPI: Cannot create context for mpi/pmix
# srun: error: MPI: Unable to load any plugin
# srun: error: Invalid MPI type 'pmix', --mpi=list for acceptable types
```
Does this change also require a rebuild of slurm? Do we need any changes in slurm or just a rebuild?

ohpcsim:
Yeah this has a circular dependency (as we discussed in Slack) that won't work on Github actions. On my local test system, I have to load the pmix module and then rebuild SLURM for it to pick it up. 

I ****think**** we will need to merge and send this to OBS to have SLURM build updated before OpenMPI with pmix will pass any tests. This goes for this PR and the duplicate one for the 2.x branch

PR #1852, opened 2023-08-02
mpi-families/openmpi: build two Open MPI variants

adrianreber:
Build a Open MPI variant with and without PMIX support. With PMIX for slurm and without for OpenPBS.


martin-g:
Weird! Now the RHEL tests also started failing ...

adrianreber:
> Weird! Now the RHEL tests also started failing ...

Yes, it is driving me crazy. Not sure what happens. The next approach is to have a Open MPI version with and without PMIX support.

PR #1853, opened 2023-08-02
Exclude Intel libraries from RPM requires (2.x)

opoplawski:
OpenHPC packages should not explicitly require Intel runtime libraries.  See https://lists.openhpc.community/g/devel/message/318


adrianreber:
Just FYI: this required a little fixup: https://github.com/openhpc/ohpc/commit/e0c945e9d2239747f5c242871663f38538a45445

PR #1854, opened 2023-08-02
Exclude Intel libraries from RPM requires

opoplawski:
OpenHPC packages should not explicitly require Intel runtime libraries.  See https://lists.openhpc.community/g/devel/message/318


adrianreber:
Cherry-picked your changes from the 2.x branch including necessary fixups:

* fa54f4968
* d591c0c16 
* 8318b817a

PR #1855, opened 2023-08-09
Bump openEuler Docker images to 22.03 LTS SP2

martin-g:
Use latest service pack of 22.03 LTS.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,7 +30,7 @@ openeuler_task:
   name: openEuler on aarch64
   timeout_in: 120m
   arm_container:
-    image: docker.io/openeuler/openeuler:22.03-lts
+    image: docker.io/openeuler/openeuler:22.03-lts-sp2
     cpu: 4
     memory: 12G
   script: uname -a
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -136,7 +136,7 @@ jobs:
     runs-on: ubuntu-latest
     name: Build on openEuler
     container:
-      image: docker.io/openeuler/openeuler:22.03-lts
+      image: docker.io/openeuler/openeuler:22.03-lts-sp2
     steps:
     - name: Switch repo
       run: sed -i "s@repo.openeuler.org@repo.huaweicloud.com/openeuler@g" /etc/yum.repos.d/openEuler.repo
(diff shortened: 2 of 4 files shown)


martin-g:
An issue that I just found is that all OHPC packages for openEuler 22.03 LTS (SP0) depend on libicu 69 while openEuler 22.03 SP2 provides `libicu 72` which leads to issues like:
```
root@euler ~# dnf install boost-gnu12-mpich-ohpc
Last metadata expiration check: 1:31:52 ago on Wed 09 Aug 2023 11:57:40 AM UTC.
Error: 
 Problem: cannot install the best candidate for the job
  - nothing provides libicui18n.so.69()(64bit) needed by boost-gnu12-mpich-ohpc-1.81.0-300.ohpc.2.4.x86_64
  - nothing provides libicuuc.so.69()(64bit) needed by boost-gnu12-mpich-ohpc-1.81.0-300.ohpc.2.4.x86_64
  - nothing provides libicudata.so.69()(64bit) needed by boost-gnu12-mpich-ohpc-1.81.0-300.ohpc.2.4.x86_64
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

I.e. if we use SP2 then we will need to re-build the OHPC packages for openEuler.
Needs discussion at the TSC meeting!

adrianreber:
That is kind of unexpected that openeuler breaks abi between service packs. That is what I would expect from a new major release.

At this point I don't think a rebuild makes sense.

Can openeuler introduce a compatibility package to not break Abi?

martin-g:
We have reported the problem with the ABI break ! Let's see what could be done!

Assuming that there is an ABI break: It should not be a problem for OpenHPC 3.1, right ?

adrianreber:
> We have reported the problem with the ABI break ! Let's see what could be done!
> 
> Assuming that there is an ABI break: It should not be a problem for OpenHPC 3.1, right ?

No. It is just bad timing at this point.

martin-g:
openEuler 22.03 SP2 will wait for OpenHPC 3.1, but in case someone needs to use OpenHPC 3.0 on SP2 here is a "transitional" package that makes it possible to install 3.0:
- https://github.com/bio-arm/rpms/blob/libicu-69-transitional/libicu-69-transitional/libicu-69-transitional.spec
- https://eur.openeuler.openatom.cn/coprs/mgrigorov/Test/build/91052/

It is a hack that just copies `.so.72` to `.so.69`.
I didn't make full testing, so it may blow up at runtime!

adrianreber:
I guess this can be closed.

martin-g:
We updated openEuler to 22.03 LTS SP3 for OpenHPC 3.1 with:
* https://github.com/openhpc/ohpc/pull/1938
* https://github.com/openhpc/ohpc/pull/1926

PR #1858, opened 2023-08-13
release: Add SLE_HPC to supported SUSE releases

e4t:


adrianreber:
The test failure is because the release package needs additional macros for building, so that is not a blocker.

Any reasons why this change is important for the 3.0 release. Just asking if we need this or not. At this point we are trying to avoid changes as much as possible before the 3.0 release 

Can this wait for 3.1?

e4t:
Hi Adrian,
I'd appreciate if this fix could make it into 3.0. 
Otherwise, one will get an error message when trying to use this with the HPC variant of SLE. This error message can be circumvented, but it is irritating to users. This change should not have any adverse effects on the release. Thanks!

e4t:
Thank!

adrianreber:
It has been built by OBS and should be soon be used for the test setups:

http://test.openhpc.community:8080/job/3.x/

PR #1860, opened 2023-08-16
Fix the urls to the openEuler repos in the documentation.

martin-g:
Use LaTeX "verbatim" instead of "itemize" - it should be easier for the reader to copy/paste.

The itemized urls were wrong (giving 404) because of the extra `openeuler/" in the middle of the urls.

### DIFF
--- a/docs/recipes/install/common/openeuler_repos.tex
+++ b/docs/recipes/install/common/openeuler_repos.tex
@@ -8,17 +8,17 @@ for which mirrors are freely available online:
 
 \begin{itemize*}
 \item OS
-  (e.g. \href{http://repo.openeuler.org/openeuler/openEuler-22.03-LTS/OS/}
-             {\color{blue}{http://repo.openeuler.org/openeuler/openEuler-22.03-LTS/OS/}} )
+  (e.g. \href{http://repo.openeuler.org/openEuler-22.03-LTS/OS/}
+            {\color{blue}{http://repo.openeuler.org/openEuler-22.03-LTS/OS/}} )
 \item Everything
-  (e.g. \href{http://repo.openeuler.org/openeuler/openEuler-22.03-LTS/everything/}
-            {\color{blue}{repo.openeuler.org/openeuler/openEuler-22.03-LTS/everything/}} )
+  (e.g. \href{http://repo.openeuler.org/openEuler-22.03-LTS/everything/}
+            {\color{blue}{http://repo.openeuler.org/openEuler-22.03-LTS/everything/}} )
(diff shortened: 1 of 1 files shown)


adrianreber:
I don't think this is the right way. The recipes contain all the commands to set up a cluster. That seems to work, because Jenkins tells us that it works. All things users should type should be part of the recipe and not part of the PDF.

From my point of view this is not the correct change. If the EPOL repository is needed I think a `-release` package would be good to have (or something like that). The same way we enable EPEL in the RHEL ecosystem.

You are right that we need the EPOL repository somewhere mentioned in the documentation because in Jenkins I am manually adding that repository and that is not correct, because that means users will not get it. The recipes need some way to enable it on the SMS and on the computes and the best way would be a EPOL release package.

docs/recipes/install/common/openeuler_repos.tex
```diff
@@ -8,17 +8,17 @@
 
 \begin{itemize*}
 \item OS
-  (e.g. \href{http://repo.openeuler.org/openeuler/openEuler-22.03-LTS/OS/}
-             {\color{blue}{http://repo.openeuler.org/openeuler/openEuler-22.03-LTS/OS/}} )
+  (e.g. \href{http://repo.openeuler.org/openEuler-22.03-LTS/OS/}
+              {\color{blue}{http://repo.openeuler.org/openEuler-22.03-LTS/OS/}} )
```
adrianreber: There is a single space added which seems unnecessary

adrianreber:
Please update the commit message. Besides that it looks correct.

docs/recipes/install/common/openeuler_repos.tex
martin-g: Fixed! Thanks!

martin-g:
> Please update the commit message.

Updated!

koomie:
What about the actual instructions to enable the repository in the recipes? It's nice to mention in the text but as @adrianreber mentioned, that is not sufficient for a user to reproduce.  

I'm surprised this necessary repo isn't at least defined in the openEuler-repos package. Even if it was defined, but not enabled, it would be easy to add the enablement within the ohpc recipe.

I guess another possibility is to add the required repo definition as part of the ohpc-release package for openeuler, but that feels pretty hackish.

adrianreber:
@martin-g and I are discussing a way to enable it properly. There will be an rpm to enable that repository.

martin-g:
Something like this: https://github.com/openhpc/ohpc/pull/1862 ?

PR #1867, opened 2023-09-10
3.0 ChangeLog update

adrianreber:


martin-g (approved):

docs/ChangeLog
```diff
@@ -1,5 +1,38 @@
 # -*- mode: sh; fill-column: 120; -*-
 
+Version 3.0 (15 September 2023)
+
+[Important Highlights/Notices for 3.x series]
+
+  * the 3.x series targets support for three new major OS distro versions: EL 9, openSUSE Leap 15.5 and openEuler 22.03.
+     o As the OpenHPC 3.x series targets major new distro versions, please note that the 3.x series
```
martin-g, line 8:
```suggestion
     o As the OpenHPC 3.x series targets new major distro versions, please note that the 3.x series
```

docs/ChangeLog
```diff
@@ -1,5 +1,38 @@
...
+       is **not** backwards compatible with the OpenHPC 2.x series.
+     o OpenHPC 3.x is intended for fresh installs with the newer distro versions (or containers based
+       on these distro versions).
+     o Note that 3.x builds for SUSE based systems are now performed against Leap instead of SLES. Leap 15 is
+       intended to be compatible with SLES 15.
+     o EL 9 based builds are performed against RHEL 9.2 and tested against Rocky Linux 9.2 and AlmaLinux 9.2
```
martin-g, line 14:
```suggestion
     o EL 9 based builds are performed against RHEL 9.2 and tested against Rocky Linux 9.2 and AlmaLinux 9.2
     o OpenHPC 3.x is the first release that supports openEuler OS! It is tested against openEuler 22.03 LTS (SP0)
```

martin-g: Here I try to emphasize that OpenHPC 3.0 is the **first** release with support for openEuler.
Also I think we should make it somehow known that there are known issues with 22.03 LTS SP1 and SP2 (related to the ABI breaks in libicu)

adrianreber:
@martin-g please also add a sentence to the know issues section about the openEuler 22.03 SP2 situation. Maybe also add a link where to download the older release from.

docs/ChangeLog
```diff
@@ -1,5 +1,38 @@
...
+  * Tech preview build of Warewulf 4.x.
+  * Switched to clang based Intel oneAPI compiler backend.
+  * Removed deprecated ADIOS package and introduced ADIOS2.
+  * Introduced two Open MPI variants. One with PMIX support and one without PMIX support.
+    The Open MPI variant with PMIX support is used in the slurm based recipes.
+  * Introduced new compiler variant gnu13 as tech preview.
+
+[Known Issues]
+
+ * AlmaLinux recipes only tested on aarch64
```
adrianreber, line 29: sounds good. thanks

Yikun: Just FYI, there are [3 levels ABI compatibility](https://github.com/openeuler-mirror/oec-application/blob/master/doc/compatibility_level.md) in openEuler [1].

- Level 1: ABI compatibility in LTS
- Level 2: ABI compatibility in SP
- Level 3: no gurantee

For icu, it's a level 3 package. that is why ABI breaks happening

> Only openEuler 22.03 LTS (aka SP0) is supported! Due to ABI breaks in the `libicu` package in 22.03 SP1 and SP2 the installation of the `-ohpc` packages would fail! Please check https://github.com/openhpc/ohpc/pull/1855 for more information and possible workaround.

I'm not sure document the known issue for untest version is right or not. Actually, we only test 22.03 LTS, I believe if we want to support `22.03 SP1/SP2` until we need test it, that is SP1/SP2 will not be supported in 3.0.

BTW, openEuler 22.03 LTS SP2 supported will be introduced in 3.1 is the plan, that means after 3.1 version we will only support 22.03 LTS SP2 ("build / test based on SP2" rather than "build on LTS and test on SP2").

[1] https://github.com/openeuler-mirror/oec-application/blob/master/doc/compatibility_level.md

So maybe `Only openEuler 22.03 LTS (aka SP0) is supported!` is enough.

martin-g: I discussed the topic with @Yikun privately and he explained to me how the `SP`s work in openEuler. 
It appears that each SP has its own lifecycle.
We agreed to remove this entry from the `known issues`.

ohpcsim:
Thanks guys; looks good

PR #1868, opened 2023-09-11
EasyBuild v4.8.1

boegel:
https://pypi.org/project/easybuild/4.8.1


adrianreber:
Thanks. We will  not include this in 3.0 but in 3.1.

boegel:
> Thanks. We will not include this in 3.0 but in 3.1.

Any particular reason for that? Too close to 3.0 release? (just wondering, I don't mind)

adrianreber:
> > Thanks. We will not include this in 3.0 but in 3.1.
> 
> Any particular reason for that? Too close to 3.0 release? (just wondering, I don't mind)

Too close to the 3.0 release. At this point we do not want any changes if possible.

PR #1872, opened 2023-09-19
updating rocky and alama recipes to the proper name for the CRB repo

ohpcsim:
Actual repo for the yum config-manager command uses lowercase for crb and not uppercase.


martin-g (approved):

adrianreber:
Thanks. Just verified it locally and this looks correct.

PR #1874, opened 2023-09-21
Switch oneAPI default module versions from softlinks to .version files

jcsiadal:
Very minor update to compiler-devel. This changes how the oneAPI version defaults are defined. Switching from "default" symlinks to ".version" files makes it much simpler for users to override defaults. The .version files are lowest priority for setting defaults.

This won't impact any previously-built packages, since the default versions remain the same.


adrianreber:
Thanks. The important change was that `intel-mpi` now also sets the defaults. In CI I saw that the defaults were set by the intel-compilers-devel package and then the default was removed once intel-mpi was installed.

PR #1878, opened 2023-09-27
ci: Support for building docs-ohpc.rpm at simple CI

martin-g:
Create dosc-ohpc.tar on the fly.
Clones the complete Git repo to be able to run `git describe` later in the Makefile`s

### DIFF
--- a/components/admin/docs/SOURCES/get_source.sh
+++ b/components/admin/docs/SOURCES/get_source.sh
@@ -1,15 +1,16 @@
 #!/usr/bin/env bash
 
+set -x
+
 # Note: The current working directory is the component's SPEC folder
+OHPC_ROOT_FOLDER="../../../.."
 
 # The folder that will contain the .git/ and docs/ folders
 DOCS_OHPC_FOLDER=docs-ohpc
-# A folder that contains the full Git history of the project
-# Full history is needed for the usage of `git archive` command
-COMPLETE_OHPC_FOLDER=$(mktemp -d)
 
 cleanup() {
-    rm -rf ${DOCS_OHPC_FOLDER} ${COMPLETE_OHPC_FOLDER}
+    rm -rf ${DOCS_OHPC_FOLDER}
+    find ${OHPC_ROOT_FOLDER}/docs/recipes/install/ -name "vc.tex" -delete
 }
 trap cleanup EXIT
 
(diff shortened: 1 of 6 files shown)


martin-g:
Successful CI results could be seen in my fork: https://github.com/martin-g/openeuler-ohpc/actions/runs/6326193045
I modified temporarily the docs.spec to trigger the build there.

adrianreber:
Good idea, but it should trigger on any Tex file changed in the repository. The script looks a bit hacky copying the whole .git history of another checkout into the PR.

martin-g:
> The script looks a bit hacky copying the whole .git history of another checkout into the PR.

Yep! Without this hack this error is thrown: https://github.com/openhpc/ohpc/blob/3.x/docs/recipes/install/almalinux9/aarch64/warewulf/slurm/Makefile#L12

Another solution is to fetch the whole history with `actions/checkout` action, but then the toll is that it will be done for all PRs. The current solution fetches the whole history only when docs.spec is changed.

martin-g:
> it should trigger on any Tex file changed in the repository

I'd say for anything in `docs/` folder. I will see how to do it!

martin-g:
@adrianreber Please review again! Changing a .tex file (without docs.spec) works!
I just checked the uploaded rpm for openEuler. The Warewulf+OpenPBS manual contains 
```
TESTING PR 1878
THIS IS A TEMPORARY CHANGE TO TEST THE BUILD OF DOCS WHEN A .TEX FILE IS MODI-
FIED
```
I will revert the last commit once you confirm!

tests/ci/run_build.py
```diff
@@ -237,9 +237,17 @@ def build_srpm_and_rpm(
...
 total = 0
+docs_spec_executed = False
 
 for spec in args.specfiles:
-    if not spec.endswith('.spec'):
+    if 'components/admin/docs/SPECS/docs.spec' == spec:
+        if docs_spec_executed:
+            continue
+        docs_spec_executed = True
+    elif not docs_spec_executed and 'docs/recipes/install/' in spec:
```
martin-g, line 247: Should I add logic for the get_source.sh plugin too ?
```suggestion
    elif not docs_spec_executed and ('docs/recipes/install/' in spec or 'components/admin/docs/SOURCES/` in spec):
```

docs/recipes/install/openeuler22.03/aarch64/warewulf/openpbs/steps.tex
```diff
@@ -277,6 +277,8 @@ \section{Run a Test Job} \label{sec:test_job}
 \input{manifest}
 \input{common/signature}
 
+\section{TESTING PR 1878}
+THIS IS A TEMPORARY CHANGE TO TEST THE BUILD OF DOCS WHEN A .TEX FILE IS MODIFIED
```
martin-g: This has to be removed before merge!

adrianreber:
@martin-g If you want, let's get this ready.

martin-g:
@adrianreber PTAL!

adrianreber:
Please rebase the latex error should be fixed.

components/admin/docs/SOURCES/get_source.sh
```diff
@@ -0,0 +1,30 @@
...
+cleanup() {
+    rm -rf ${DOCS_OHPC_FOLDER} ${COMPLETE_OHPC_FOLDER}
+}
+
+# 1. Prepare
+cleanup
+mkdir -p ${DOCS_OHPC_FOLDER}
+
+# 2. Fetch the full Git history so that `git describe` works later
+git clone https://github.com/openhpc/ohpc ${COMPLETE_OHPC_FOLDER}
```
adrianreber: Would it be possible to detect if it is currently running as a git checkout and then just use the existing git checkout instead of doing a remote clone? Or maybe a local clone? Or just

components/admin/docs/SOURCES/get_source.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+
+# Note: The current working directory is the component's SPEC folder
+
+# The folder that will contain the .git/ and docs/ folders
+DOCS_OHPC_FOLDER=docs-ohpc
+# A folder that contains the full Git history of the project
+# Full history is needed for the usage of `git archive` command
+COMPLETE_OHPC_FOLDER=/tmp/ohpc-for-docs
```
adrianreber: Please use `mktemp -d`.

components/admin/docs/SOURCES/get_source.sh
```diff
@@ -0,0 +1,30 @@
...
+cp -r ${COMPLETE_OHPC_FOLDER}/.git ${DOCS_OHPC_FOLDER}
+
+# 3. Copy the local docs/
+cp -r ../../../../docs ${DOCS_OHPC_FOLDER}
+
+# 4. Create docs-ohpc.tar
+tar cf ../SOURCES/docs-ohpc.tar ${DOCS_OHPC_FOLDER}
+
+# 5. Cleanup
+cleanup
```
adrianreber: Maybe use `trap` to ensure cleanup will always run.

tests/ci/run_build.py
```diff
@@ -237,9 +237,19 @@ def build_srpm_and_rpm(
 failed = []
 rebuild_success = []
 total = 0
+docs_spec_executed = False
 
 for spec in args.specfiles:
-    if not spec.endswith('.spec'):
+    if 'components/admin/docs/SPECS/docs.spec' == spec:
```
adrianreber, line 246: Ah comment here would be nice. My understanding is that this tries to do a test build if any file changed in the `docs/` tree, right?

components/admin/docs/SOURCES/get_source.sh
martin-g: The problem is that the CI jobs fetches just the last commit (https://github.com/openhpc/ohpc/blob/5ffbf3e2ed0ed558c1ae5672f7e5023298b7c2a8/.github/workflows/validate.yml#L31C13-L31C29) and later `git describe` fails (https://github.com/openhpc/ohpc/blob/5ffbf3e2ed0ed558c1ae5672f7e5023298b7c2a8/docs/recipes/install/rocky9/aarch64/warewulf/slurm/Makefile#L3)

We can use the local checkout if we fetch the whole history in the CI workflows (`fetch-depth: 0`).

components/admin/docs/SOURCES/get_source.sh
martin-g: Done!

components/admin/docs/SOURCES/get_source.sh
martin-g: Done!

tests/ci/run_build.py
martin-g: The idea is to not build the docs more than once if `docs.spec`, `components/admin/docs/SOURCES/**` and any file in `docs/recipes/install/**` are modified

components/admin/docs/SOURCES/get_source.sh
adrianreber: Can we do a workaround? Maybe create an empty `vc.tex` file. Just trying to find a way around...

PR #1879, opened 2023-09-27
CI: use official repository for CI

adrianreber:


adrianreber:
Hmm, CentOS Stream does not handle "redhat-release >= 9.1". Need to figure out what to do.

adrianreber:
The CentOS Stream error is discussed here: https://issues.redhat.com/browse/CS-1757

Not sure we actually need to require a certain minimum version of the OS...

PR #1880, opened 2023-09-28
ci: also build and test with the Intel compiler

adrianreber:


adrianreber:
For now just as a test.

adrianreber:
This fails because there is not enough disk space on the GitHub runners.

martin-g:
openEuler and Leap fail due to:
```
No provider of 'mpich-.github/workflows/test-analysis.yml-ohpc' found.
No provider of 'openmpi4-.github/workflows/test-analysis.yml-ohpc' found.
```

martin-g:
now the problem is that the test jobs need `man` to be installed

adrianreber:
Now the Intel signing key seems broken.

PR #1887, opened 2023-10-20
ci: handle lenovo ipmi CPU temperature

adrianreber:


tests/oob/ipmitool
```diff
@@ -99,7 +99,7 @@ fi
     
     # read sensor data locally
     ipmitool -E -I lanplus -H $my_bmc_ip -U root sensor > .cmd_output || flunk "Unable to query sensor data"
-    egrep -q "CPU1 Temp|P1 Therm Margin" .cmd_output   || flunk "CPU1 Temp not found in sensor data"
+    egrep -q "CPU1 Temp|P1 Therm Margin|CPU 1 Temp" .cmd_output   || flunk "CPU1 Temp not found in sensor data"
```
martin-g: Alternative:
```suggestion
    egrep -q "CPU\s?1 Temp|P1 Therm Margin" .cmd_output   || flunk "CPU1 Temp not found in sensor data"
```
I could not say whether it is simpler or not.

martin-g (approved):

tests/oob/ipmitool
adrianreber: I also was thinking about it. I will use your version. Good idea.

Issue #1905, opened 2023-11-02
Please update lmod to 8.7.31 to support Intel oneAPI debugger 2023.2.0 module file


opoplawski:
When trying to load the Intel debugger module I'm getting:
```
Loading debugger version 2023.2.0
Lmod has detected the following error:
/opt/intel/oneapi/modulefiles/debugger/2023.2.0: (debugger/2023.2.0): invalid command name
"getenv"
While processing the following module(s):
    Module fullname    Module Filename
    ---------------    ---------------
    debugger/2023.2.0  /opt/intel/oneapi/modulefiles/debugger/2023.2.0
```
Support for getenv was added to Lmod in version 8.7.31.

adrianreber:
Thanks for the report. We have been aware of the problems but we didn't know that a newer lmod version fixes it.

@jcsiadal @mghpcsim fyi

adrianreber:
2.7 will have 8.7.32. I just updated it and the first test runs are without errors: 
http://test.openhpc.community:8080/job/2.x/job/(2.7)%20-%20(rocky8.8,x86_64)%20(warewulf+slurm)%20(fabric=eth)/15/

@jcsiadal does this mean we can use 2023.2.0 or 2023.2.1 for 2.7? If we can add the necessary changes today we can get a rebuild for the 2.7 branch and still have enough time to test next week.

adrianreber:
Related: #1907

With the latest lmod I can install and load 2023.2.0 locally in a test container. So the lmod update really helps.

adrianreber:
2.7 release is out with the latest version of Lmod. I will keep this open until we also have it in the 3.x branch

adrianreber:
3.1 will have 8.7.34

PR #1906, opened 2023-11-03
Update lmod to 8.7.32

adrianreber:
See: #1905

### DIFF
--- a/components/admin/lmod/SPECS/lmod.spec
+++ b/components/admin/lmod/SPECS/lmod.spec
@@ -16,7 +16,7 @@
 
 Summary:   Lua based Modules (lmod)
 Name:      %{pname}%{PROJ_DELIM}
-Version:   8.7.14
+Version:   8.7.32
 Release:   %{?dist}.1
 License:   MIT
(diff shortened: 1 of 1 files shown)


components/admin/lmod/SPECS/lmod.spec
```diff
@@ -97,7 +97,7 @@ make DESTDIR=$RPM_BUILD_ROOT install
 ########################################################################
 
 # NOOP if running under known resource manager
-if [ ! -z "\$SLURM_NODELIST" ] | [ ! -z "\$PBS_NODEFILE" ]; then
+if [ -n "\$SLURM_NODELIST" ] || [ -n "\$PBS_NODEFILE" ]; then
```
martin-g, line 100: Shouldn't `||` be `&&` ? I.e. if neither is defined then return early.

adrianreber: Hmm, good question. Let me take a look.

adrianreber: Just tested a couple of combinations and it works as expected. The goal is, if one is set it should return.

PR #1907, opened 2023-11-03
intel-compilers-devel: up minimum version to 2023.2

adrianreber:


adrianreber:
@jcsiadal PTAL. I tried to up the minimum version to 2023.2. This seems to avoid the signing errors and the newest version of lmod also seems to handle everything correctly.

PR #1909, opened 2023-11-23
Use a .patch to comment out Slurm settings which need to be overwritten by OpenHPC

martin-g:
Related-to: https://github.com/openhpc/ohpc/pull/1830

### DIFF
--- a/components/rms/slurm/SOURCES/slurm.conf.example.patch
+++ b/components/rms/slurm/SOURCES/slurm.conf.example.patch
@@ -1,0 +1,45 @@
+--- orig/etc/slurm.conf.example	2023-10-11 22:23:30.000000000 +0300
++++ new/etc/slurm.conf.ohpc	2023-11-24 15:01:18.989659566 +0200
+@@ -57,7 +57,7 @@
+ StateSaveLocation=/var/spool/slurmctld
+ SwitchType=switch/none
+ #TaskEpilog=
+-TaskPlugin=task/affinity
++#TaskPlugin=task/affinity
+ #TaskProlog=
+ #TopologyPlugin=topology/tree
+ #TmpFS=/tmp
+@@ -122,7 +122,7 @@
+ #JobCompLoc=
+ #JobCompPass=
+ #JobCompPort=
+-JobCompType=jobcomp/none
++#JobCompType=jobcomp/none
+ #JobCompUser=
+ #JobContainerType=job_container/none
+ JobAcctGatherFrequency=30
+@@ -149,5 +149,22 @@
+ #
+ #
+ # COMPUTE NODES
+-NodeName=linux[1-32] CPUs=1 State=UNKNOWN
+-PartitionName=debug Nodes=ALL Default=YES MaxTime=INFINITE State=UP
++#NodeName=linux[1-32] CPUs=1 State=UNKNOWN
++#PartitionName=debug Nodes=ALL Default=YES MaxTime=INFINITE State=UP
++
++
++
++# OpenHPC default configuration
++# Enable the task/affinity plugin to add the --cpu-bind option to srun for GEOPM
++TaskPlugin=task/affinity
++PropagateResourceLimitsExcept=MEMLOCK
++JobCompType=jobcomp/filetxt
++Epilog=/etc/slurm/slurm.epilog.clean
++NodeName=c[1-4] Sockets=2 CoresPerSocket=8 ThreadsPerCore=2 State=UNKNOWN
++PartitionName=normal Nodes=c[1-4] Default=YES MaxTime=24:00:00 State=UP Oversubscribe=EXCLUSIVE
++# Enable configless option
++SlurmctldParameters=enable_configless
++# Setup interactive jobs for salloc
++LaunchParameters=use_interactive_step
++# Adding ReturnToService Directive to starting config file (note removal of variable during above creation)
++ReturnToService=1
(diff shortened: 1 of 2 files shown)


adrianreber:
How do you feel about having the changes from all the `echo` lines also be included in the patch?

martin-g:
It will work! 
I just wonder how often we will need to adapt the patch when upgrading Slurm.
Looking at the Git history of https://github.com/SchedMD/slurm/commits/master/etc/slurm.conf.example I see that they make 1 or 2 changes per year. But most of the changes are in settings' values, not adding/removing settings (and thus change the line numbers). So probably it should be OK!

adrianreber:
Right, having a patch is more work if upstream changes things. On the other hand it is probably more reliable to actually change the file as desired. `grep`, `sed`, `echo` will probably not give us feedback if something failed.

martin-g:
OK! I will move those lines to the patch!

components/rms/slurm/SPECS/slurm.spec
```diff
@@ -487,21 +490,7 @@ install -D -m755 contribs/sjstat %{buildroot}/%{_bindir}/sjstat
...
-echo "PartitionName=normal Nodes=c[1-4] Default=YES MaxTime=24:00:00 State=UP Oversubscribe=EXCLUSIVE" >> $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.ohpc
-# 5/5/2020 karl@oden.utexas.edu - enable configless option
-echo "SlurmctldParameters=enable_configless" >> $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.ohpc
-# 11/9/2021 karl@oden.utexas.edu - setup interactive jobs for salloc
-LaunchParameters=use_interactive_step >> $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.ohpc
-# 6/3/16 nirmalasrjn@gmail.com - Adding ReturnToService Directive to starting config file (note removal of variable during above creation)
-echo "ReturnToService=1" >> $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.ohpc
+cp $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.example $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.conf.ohpc
 # 9/17/14 karl.w.schulz@intel.com - Add option to drop VM cache during epilog
 sed -i '/^# No other SLURM jobs,/i \\n# Drop clean caches (OpenHPC)\necho 3 > /proc/sys/vm/drop_caches\n\n#' $RPM_BUILD_ROOT/%{_sysconfdir}/slurm.epilog.clean
```
martin-g: Here is an example of a silent error!
There is no `slurm.epilog.clean` at https://github.com/SchedMD/slurm/tree/slurm-22-05-10-1/etc, so this `sed` command does nothing!

martin-g: The file has been removed 5 years ago with https://github.com/SchedMD/slurm/commits/master/etc/slurm.epilog.clean

adrianreber: It still seems to exist:

```
[root@c1 ~]# rpm -qf /etc/slurm/slurm.epilog.clean
slurm-example-configs-ohpc-22.05.10-2.1.ohpc.2.6.2.x86_64
```

adrianreber: https://github.com/openhpc/ohpc/blob/15dab81c0d831e768c50d9e5e1e7d98ea202057b/components/rms/slurm/SPECS/slurm.spec#L485

martin-g: Yep!
I see now that it is copied in SOURCES/ - https://github.com/openhpc/ohpc/blob/3.x/components/rms/slurm/SOURCES/slurm.epilog.clean
All is good!
Except that we could move https://github.com/openhpc/ohpc/blob/15dab81c0d831e768c50d9e5e1e7d98ea202057b/components/rms/slurm/SPECS/slurm.spec#L506 directly in the file.

adrianreber: Right. No need to modify the file with sed.

components/rms/slurm/SOURCES/slurm.conf.example.patch
```diff
@@ -0,0 +1,45 @@
...
+ # COMPUTE NODES
+-NodeName=linux[1-32] CPUs=1 State=UNKNOWN
+-PartitionName=debug Nodes=ALL Default=YES MaxTime=INFINITE State=UP
++#NodeName=linux[1-32] CPUs=1 State=UNKNOWN
++#PartitionName=debug Nodes=ALL Default=YES MaxTime=INFINITE State=UP
++
++
++
++# OpenHPC default configuration
++# 10/2/18 brad.geltz@intel.com - Enabling the task/affinity plugin to add the --cpu-bind option to srun for GEOPM
```
adrianreber: Let's remove the dates and the email addresses from the file. We have the authors of the changes in git.

components/rms/slurm/SOURCES/slurm.epilog.clean
```diff
@@ -30,6 +30,10 @@ do
 	fi
 done
 
+# 9/17/14 karl.w.schulz@intel.com - Add option to drop VM cache during epilog
+# Drop clean caches (OpenHPC)
+echo 3 > /proc/sys/vm/drop_caches
```
adrianreber: Let's remove the changes to the epilog file.

martin-g: Done!

components/rms/slurm/SOURCES/slurm.conf.example.patch
martin-g: Done!

adrianreber:
Thanks

PR #1913, opened 2023-12-05
Remove hardcoded values for num_nodes, compute_prefix, allow overrides for input.local values

mikerenfro:
Things corrected:

- `slurm_node_config` in input.local was broken by default, since it referred to variables that hadn't necessarily been set
- some compute node attributes in input.local would accept existing environment variables, others wouldn't. Now they all do.
-  by default, there were duplicate MAC addresses for compute nodes in input.local -- granted, sites should change those values regardless, but there's no reason to throw an error by default
- more consistent use of curly braces for `compute_prefix` and `num_computes`


martin-g (approved):
LGTM!

adrianreber:
This actually breaks the build. I will do a fixup.

PR #1916, opened 2023-12-15
rms/slurm: Update to 22.05.11

martin-g:
See https://lists.schedmd.com/pipermail/slurm-announce/2023/000103.html 
Addresses CVE-2023-49933 through CVE-2023-49938

### DIFF
--- a/components/rms/slurm/SPECS/slurm.spec
+++ b/components/rms/slurm/SPECS/slurm.spec
@@ -21,7 +21,7 @@
 # $Id$
 #
 Name:		%{pname}%{PROJ_DELIM}
-Version:	22.05.10
+Version:	22.05.11
 %global rel	1
 Release:	%{?dist}.1
 Summary:	Slurm Workload Manager


adrianreber:
Can you also open one PR for 2.x?

martin-g:
@adrianreber Is this intentional https://github.com/openhpc/ohpc/blob/3.x/tests/ci/spec_to_test_mapping.py#L18 ?
When `Slurm` is changed then `munge` is enabled ?!

I guess it is intentional. Slurm is the only RM used in SIMPLE_CI (Github Actions and CircleCI), so it is tested in all cases, and `munge` is probably just a random component to test too.

adrianreber:
> @adrianreber Is this intentional https://github.com/openhpc/ohpc/blob/3.x/tests/ci/spec_to_test_mapping.py#L18 ?
> When `Slurm` is changed then `munge` is enabled ?!

That should be correct. Not totally sure, but slurm depends on munge. Cannot remember right now, but it doesn't sound totally wrong.

We are testing slurm anyway all the time. Maybe it would no longer be necessary.

sjpb:
> @adrianreber Is this intentional https://github.com/openhpc/ohpc/blob/3.x/tests/ci/spec_to_test_mapping.py#L18 ? When `Slurm` is changed then `munge` is enabled ?!
> 
> I guess it is intentional. Slurm is the only RM used in SIMPLE_CI (Github Actions and CircleCI), so it is tested in all cases, and `munge` is probably just a random component to test too.

I may have misunderstood but Slurm always requires Munge so sounds right to me.

PR #1922, opened 2024-02-22
Update openEuler from 22.03-LTS-SP0 to 22.03-LTS-SP3

martin-g:
OpenHPC 3.1 will build and test against openEuler 22.03-LTS-SP3

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -30,7 +30,7 @@ openeuler_task:
   name: openEuler on aarch64
   timeout_in: 120m
   arm_container:
-    image: docker.io/openeuler/openeuler:22.03-lts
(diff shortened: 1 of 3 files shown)


adrianreber:
Please exclude the warewulf changes for now. I would also say, if we change warewulf, maybe there should be an additional target next to 22.03. Not replacing the existing template but adding another one. Not sure maybe it doesn't make sense as the old template will not work with the new packages. Anyway, just do the CI changes in this PR.

martin-g:
Let's see how https://github.com/openhpc/ohpc/pull/1926 goes first.

martin-g:
The non-provisioning changes have been applied with https://github.com/openhpc/ohpc/pull/1926

PR #1933, opened 2024-02-26
python: use 3.11 Python stack on RHEL and LEAP

adrianreber:


adrianreber:
Second try. GitHub seems broken right now.

PR #1948, opened 2024-03-06
numpy: fix building with the Intel compiler

adrianreber:


adrianreber:
CI error is not reproducible. Merging to see if OBS like the change.

PR #1962, opened 2024-03-18
Use my user Coprs repository to install missing packages for openEuler

martin-g:
- EPOL update repo
- bats (and its dependencies)


adrianreber:
Thanks. I had a look at the resulting recipe and tried you changes and it works. Thanks.

PR #1964, opened 2024-03-19
slurm: update to 23.11.4

adrianreber:


adrianreber:
To resolve CI errors we need to rebuild pdsh against the new version of slurm.

PR #1975, opened 2024-04-15
add apparmor exception for lmod and lessopen.sh

### LINKED ISSUE #1974: Lmod init/bash errors
On OpenSUSE Leap 15.5, I have the following output from the terminal any time I exit from "less" viewing a file, and sometimes for other commands too. It doesn't appear to have any ill effects, but I'm not sure what could be causing it. 

```
/opt/ohpc/admin/lmod/lmod/init/bash: line 58: /usr/bin/readlink: Permission denied
/opt/ohpc/admin/lmod/lmod/init/bash: line 62: /usr/bin/expr: Permission denied
/opt/ohpc/admin/lmod/lmod/init/bash: line 63: /usr/bin/basename: Permission denied
```

Running the lines in the bash init script manually does not produce an error. The permissions on the above executables also appear normal.

mslacken:
close #1974

### DIFF
--- a/components/admin/lmod/SPECS/lmod.spec
+++ b/components/admin/lmod/SPECS/lmod.spec
@@ -184,6 +184,21 @@ EOF
 
 %{__ln_s} %{OHPC_ADMIN}/lmod/lmod/libexec/lmod %{buildroot}/%{_bindir}/modulecmd
 
+%post
+if [ -e %{_sysconfdir}/apparmor.d/local/usr.bin.lessopen.sh ] ; then
+%{__cat} << EOF >> %{_sysconfdir}/apparmor.d/local/usr.bin.lessopen.sh
+# allow lmod for OHPC-lmod
+  /usr/bin/readlink mrix, #OHPC-lmod
+  /usr/bin/expr mrix, #OHPC-lmod
+  /usr/bin/basename mrix, #OHPC-lmod
+EOF
+fi 
+
+%postun 
+if [ -e %{_sysconfdir}/apparmor.d/local/usr.bin.lessopen.sh ] ; then
+  sed -i '/OHPC-lmod/d' %{_sysconfdir}/apparmor.d/local/usr.bin.lessopen.sh
+fi
+
 %files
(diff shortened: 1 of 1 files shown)


adrianreber:
The CI error seems to be due to a missing `BuildRequires: gcc` and seems unrelated.

I am, however, not yet convinced of this change. Editing files in `%post` and `%postun` feels dangerous. But I don't have a better idea. Replacing `/etc/apparmor.d/local/usr.bin.lessopen.sh` also doesn't feel right, if the user has actually changed that file.

Currently, reinstalling the package I get the entry twice:
```shell
$ cat /etc/apparmor.d/local/usr.bin.lessopen.sh 
# Site-specific additions and overrides for 'usr.bin.lessopen.sh'
# allow lmod for OHPC-lmod
  /usr/bin/readlink mrix, #OHPC-lmod
  /usr/bin/expr mrix, #OHPC-lmod
  /usr/bin/basename mrix, #OHPC-lmod
# allow lmod for OHPC-lmod
  /usr/bin/readlink mrix, #OHPC-lmod
  /usr/bin/expr mrix, #OHPC-lmod
  /usr/bin/basename mrix, #OHPC-lmod
```

adrianreber:
Looks we do not need this with the newer version of Lmod.

mslacken:
Will the new lmod version make it 3.x?
If yes, we can close this one and adding a note to documentation that lmod and apparmor cause problems on 2.x.
I also totally agree with you that changing configuration files in %post is ugly and don't know why the apparmor rpm adds this local config files without any content.

adrianreber:
> Will the new lmod version make it 3.x?

Yes, before ISC 2024 we want to release 3.1 and 2.8. Both releases will contain the latest version of Lmod (as of today).

> If yes, we can close this one and adding a note to documentation that lmod and apparmor cause problems on 2.x.

As we will bring the latest Lmod also to 2.x it should not be necessary.

> I also totally agree with you that changing configuration files in %post is ugly and don't know why the apparmor rpm adds this local config files without any content.

:wink:

Issue #1981, opened 2024-05-15
Update Score-P / Scalasca to new versions


Thyre:
Hi there,

I'm one of the developers working on the performance measurement infrastructure Score-P, mainly on the OpenMP and accelerator support.

After updating Score-P and OPARI2 to newer versions in the Fedora repositories, @adrianreber made me aware of Score-P and Scalasca being in the OpenHPC repositories and also mentioned that these are a bit outdated.

---

After discussing with colleagues for a bit, we're interested in providing newer versions of Score-P and Scalasca and try to keep them updated as well. We're doing the same thing for EasyBuild and spack already and also maintain a Ubuntu PPA for this. 

Based on the current state of the SPEC files, I would suggest the following approach:

- First, split the vendor packages delivered in Score-P into its own packages. Score-P normally includes the latest and greatest versions when it releases, but sometimes new versions of our dependent libraries release before a new version of Score-P. This is especially important with the idea to also integrate Cube GUI into OpenHPC (see https://github.com/openhpc/submissions/issues/49 ).
- Next, update Score-P for the latest versions. We are able to test things for Intel compilers and GCC and have a extensive internal testing suite for it. For Intel compilers (if the LLVM-based compilers are used), oneAPI 2024.0 or newer will be required as older versions will not give any information about called functions (unfortunately). 
- Then, update Scalasca. This should be straight-forward.

---

I'll probably start working on this in the coming days / weeks, depending on how much other work there is.

adrianreber:
@Thyre Thanks a lot for your help. I agree with your approach. Let's do it.

PR #1983, opened 2024-05-19
scorep: Split components in preparation for Score-P & Scalasca update

Thyre:
Hi there,

as described in #1981, we would like to provide a new version of Score-P for OpenHPC. 
In the issue, I've described my strategy for doing this, so that future maintainability is easier and Score-P components can both be updated independently and used by other projects without needing to load Score-P. This is mostly useful for OTF2 and OPARI2, where we are aware of several other tools using these libraries for their performance tools.

---

This PR is the first step towards updating Score-P and Scalasca to new versions. While the Score-P SPEC file is untouched here (this will be a separate PR, together with Scalasca), this PR provides the individual components. 

Notable things:
- CubeW, CubeLib and OPARI2 do not need a specific MPI to be built. For OTF2, this is only necessary because of SIONlib being built with different MPIs as well.
- While the ARM compilers are also included in this PR, we are (currently) unable to test them internally. We do test Clang compilers very thoroughly though, so I hope that we do not run into issues here.
- For Intel compilers, we would either like to focus the Classic Intel compilers, or preferably oneAPI 2024.0 or newer. This will be important for Score-P as the LLVM-based compilers in oneAPI 2023.2.1 and lower did not support `-finstrument-functions` for Fortran, which is crucial to get the called functions in our profile / trace.

### DIFF
--- a/components/io-libs/otf2/SPECS/otf2.spec
+++ b/components/io-libs/otf2/SPECS/otf2.spec
@@ -1,0 +1,157 @@
+#----------------------------------------------------------------------------bh-
+# This RPM .spec file is part of the OpenHPC project.
+#
+# It may have been modified from the default version supplied by the underlying
+# release package (if available) in order to apply patches, perform customized
+# build/install configurations, and supply additional files to support
+# desired integration conventions.
+#
+#----------------------------------------------------------------------------eh-
+
+# Build that is dependent on compiler toolchains
+%define ohpc_compiler_dependent 1
+%define ohpc_mpi_dependent 1
+%include %{_sourcedir}/OHPC_macros
+
+# Base package name
+%define pname otf2
+
+Summary:        Open Trace Format 2 library
+Name:           %{pname}-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
+Version:        3.0.3
+Release:        1%{?dist}
+License:        BSD-3-Clause
+Group:          %{PROJ_NAME}/io-libs
+URL:            http://score-p.org
+Source0:        http://perftools.pages.jsc.fz-juelich.de/cicd/otf2/tags/%{pname}-%{version}/%{pname}-%{version}.tar.gz
+BuildRequires:  gcc-c++
+BuildRequires:  make
+BuildRequires:  which
+BuildRequires:  chrpath dos2unix
+# Need a new py-compile for Python 3.12
+BuildRequires:  libtool automake
+BuildRequires:  sionlib-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
+Requires:       lmod%{PROJ_DELIM} >= 7.6.1
+
+# Default library install path
+%define install_path %{OHPC_LIBS}/%{compiler_family}/%{mpi_family}/%{pname}/%version
+
+%description
+The Open Trace Format 2 (OTF2) is a highly scalable, memory efficient
+event trace data format plus support library.
+
+This is the %{compiler_family}-%{mpi_family} version.
+
+%prep
+
+%setup -q -n %{pname}-%{version}
+dos2unix doc/examples/otf2_high_level_writer_example.py
+rm build-config/py-compile
+for d in . build-backend build-frontend	
+do
+  cd $d
+  autoreconf -f -i -v
+  cd -
+done
+
+%build
+
+# OpenHPC compiler/mpi designation
+%ohpc_setup_compiler
+
+module load sionlib
+
+%if "%{compiler_family}" == "intel"
+CONFIGURE_OPTIONS="--with-nocross-compiler-suite=oneapi "
+%endif
+
+%if "%{compiler_family}" == "arm1"
+export CFLAGS="${CFLAGS} -fsimdmath"
+export CXXFLAGS="${CXXFLAGS} -fsimdmath"
+%endif
+
+./configure --prefix=%{install_path} \
+            --disable-static \
+            --enable-shared \
+            --disable-silent-rules \
+            --enable-backend-test-runs \
+            --with-platform=linux \
+            CC="$CC" \
+            CXX="$CXX" \
+            CFLAGS="$CFLAGS" \
+            CXXFLAGS="$CXXFLAGS" \
+            LDFLAGS="$LDFLAGS" \
+            ${CONFIGURE_OPTIONS}
+
+make %{?_smp_mflags} V=1
+
+%check
+
+# OpenHPC compiler/mpi designation
+%ohpc_setup_compiler
+
+module load sionlib
+
+make check
+
+%install
+
+# OpenHPC compiler/mpi designation
+%ohpc_setup_compiler
+
+module load sionlib
+
+make DESTDIR=$RPM_BUILD_ROOT install
+
+# don't package static libs
+rm -rf $RPM_BUILD_ROOT%{install_path}/lib/*.a \
+       $RPM_BUILD_ROOT%{install_path}/lib/*.la
+
+# OpenHPC module file
+%{__mkdir} -p %{buildroot}%{OHPC_MODULEDEPS}/%{compiler_family}-%{mpi_family}/%{pname}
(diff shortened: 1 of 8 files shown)


adrianreber:
The openEuler error seems to be related to broken repository errors. Independent of your change.

The OpenHPC project tries to have a lot of tests which also verify that the components actually work. Are there any upstream tests we could take from each of the four upstream projects to include in our test tree? If you want to look into that even better, but the test suite is a bit complicated to figure out. What I usually do locally is a container and then I run commands like this:
```shell
$ podman run -it -v /<checkout>/ohpc:/<checkout>/ohpc docker.io/library/almalinux:9 bash
$ tests/ci/prepare-ci-environment.sh
$ tests/ci/run_build.py ohpc components/perf-tools/omb/SPECS/omb.spec
$ tests/ci/setup_slurm_and_run_tests.sh ohpc gnu13 components/perf-tools/omb/SPECS/omb.spec
```

This are basically the steps which are run in CI. If you can point me to some upstream tests I can also add the tests once your changes are merged.

I would expect the openEuler errors to disappear in a couple of days.

Thyre:
> The openEuler error seems to be related to broken repository errors. Independent of your change.

I've noticed that in my fork as well with openEuler failing the CI each time while trying to install git. 

> The OpenHPC project tries to have a lot of tests which also verify that the components actually work. Are there any upstream tests we could take from each of the four upstream projects to include in our test tree? If you want to look into that even better, but the test suite is a bit complicated to figure out. What I usually do locally is a container and then I run commands like this:
> 
> ```shell
> $ podman run -it -v /<checkout>/ohpc:/<checkout>/ohpc docker.io/library/almalinux:9 bash
> $ tests/ci/prepare-ci-environment.sh
> $ tests/ci/run_build.py ohpc components/perf-tools/omb/SPECS/omb.spec
> $ tests/ci/setup_slurm_and_run_tests.sh ohpc gnu13 components/perf-tools/omb/SPECS/omb.spec
> ```
> 
> This are basically the steps which are run in CI. If you can point me to some upstream tests I can also add the tests once your changes are merged.

Right now, all four components will run `make check` during the build process to ensure that our "normal" tests will succeed. We also have additional tests which can be run with `make installcheck` after `make install` to ensure that everything (we test) works as expected even after installation. Those are currently not included, as I did not know where to put them. In the case of CubeLib, CubeW and OPARI2, the installcheck tests should try to build / run the examples of the documentation. The OTF2 installcheck seems to be a bit larger. I would need to ask around what is checked there specifically.

I also would need to check if its possible to extract these tests from the source tree to include them here. It might be difficult.

For Score-P, having both checks would actually be very helpful, since `make check` and `make installcheck` essentially do different things.

adrianreber:
Good to know, thanks. Let's look into separate tests later. OpenEuler repositories seem to be fixed, if you trigger a rerun of the CI it should pass now.

Thyre:
> Good to know, thanks. Let's look into separate tests later. OpenEuler repositories seem to be fixed, if you trigger a rerun of the CI it should pass now.

Re-triggering the failed job seemed to work fine. All checks are green now.

components/serial-libs/cubelib/SPECS/cubelib.spec
```diff
@@ -0,0 +1,142 @@
...
+# Build that is dependent on compiler toolchains
+%define ohpc_compiler_dependent 1
+%include %{_sourcedir}/OHPC_macros
+
+# Base package name
+%define pname cubelib
+%define shortwv 4.8
+
+Summary:        CUBE Uniform Behavioral Encoding generic presentation library component
+Name:           %{pname}%{PROJ_DELIM}
```
Thyre: Should be `%{pname}-%{compiler_family}%{PROJ_DELIM}`

components/serial-libs/opari2/SPECS/opari2.spec
```diff
@@ -0,0 +1,137 @@
...
+
+# Build that is dependent on compiler toolchains
+%define ohpc_compiler_dependent 1
+%include %{_sourcedir}/OHPC_macros
+
+# Base package name
+%define pname opari2
+
+Summary:        An OpenMP runtime performance measurement instrumenter
+Name:           %{pname}%{PROJ_DELIM}
```
Thyre: Should be `%{pname}-%{compiler_family}%{PROJ_DELIM}`

components/io-libs/otf2/SPECS/otf2.spec
```diff
@@ -0,0 +1,153 @@
...
+module-whatis "Category: runtime library"
+module-whatis "Description: %{summary}"
+module-whatis "URL %{url}"
+
+set     version			    %{version}
+
+depends-on sionlib
+
+prepend-path    PATH                %{install_path}/bin
+prepend-path    MANPATH             %{install_path}/share/man
```
Thyre, line 134: We need to add  `LD_LIBRARY_PATH` / `INCLUDE` as well.

adrianreber:
Is this ready from your side? Tests are all good so far.

Thyre:
Yes, this PR should be good.

adrianreber:
@Thyre The results are here: https://obs.openhpc.community/project/show/OpenHPC3:3.2:Factory

All Leap builds are failing because of rpmlint errors. Can you update the rpmlint files to correctly ignore a couple of the checks.

Thyre:
> @Thyre The results are here: https://obs.openhpc.community/project/show/OpenHPC3:3.2:Factory
> 
> All Leap builds are failing because of rpmlint errors. Can you update the rpmlint files to correctly ignore a couple of the checks.

Sure, I'll look into it

adrianreber:
The opari2 build fails with [   64s] + /usr/lib/rpm/brp-python-hardlink
```
[   64s] + /usr/lib/rpm/redhat/brp-mangle-shebangs
[   64s] *** WARNING: ./opt/ohpc/pub/libs/gnu13/opari2/2.0.8/share/doc/opari2/example/lib/libpomp.la is executable but has no shebang, removing executable bit
[   64s] *** ERROR: ./opt/ohpc/pub/libs/gnu13/opari2/2.0.8/libexec/pomp2-parse-init-regions.awk has shebang which doesn't start with '/' (-f)
[   64s] error: Bad exit status from /var/tmp/rpm-tmp.aWrFUM (%install)
```
on RHEL

Thyre:
> The opari2 build fails with [ 64s] + /usr/lib/rpm/brp-python-hardlink
> 
> ```
> [   64s] + /usr/lib/rpm/redhat/brp-mangle-shebangs
> [   64s] *** WARNING: ./opt/ohpc/pub/libs/gnu13/opari2/2.0.8/share/doc/opari2/example/lib/libpomp.la is executable but has no shebang, removing executable bit
> [   64s] *** ERROR: ./opt/ohpc/pub/libs/gnu13/opari2/2.0.8/libexec/pomp2-parse-init-regions.awk has shebang which doesn't start with '/' (-f)
> [   64s] error: Bad exit status from /var/tmp/rpm-tmp.aWrFUM (%install)
> ```
> 
> on RHEL

That's interesting. I'll look into it, but haven't seen this issue yet. On the Alma container I'm working in right now, the shebang is correct:

```console
[root@072fd91fc1b3 libexec]# head -n 1 pomp2-parse-init-regions.awk 
#!/usr/bin/gawk -f
```

It looks like `awk` is missing entirely. There probably needs to be another dependency which is missing right now. Will fix it.

adrianreber:
The build system (OBS) builds things a bit differently. Maybe something is not installed or differently.

I can actually look at the build results and see following shebang:
```shell
$ head -1 pomp2-parse-init-regions.awk
#! -f
```

Ah, you need to buildrequire `which`. That is missing the chroot.

adrianreber:
Once the builds are done in OBS (https://obs.openhpc.community/project/show/OpenHPC3:3.2:Factory) you can change https://github.com/openhpc/ohpc/blob/3.x/tests/ci/prepare-ci-environment.sh#L8 to 3.2 and try to retrigger CI again. I hope it should then use the new packages available in OBS.

PR #1989, opened 2024-06-18
easybuild: update to 4.9.2

boegel:


adrianreber:
It seems CI fails because downloading the tarballs no longer works. The delimiter used to be `-` and now it changed to `_`.

You need to change the delimiter for the additional tarballs. Not sure if it is needed for all of them.

adrianreber:
I think it now fails because the directories are using `_` and in the spec file there are still `-`.

boegel:
superseded by #2013

PR #1993, opened 2024-07-04
Update openEuler to 22.03-LTS-SP4

martin-g:


adrianreber:
Does this mean you want to move to LTS-SP4 for the next 3.x release? Does this also require changes on the builder side?

martin-g:
Actually it is not really needed at the moment!

PR #1994, opened 2024-07-04
Remove duplicate ReturnToService line in /etc/slurm/slurm.conf

### LINKED ISSUE #1991: Duplicate ReturnToService line in /etc/slurm/slurm.conf?
Currently have slurm-example-configs-ohpc-23.11.6-310.ohpc.5.1.x86_64 installed on Rocky 9.

https://github.com/openhpc/ohpc/blob/792e7cd3ba1aa486f2f8e2d4cd8494c98ca161c9/components/rms/slurm/SOURCES/slurm.conf.example.patch#L45

There are notes on the previous line indicating there was a `ReturnToService` variable removed earlier, but that doesn't appear to be the case.

```bash
[rocky@sms ~]$ grep ReturnToService /etc/slurm/slurm.conf.example
ReturnToService=1
# Adding ReturnToService Directive to starting config file (note removal of variable during above creation)
ReturnToService=1
```

Did I miss something, or should this line be removed from the patch, or the earlier ReturnToService line be removed via whatever method was proposed?

mikerenfro:
Fixes #1991

### DIFF
--- a/components/rms/slurm/SOURCES/slurm.conf.example.patch
+++ b/components/rms/slurm/SOURCES/slurm.conf.example.patch
@@ -41,5 +41,3 @@
 +SlurmctldParameters=enable_configless
 +# Setup interactive jobs for salloc
 +LaunchParameters=use_interactive_step
-+# Adding ReturnToService Directive to starting config file (note removal of variable during above creation)
-+ReturnToService=1


adrianreber:
Unfortunately our CI is not clever enough to detect that your change means that a rebuild of slurm is required to test the PR. Test builds in CI are only triggered if the actual spec file is changed. The changed patch does not trigger a rebuild.

Trying it out locally your PR fails building. Following change on top helps:

```diff
-@@ -149,5 +149,22 @@
+@@ -149,5 +149,20 @@
```

Also, we require patches to be signed-off-by as described in https://github.com/openhpc/ohpc/blob/3.x/CONTRIBUTING.md

Fixing the patch with what I posted and adding the correct sign off should make this ready to be merged.

adrianreber:
Locally adapted and applied.

PR #2006, opened 2024-09-12
compiler-families: update GCC to 14.2

adrianreber:


components/OHPC_macros
```diff
@@ -79,13 +79,18 @@ BuildRequires: ohpc-buildroot
...
-%{!?compiler_family: %global compiler_family gnu13}
+%{!?compiler_family: %global compiler_family gnu14}
 %{!?mpi_family: %global mpi_family openmpi5}
 %{!?python_family: %global python_family python3}
 
 # Compiler dependencies
 %if 0%{?ohpc_compiler_dependent} == 1
 
+%if "%{compiler_family}" == "gnu14"
+BuildRequires: gnu14-compilers%{PROJ_DELIM} >= 13.1.0
```
martin-g:
```suggestion
BuildRequires: gnu14-compilers%{PROJ_DELIM} >= 14.2.0
```

components/OHPC_macros
```diff
@@ -79,13 +79,18 @@ BuildRequires: ohpc-buildroot
...
+Requires:      gnu14-compilers%{PROJ_DELIM} >= 13.1.0
```
martin-g:
```suggestion
Requires:      gnu14-compilers%{PROJ_DELIM} >= 14.2.0
```

components/OHPC_macros
adrianreber: Thanks. Too much copy and paste.

PR #2015, opened 2024-09-16
metis: switch to GitHub location

adrianreber:


adrianreber:
This new version segfaults during tests. Not merging.

PR #2016, opened 2024-09-17
scotch: handle gcc 14

adrianreber:


adrianreber:
The CI cannot handle renamed files correctly. That is why this fails. Still merging.

PR #2017, opened 2024-09-18
Score-P & Scalasca: Add minor fixes

Thyre:
Add two minor fixes to Score-P & Scalasca, which were partially already merged in other build systems.

**Score-P**:

GCC 14 has changed some things on GMP, which caused our GCC plug-in check to fail. We fixed this in the Score-P 9.0 development cycle by updating our "fake" header. This commit backports these changes to Score-P 8.4, so that the GCC plug-in continues to work. This patch is already merged in Fedora for example (see [here](https://src.fedoraproject.org/rpms/scorep/c/23c83219cbb94a65f67933c97832c3178500225d?branch=rawhide)).

**Scalasca**:

Add a small patch which fixes a typo in configure. This can cause tests to during configure to fail, as all tests after the OpenMP detection will be checked for any output on `stdout` and `stderr`. A PR for this fix exists for [Fedora](https://src.fedoraproject.org/rpms/scalasca/pull-request/3) & [EasyBuild](https://github.com/easybuilders/easybuild-easyconfigs/pull/21135), with our internal JSC EasyBuild having this patch for quite some while already.

### DIFF
--- a/components/perf-tools/scalasca/SOURCES/scalasca-2.6.1-nowarn-omp-pragmas.patch
+++ b/components/perf-tools/scalasca/SOURCES/scalasca-2.6.1-nowarn-omp-pragmas.patch
@@ -1,0 +1,39 @@
+This patch fixes a missing $ in configure, which causes following checks to
+fail if they produce any output to stdout or stderr.
+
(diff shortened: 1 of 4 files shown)


adrianreber:
Thanks. You don't need to change the release from 1 to 2. This is done automatically by the build system. Please undo those changes.

PR #2029, opened 2024-09-26
Update Superlu to v7.0.0

martin-g:
Remove patches that are applied upstream. Update the others.


martin-g:
@adrianreber Please review!

components/serial-libs/superlu/SPECS/superlu.spec
```diff
@@ -54,34 +50,34 @@ linear systems that the author is aware of.
...
 
 %install
 mkdir -p %{buildroot}%{install_path}/lib
 mkdir -p %{buildroot}%{install_path}/include
 install -m644 SRC/*.h %{buildroot}%{install_path}/include
 install -m755 lib/libsuperlu.so %{buildroot}%{install_path}/lib/libsuperlu.so.%{version}
 pushd %{buildroot}%{install_path}/lib
-ln -s libsuperlu.so.%{version} libsuperlu.so.4
-ln -s libsuperlu.so.4 libsuperlu.so
+ln -s libsuperlu.so.%{version} libsuperlu.so.7
```
adrianreber, line 79: Why did you use 7? Is that used by upstream or are just following the package version?

martin-g: Following the package version. 
Should I revert to v4?

adrianreber: No idea. In theory we should figure out if the ABI changed if we change the SO version, but as we would distribute this update with a new compiler it means recompilation anyway for anybody who wants to use it.

Just leave it at seven.

adrianreber:
I am not totally happy with the commits. Maybe just squash it together as it is all part of the package update.

martin-g:
I forgot about this!
I will do it later on my dev machine!

How about enabling GitHub UI to do it for us ? It is handy for simple squashes and rewording.

adrianreber:
> I forgot about this! I will do it later on my dev machine!
> 
> How about enabling GitHub UI to do it for us ? It is handy for simple squashes and rewording.

It was already enabled. Never tried it before. Just tried it now and it worked. Thanks.

PR #2034, opened 2024-10-01
lmod: Update to 8.7.49

martin-g:


martin-g:
https://github.com/openhpc/ohpc/actions/runs/11138923266/job/30954689982?pr=2035
Is Lmod broken in the new version ?!

martin-g:
I think this is the problem we face too - https://github.com/TACC/Lmod/issues/718#issuecomment-2343334206

adrianreber:
There have been a couple of lmod releases since your revert. Do you think we should try upgrade lmod once more? The git history at lmod is not clear if this is now fixed or not.

martin-g:
https://github.com/TACC/Lmod/issues/718 is still opened, but there are several commits to https://github.com/TACC/Lmod/commits/main/init/bash.in
Let's try !

PR #2035, opened 2024-10-01
 powerman: Remove the spec and the documentation for powerman

martin-g:


adrianreber:
I don't think we ship that for a long time. I do not see it mentioned in the OBS configuration for 2.x and 3.x. Also for 1.3.x and even older it was never part of the documentation.

The right thing would be to just remove it. Do you want to remove the SPEC file and from the documentation?

martin-g:
I'll remove it!

martin-g:
It seems to be used by `conman` - https://github.com/openhpc/ohpc/blob/184227cefb6dbcd64efe57d17da107847031a690/components/admin/conman/SPECS/conman.spec#L84
I'd expect a `Requires:` for it.

adrianreber:
> It seems to be used by `conman` -
> 
> https://github.com/openhpc/ohpc/blob/184227cefb6dbcd64efe57d17da107847031a690/components/admin/conman/SPECS/conman.spec#L84
> 
> I'd expect a `Requires:` for it.

It is strange, indeed. But powerman was never built. I also saw that line and was confused.

adrianreber:
We probably should remove the line. It never worked anyway.

adrianreber:
The `skip` command in those bats tests does not seem to work. Strange. Seems unrelated to this change, however.

Maybe the skip command does not work outside of `@test`.

adrianreber:
Almost ready to be merged. Could you remove your debug statements from the first commit? Currently you have it in the second commit which means there are some unnecessary code changes in this PR which I would prefer to not have in the repository.

martin-g:
> Could you remove your debug statements from the first commit?

I didn't know how to do it until now! Thanks for pushing me over my limits ! :-)

adrianreber:
> > Could you remove your debug statements from the first commit?
> 
> I didn't know how to do it until now! Thanks for pushing me over my limits ! :-)

Glad I could help! :wink:

PR #2036, opened 2024-10-01
tests: Fix the spec URL value

martin-g:
https://github.com/openhpc/ohpc/tests leads to 404. 
The correct one would be https://github.com/openhpc/ohpc/tree/3.x/tests but it is version specific


adrianreber:
This probably needs a `get_source.sh` just as you did for `docs`.

adrianreber:
Thanks. The PR has gotten pretty big compared to that it was single line change. But thanks for your work.

PR #2038, opened 2024-10-10
Spack: Update to 0.22.2 (2024-09-21)

martin-g:
Remove an unused patch


adrianreber:
This is a bit more complicated and I already have prepared something locally. We want to make software installed via slack available for all users. Currently that does not work the way we install it.

martin-g:
OK! I'll leave it to you, @adrianreber !

Issue #1995, opened 2024-07-09
spack usage


gustavoberman:
I didn't found documentation on how it was intended to use Spack inside OHPC.

- The spack-ohpc pakage leaves Spack at the `/opt/ohpc/admin` directory.
- It has a module that it's only accessed by root 
- All spack compiled packages and corresponding modules stay inside the `/opt/ohpc/admin/spack` dir.
- So cluster users have no easy way to access those packages and modules.
- And default NFS export don't even allows to access those packages from the compute nodes.
 
Didn't make much sense to me if users and nodes cannot access it.

I guess that the basic usage should be that only cluster admins have access to compile packages and create modules using that installation of spack, and cluster users are allowed to use those packages by loading the modules. This way cluster admins can decide which packages to support and they should be the ones with enough knowledge to understand what options to choose and how to compile packages for the cluster.

So in order to do that I had to make some changes.

- Allow users to access spack created modules:
By changing `/etc/profile.d/lmod.sh` both in frontend and nodes and adding
`/opt/ohpc/admin/spack/0.21.2/share/spack/modules/linux-rocky9-sandybridge/`
to the cluster users `MODULEPATH` variable

- Allow compute nodes to access compiled packages and modules:
By adding spack directory to `/etc/exports`:
`/opt/ohpc/admin/spack *(ro,no_subtree_check,fsid=12)`
And mounting that export in the nodes adding the following line to `/etc/fstab` of the nodes:
`frontend.ip.address:/opt/ohpc/admin/spack /opt/ohpc/admin/spack nfs nfsvers=3,nodev 0 0 `

Perhaps there is need to decide what could be the default way to use it in ohpc and add something on the documentation?

Thanks!

adrianreber:
Thanks for the detailed description. We will discuss this in the technical steering committee and see if we can make it easier to use spack.

adrianreber:
After discussing it in the TSC we will look into making spack available also to non-root users.

adrianreber:
Should be fixed in 3.2

PR #2042, opened 2024-10-13
spack: upgrade to 0.22.2

adrianreber:
This includes better integration into OpenHPC (again).

This also removes an old patch which was not applied for a long time. This patch was important to make spack installed software packages available to OpenHPC users.

In its default configuration in OpenHPC only root can install additional software packages using spack, but these packages will be made available to all users again.

Non-root users can also use spack, but it requires additional configuration for each user. Unfortunately spack does not automatically set different defaults when running as non-root like EasyBuild for example seems to do.


adrianreber:
@martin-g This tries to solve the problem described in #1995

adrianreber:
The test failure is due to a limitation of the test script and our defaults packages.

PR #2050, opened 2024-10-25
HDF5: Fix the build with Intel compiler

martin-g:
Uses a patch to revert https://github.com/HDFGroup/hdf5/commit/f6ae24103290215a92b981a205215e59a6ada5f3#diff-eff1bae8d512738be4c5337b594867ef4f22b5f8940b8fbdcc3e9a78e1419bb6R225

Also see https://github.com/HDFGroup/hdf5/issues/4637


adrianreber:
The package that fails in OBS is phdf5. Did we also see problems with hdf5? So I think we need this change in phdf5.

martin-g:
> The package that fails in OBS is phdf5

Oh.

> Did we also see problems with hdf5? So I think we need this change in phdf5.

https://obs.openhpc.community/package/show/OpenHPC3:3.2:Factory/hdf5-intel is green!
Let me open a new PR!

PR #2053, opened 2024-11-01
Fix default network and DNS for compute nodes on Warewulf 4

middelkoopt:
Add default networking to node profiles   
    * Add netmask, gateway, and dns at the node profile level
    * Add internal_network, ipv4_gateway, dns_servers across
      input.local.template and inputs.tex

### DIFF
--- a/docs/recipes/install/common/warewulf4_setup_centos.tex
+++ b/docs/recipes/install/common/warewulf4_setup_centos.tex
@@ -5,10 +5,6 @@
 [sms](*\#*) ip link set dev ${sms_eth_internal} up
 [sms](*\#*) ip address add ${sms_ip}/${internal_netmask} broadcast + dev ${sms_eth_internal}
 
-# Compute the network address for the internal network
-[sms](*\#*) internal_cidr=$(netmask ${sms_ip}/${internal_netmask})
-[sms](*\#*) internal_network=${internal_cidr%/*}
-
 # Edit the warewulf.conf file to use appropriate interface and settings
 [sms](*\#*) perl -pi -e "s/ipaddr:.*/ipaddr: ${sms_ip}/" /etc/warewulf/warewulf.conf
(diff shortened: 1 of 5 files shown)


middelkoopt:
Hold on, something's not right with this PR/commit.

middelkoopt:
PR Fixed, something went wonky while rebasing against the latest 3.x.

docs/recipes/install/rocky9/input.local.template
```diff
@@ -21,6 +21,9 @@ sms_eth_internal="${sms_eth_internal:-eth1}"
 # Subnet netmask for internal cluster network
 internal_netmask="${internal_netmask:-255.255.0.0}"
 
+# Subnet network for internal cluster network
+internal_netmask="${internal_network:-172.16.0.0}"
```
adrianreber: Copy paste typo here.

docs/recipes/install/common/warewulf4_setup_centos.tex
```diff
@@ -5,10 +5,6 @@
 [sms](*\#*) ip link set dev ${sms_eth_internal} up
 [sms](*\#*) ip address add ${sms_ip}/${internal_netmask} broadcast + dev ${sms_eth_internal}
 
-# Compute the network address for the internal network
-[sms](*\#*) internal_cidr=$(netmask ${sms_ip}/${internal_netmask})
```
adrianreber, line 9: This means you can also remove the netmask package installation in the recipe.

middelkoopt:
Good eye. Fixes applied.  I forced pushed this time since there were small changes, let me know if I should stick to delta's and squash a t the end.

adrianreber:
Force pushing is almost always the right solution from my point of view. Except you want to split commits for some reason. I would not push fixup commits to a PR. But that is just me.

PR #2058, opened 2024-11-15
Update likwid.spec with version 5.4.0

TomTheBear:


adrianreber:
Thanks for the PR. The commit needs a `Signed-off-by` as described in https://github.com/openhpc/ohpc/blob/3.x/CONTRIBUTING.md

It also seems that the documentation moved, at least the RPM build process in CI fails with a missing file.

TomTheBear:
Ask Github to add signed-off when changing a file over the web... I changed exactly on character. I don't clone the repo for such a change.

The documentation has not moved but we worked on it (few files deleted, some added). I would say, the `%doc %{install_path}/share/likwid/docs/*` should catch all of that. That's also why I didn't change anything there, I expected it to work.

PR #2062, opened 2024-12-10
Update LIKWID to 5.4.1

TomTheBear:
Now with `--sign-off` and without `share/likwid/docs`, it never contained any files


adrianreber:
Thanks.

Issue #2063, opened 2024-12-10
apostrophe rendered instead of single quote in syslog section of recipes


graham-cs:
Minor note: It looks like the first four single quotes in the code block under _3.8.4.7 Enable forwarding of system logs_ in [Install_guide-Rocky9-Warewulf4-SLURM-3.2-x86_64.pdf](https://github.com/openhpc/ohpc/releases/download/v3.2.GA/Install_guide-Rocky9-Warewulf4-SLURM-3.2-x86_64.pdf) are rendered as apostrophes. The rest of the single quotes in the document appear as expected. Also saw this under _3.6.4.1_ of [Install_guide-Leap15-Warewulf-OpenPBS-3.2-aarch64.pdf](https://github.com/openhpc/ohpc/releases/download/v3.2.GA/Install_guide-Leap15-Warewulf-OpenPBS-3.2-aarch64.pdf).

Not sure if this is an issue on my end - i tried multiple pdf viewers. The block appears to be derived from `ohpc/docs/recipes/install/common/syslog.tex` (which looks correct)?

https://github.com/openhpc/ohpc/blob/7b406be1cc22ffe6a6c886cb187f4e908eb982e4/docs/recipes/install/common/syslog.tex#L12-L13

adrianreber:
Thanks for the report. Looking at other files with single quotes it seems we need `,upquote=true` in `\begin{lstlisting}`.

Would you be interested in providing a pull request to fix this?

graham-cs:
sure, thanks.

adrianreber:
Fixed with #2065

PR #2065, opened 2024-12-10
added upquote=true to syslog.tex

graham-cs:
added upquote=true to fix single quote formatting per https://github.com/openhpc/ohpc/issues/2063


adrianreber:
Thanks

PR #2068, opened 2024-12-24
Include f2py back into OHPC Numpy

sunghjung3:
This follows the discussion from the [OHPC mailing list](https://lists.openhpc.community/g/users/topic/no_f2py_with/110266952).

Commit 1ab2e9c9ec42a2b170a1a125e5359606090be9f3 in 2019 added a line in the .spec file to explicitly exclude f2py installation. This commit simply removes this line.

### DIFF
--- a/components/dev-tools/numpy/SPECS/python-numpy.spec
+++ b/components/dev-tools/numpy/SPECS/python-numpy.spec
@@ -147,7 +147,6 @@ EOF
 %{__mkdir_p} ${RPM_BUILD_ROOT}/%{_docdir}
 
 %files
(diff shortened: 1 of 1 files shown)


adrianreber:
@crbaird do yo remember why you removed `f2py` many years ago. According to the git history it was you, but no detailed description why. Adding it back doesn't seem like a problem to me.

adrianreber:
Thanks. Merging it. There was a comment that it might have been related to some compilers not entirely happy with the `f2py` script. Since the compilers have changed since it was removed many years ago, let's try to re-introduce it.

PR #2071, opened 2025-02-04
create apparmor abstractions for openSUSE

### LINKED ISSUE #2069: Lmod issues with openSuse Leap
I have a somewhat similar issue to #1974, where `less` used on any file prints this message:
```
/opt/ohpc/admin/lmod/lmod/init/bash: /opt/ohpc/admin/lmod/lmod/libexec/addto : /usr/bin/lua5.3 : bad interpreter: Permission denied
```

I have tracked down the problem to the line:
```bash
export FPATH=$(/opt/ohpc/admin/lmod/lmod/libexec/addto --append FPATH ${__zsh_fpath:-$FPATH} /opt/ohpc/admin/lmod/lmod/init/ksh_funcs)
```
in `/opt/ohpc/admin/lmod/8.7.53/init/bash` (when commenting this line the message doesn't appear anymore). 

Running this line in bash directly works normally.

I confirmed that this is an AppArmor problem by looking at the logs in `/var/log/audit/audit.log`:
```
type=AVC msg=audit(1738240871.101:5569): apparmor="DENIED" operation="exec" class="file" profile="/usr/bin/lessopen.sh" name="/opt/ohpc/admin/lmod/8.7.53/libexec/addto" pid=14071 comm="lessopen.sh" requested_mask="x" denied_mask="x" fsuid=0 ouid=0
```

I have tried to add `/usr/bin/lua5.3` to `/etc/apparmor.d/usr.bin.lessopen.sh`, to try the solution proposed in #1974, but that didn't solve the problem.

I have reproduced the issue on fresh installs of both openSuse Leap 15.5 and 15.6, after setting up OpenHPC repository and installing only `ohpc-base`.

I noticed that package `lua-lmod` provided in the openSuse repository doesn't have this problem, and seems to enclose the `export FPATH...` part with:
```bash
SUPPORT_KSH="no"
if [ $SUPPORT_KSH = yes -o -n "${KSH_VERSION+x}" ]; then
   if [ -z "${__LMOD_SET_FPATH+x}" ]; then
      export FPATH=$(/usr/share/lmod/lmod/libexec/addto --append FPATH  /usr/share/lmod/lmod/init/ksh_funcs)
      export __LMOD_SET_FPATH=1
   fi
fi
unset SUPPORT_KSH
```

mslacken:
Creates an apparmor profile for lmod.
Can be a separate package in opposite as I wrote in #2069
close #2069

### DIFF
--- a/components/admin/lmod/SPECS/lmod.spec
+++ b/components/admin/lmod/SPECS/lmod.spec
@@ -45,6 +45,7 @@ BuildRequires: lua53-devel
 BuildRequires: procps
 Requires: lua53-luafilesystem
 Requires: lua53-luaposix
+Requires: (lua-lmod-apparmor-abstractions%{PROJ_DELIM} if apparmor-abstractions)
 Conflicts: Modules
 %endif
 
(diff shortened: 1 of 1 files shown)


adrianreber:
Thanks. Tests are happy. Let's merge it.

PR #2076, opened 2025-02-25
Add Development Container for Visual Studio Code

middelkoopt:
Add support for using a Development Container.  Uses a Rocky 9 image with minimal packages.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,18 @@
+{
+  "name": "OpenHPC Development",
+  "image": "registry.docker.com/library/rockylinux:9",
+  "remoteUser": "vscode",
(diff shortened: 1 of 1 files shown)


adrianreber:
Good idea. We already have a script to setup our CI environment in GitHub and I also use it locally in containers:

https://github.com/openhpc/ohpc/blob/3.x/tests/ci/prepare-ci-environment.sh

Would that also work for your use case? Instead of having a couple of commands just call that script in your `onCreateCommand`.

middelkoopt:
Perfect, it works well. Updated the PR.

adrianreber:
Thanks for the PR. I never used Development Containers so I cannot really tell how and if it works, but it sounds like a good addition making contributing and local testing easier.

PR #2082, opened 2025-03-18
Add AlmaLinux

middelkoopt:
This PR adds AlmaLinux by refactoring RockyLinux to support both via substitutions.
Of note is a patch to `parse_doc.pl` to substitute within a `for` loop (separate commit).
The recipe has only been lightly tested.


adrianreber:
Merging it now. Will let you know if I see problems and then we can fix it. Thanks!

adrianreber:
@MiddelkoopT Just after merging it I saw you are manually adding users for munge to the host and the compute nodes. The munge RPM already creates a user and by syncing users between hosts that should not be necessary, right?

middelkoopt:
Adding the user directly is a workaround for syncuser (probably not the final solution) to claim a UID that does not get allocated at a future point in order to get the tests working.   There is a collision in Alma between the compute node and host, this ensures that munge will get proper ID on the compute node to avoid issues when syncuser is run.  I have some thoughts on this and will bring it up at the TSC meeting.

PR #2086, opened 2025-03-20
Reformat modification of /etc/hosts.ww

middelkoopt:
Reformat (word-wrap) modification of /etc/hosts.ww in the host overlay that modifies /etc/hosts on the SMS.

I don't believe this is strictly needed except in the CI/test clusters, perhaps it could be only run during CI?


adrianreber:
> Reformat (word-wrap) modification of /etc/hosts.ww in the host overlay that modifies /etc/hosts on the SMS.

Thanks.

> I don't believe this is strictly needed except in the CI/test clusters, perhaps it could be only run during CI?

Yes and no. We ship tools which users can install to manage the cluster which depend on the existing layout of `/etc/hosts`. We could also adapt the tool (`koomie_cf`).

adrianreber:
Please add a `Signed-Off-By` to your commit (`git commit -s --amend`).

middelkoopt:
Just did - sorry about that.

PR #2090, opened 2025-03-21
Fix the use of netname and set netdev for Warewulf4

middelkoopt:
* Fix the incorrect use of `--netname` when setting the network in a profile.  
* Specify the network device (`--netdev`) for a compute node  in the nodes profile.
* Fix setting the network information for IB interfaces

### DIFF
--- a/docs/recipes/install/common/add_ww4_hosts_finalize.tex
+++ b/docs/recipes/install/common/add_ww4_hosts_finalize.tex
@@ -6,7 +6,7 @@
 \begin{lstlisting}[language=bash,keywords={},upquote=true,basicstyle=\footnotesize\ttfamily]
 # Optionally define IPoIB network settings (required if planning to mount Lustre/BeeGFS over IB)
 [sms](*\#*) for ((i=0; i<$num_computes; i++)) ; do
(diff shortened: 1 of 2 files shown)


docs/recipes/install/common/add_ww4_hosts_finalize.tex
```diff
@@ -6,7 +6,7 @@
 \begin{lstlisting}[language=bash,keywords={},upquote=true,basicstyle=\footnotesize\ttfamily]
 # Optionally define IPoIB network settings (required if planning to mount Lustre/BeeGFS over IB)
 [sms](*\#*) for ((i=0; i<$num_computes; i++)) ; do
-  wwctl node set --yes ${c_name[$i]} --netdev=ib0 --ipaddr=${c_ipoib[$i]} --netmask=${ipoib_netmask}
+  wwctl node set --yes ${c_name[$i]} --netname ib --netdev=ib0 --ipaddr=${c_ipoib[$i]} --netmask=${ipoib_netmask}
```
adrianreber: This is the only place where you do not use a `=`. I don't mind, but maybe for consistency add a `=`.

middelkoopt:
Thanks for the catch, I'm a strong believer in consistency. Pushed.

PR #2092, opened 2025-03-24
confluent alma

adrianreber:


adrianreber:
Verified on the test cluster that it still works with rocky and now also with almalinux.

Issue #2091, opened 2025-03-22
Nodes will sometimes not mount `/opt` or `/home`


middelkoopt:
Sometimes when a node boot's on Rocky it will fail to mount nfs volumes
```
Mar 19 17:05:22 c1 mount[813]: mount.nfs: Network is unreachable
Mar 19 17:05:22 c1 systemd[1]: opt.mount: Mount process exited, code=exited, status=32/n/a
Mar 19 17:05:22 c1 systemd[1]: opt.mount: Failed with result 'exit-code'.
Mar 19 17:05:22 c1 systemd[1]: Failed to mount /opt.
Mar 19 17:05:22 c1 mount[806]: mount.nfs: Network is unreachable
Mar 19 17:05:22 c1 systemd[1]: home.mount: Mount process exited, code=exited, status=32/n/a
Mar 19 17:05:22 c1 systemd[1]: home.mount: Failed with result 'exit-code'.
Mar 19 17:05:22 c1 systemd[1]: Failed to mount /home.
```

Originally reported by @adrianreber on the OpenHPC ww4 slack channel.

middelkoopt:
Bootlog of a node running AlmaLinux (fails every time) and is probably related.

[alma-journalctl2.txt](https://github.com/user-attachments/files/19402825/alma-journalctl2.txt)

adrianreber:
Thanks for tracking it here. It sounds a bit like a warewulf problem and not something OpenHPC might be able to fix.

snehring:
You may be able to drop in some site specific config to work around this. In my overlays for example I have a:
`/etc/systemd/system/check-ib-working.service` which contains:
```
[Unit]
DefaultDependencies=no
Before=network-online.target

[Service]
Type=oneshot
RemainAfterExit=yes
ExecStart=sh -c 'until ping -c 1 172.16.7.253; do sleep 1; done'
TimeoutStartSec=180s

[Install]
WantedBy=network-online.target
```
That ip is the ipoib ip of our warewulfd host, but you can use any ip address on the ib network that should always be up. There is probably a more generic way to accomplish this well, but ping is straightforward.

Then there's a symlink in `/etc/systemd/system/network-online.target.wants` called 'check-ib-working.service' that points to `/etc/systemd/system/check-ib-working.service` (this enables the service for that target)

This should make the `network-online.target` depend on `check-ib-working.service` before it gives the okay for other units (like the `remote-fs.target`) to go ahead.

[...]

PR #2099, opened 2025-03-30
Configure NetworkManager-wait-online to wait for network to be up

middelkoopt:
Addresses race condition during bootup between the network and mounting (#2091).  NetworManager-wait-online.service is set to use `-s`, which only waits for NetworkManager to statup, not the network.  This PR adds the step to remove the `-s` switch from the service unit.  Tested against AlmaLinux.


adrianreber:
Looks good. I will wait for CI results before merging.

PR #2103, opened 2025-04-07
Update OTF2, OPARI2, CubeLib, CubeW

Thyre:
Package updates:
- OPARI2 2.0.8 -> 2.0.9
- OTF2 3.0.3 -> 3.1.1
- CubeLib 4.8.2 -> 4.9
- CubeW 4.8.2 -> 4.9

Tested on SUSE Leap 15 with Intel & GNU so far.


Thyre:
I'll split GOTCHA into its own PR. This should help with reviewing the changes here, and may help with the CI to get a bit further.

adrianreber:
Unfortunately our CI cannot handle ordering. You probably need to split this PR further. Everything which is needed by something else during build time needs to be a separate PR.

Thyre:
> Unfortunately our CI cannot handle ordering. You probably need to split this PR further. Everything which is needed by something else during build time needs to be a separate PR.

I'll split Score-P into its own PR. The remaining packages should work fine, as Scalasca can still build with older versions of CubeLib, CubeW and OTF2.

adrianreber:
Thanks

adrianreber:
Looking into the test failures right now. There was a change around MANPATH.

adrianreber:
On OpenSUSE there is a strange error that the zypper tries to pull in the intel compiler. Not sure what is going on there.

adrianreber:
Ah, I think the problem is that scalasca requires scorep. You also need to split out scalasca.

Thyre:
> On OpenSUSE there is a strange error that the zypper tries to pull in the intel compiler. Not sure what is going on there.

Yeah:
```
2025-04-28T14:50:00.1429519Z Problem: 1: nothing provides 'intel-oneapi-compiler-fortran-2023.1.0' needed by the to be installed intel-compilers-devel-ohpc-2023.1-300.ohpc.6.1.x86_64
2025-04-28T14:50:00.1430558Z  Solution 1: Following actions will be done:
2025-04-28T14:50:00.1431172Z   do not install scalasca-gnu14-openmpi5-ohpc-2.6.2-19999.ci.ohpc.x86_64
2025-04-28T14:50:00.1431897Z   do not install scalasca-gnu14-mvapich2-ohpc-2.6.2-19999.ci.ohpc.x86_64
2025-04-28T14:50:00.1432610Z   do not install scalasca-gnu14-mpich-ohpc-2.6.2-19999.ci.ohpc.x86_64
2025-04-28T14:50:00.1433325Z  Solution 2: do not install cubelib-gnu14-ohpc-4.9-19999.ci.ohpc.x86_64
2025-04-28T14:50:00.1434299Z  Solution 3: break intel-compilers-devel-ohpc-2023.1-300.ohpc.6.1.x86_64 by ignoring some of its dependencies
```

> Ah, I think the problem is that scalasca requires scorep. You also need to split out scalasca.

Scalasca _should_ work fine with existing versions on OpenHPC, but I'll try. Maybe it gets confused somehow.

adrianreber:
It seems to be that the old version depends on the older version of cubelib which has a different SO name and therefore it cannot resolve the dependencies correctly.

adrianreber:
A rebase should fix the CI error.

adrianreber:
Thanks for your patience with out CI system. Merging now.

PR #2108, opened 2025-04-14
Add GOTCHA v1.0.8

Thyre:
Separated from #2103.

Adds [GOTCHA](https://github.com/LLNL/GOTCHA), a function wrapping library.
This is a dependency for Score-P v9.0, which is unfortunately not available in SLES / RHEL (only Fedora).

Intel test suite will likely fail due to:
```
/home/ohpc/rpmbuild/BUILD/GOTCHA-1.0.8/gotcha-build/test/dispatcher/test_dispatcher: error while loading shared libraries: libsvml.so: cannot open shared object file: No such file or directory
```

This looks like the CI triggers the following issue: https://github.com/intel/llvm/issues/12545

### DIFF
--- a/components/serial-libs/gotcha/LICENSE
+++ b/components/serial-libs/gotcha/LICENSE
@@ -1,0 +1,367 @@
+OUR NOTICE AND TERMS AND CONDITIONS OF THE GNU GENERAL PUBLIC LICENSE
+
+Our Preamble Notice
+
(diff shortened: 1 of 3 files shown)


adrianreber:
Can't you add the requirement to the RPM if using the Intel compiler? Is it just missing an additional RPM?

PR #2109, opened 2025-04-16
admin: upgrade lmod to 8.7.59

adrianreber:
Fix leap apparamor fix.


adrianreber:
Merging as the error is coming from the OBS package and this should fix it.

PR #2110, opened 2025-04-16
Upgrade to Warewulf 4.6.1

middelkoopt:
Migrate changes from upstream SPEC file along with upgrading to version 4.6.1

### DIFF
--- a/components/provisioning/warewulf/SPECS/warewulf.spec
+++ b/components/provisioning/warewulf/SPECS/warewulf.spec
@@ -10,29 +10,35 @@
 
 ## Inserted by OHPC
 %include %{_sourcedir}/OHPC_macros
+%global pname warewulf
+## End OHPC
 
 %global debug_package %{nil}
 
-%global api 0
+%if 0%{?suse_version}
+%global tftpdir /srv/tftpboot
+%else
+# Assume Fedora-based OS if not SUSE-based
+%global tftpdir /var/lib/tftpboot
+%endif
+%global srvdir %{_sharedstatedir}
 
-# Base package name
-%global pname warewulf
-
-# Group for warewulfd and other WW operations
 %global wwgroup warewulf
 
 %if 0%{?fedora}
 %define _build_id_links none
 %endif
 
+## Inserted by OHPC
 # Service directories (change /var/lib/* default to match WW3 build)
 %global tftpdir /srv/tftpboot
 %global srvdir /srv
 %global statedir /srv
+## End OHPC
 
 Name:    %{pname}%{PROJ_DELIM}
 Summary: A provisioning system for large clusters of bare metal and/or virtual systems
-Version: 4.6.0
+Version: 4.6.1
 Release: 1%{?dist}
 License: BSD-3-Clause
 Group:   %{PROJ_NAME}/provisioning
@@ -49,6 +55,7 @@ Conflicts: warewulf-provision
 Conflicts: warewulf-ipmi
 
 %if 0%{?suse_version} || 0%{?sle_version}
+#BuildRequires: distribution-release ## OHPC Removed
 BuildRequires: systemd-rpm-macros
 BuildRequires: go > 1.20
 BuildRequires: firewall-macros
@@ -61,6 +68,7 @@ Requires: firewalld
 Requires: ipxe-bootimgs
 %else
 # Assume Red Hat/Fedora build
+#BuildRequires: system-release ## OHPC Removed
 BuildRequires: systemd
 BuildRequires: golang > 1.20
 BuildRequires: firewalld-filesystem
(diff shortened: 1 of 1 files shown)


components/provisioning/warewulf/SPECS/warewulf.spec
```diff
@@ -49,6 +55,7 @@ Conflicts: warewulf-provision
 Conflicts: warewulf-ipmi
 
 %if 0%{?suse_version} || 0%{?sle_version}
+BuildRequires: distribution-release
```
adrianreber: This breaks the build in our build system. This seems unnecessary anyway even if upstream has this line. Please remove it.

components/provisioning/warewulf/SPECS/warewulf.spec
```diff
@@ -61,6 +68,7 @@ Requires: firewalld
 Requires: ipxe-bootimgs
 %else
 # Assume Red Hat/Fedora build
+BuildRequires: system-release
```
adrianreber: Also remove this line.

middelkoopt:
Changed made and pushed.  Passed light testing.

components/provisioning/warewulf/SPECS/warewulf.spec
```diff
@@ -10,29 +10,35 @@
...
 
-# Group for warewulfd and other WW operations
 %global wwgroup warewulf
 
 %if 0%{?fedora}
 %define _build_id_links none
 %endif
 
+## Inserted by OHPC
 # Service directories (change /var/lib/* default to match WW3 build)
```
martin-g, line 33:
```suggestion
# Service directories (change /var/lib/* default to match WW4 build)
```

components/provisioning/warewulf/SPECS/warewulf.spec
```diff
@@ -10,29 +10,35 @@
...
 %global tftpdir /srv/tftpboot
```
martin-g, line 34: This variable is defined earlier depending on the distro. Maybe it should be removed here ?

components/provisioning/warewulf/SPECS/warewulf.spec
```diff
@@ -244,16 +229,52 @@ getent group %{wwgroup} >/dev/null || groupadd -r %{wwgroup}
...
+%if 0%{?suse_version}
+%else
+Requires: dracut-network
+%endif
+Requires: curl
+Requires: cpio
+Requires: dmidecode
+
+%description dracut
+Warewulf is a stateless and diskless provisioning system for large clusters of
```
martin-g, line 252: Is this description needed here ?
I think only the `dracut` related text should be here.

components/provisioning/warewulf/SPECS/warewulf.spec
```diff
@@ -10,29 +10,35 @@
...
 %global wwgroup warewulf
 
 %if 0%{?fedora}
 %define _build_id_links none
 %endif
 
+## Inserted by OHPC
 # Service directories (change /var/lib/* default to match WW3 build)
 %global tftpdir /srv/tftpboot
 %global srvdir /srv
```
martin-g, line 35: Same comment as for `tftpdir`

components/provisioning/warewulf/SPECS/warewulf.spec
middelkoopt: This was in the original .SPEC file in the PR for Warewulf4, I assumed it is there to match how how Warewulf3 is/was built.  I believe it comes from the following

https://github.com/openhpc/ohpc/blob/9890aca56c2e530f58bf243daa280ec073f475d3/components/provisioning/warewulf-common/SPECS/warewulf-common.spec#L15

defined in `warewulf-common.spec`  and other locations. 

Warewulf4 defaults to `/var` in most cases.
https://github.com/warewulf/warewulf/blob/f2b202964c5def7c819c8569e4f47265a531df05/Variables.mk#L38-L45
```makefile
ifeq ($(USE_LSB_PATHS),true)
  SRVDIR ?= /srv
  LOCALSTATEDIR ?= /var/local
else
  SRVDIR ?= $(PREFIX)/srv
  LOCALSTATEDIR ?= $(PREFIX)/var
endif
CACHEDIR ?= $(LOCALSTATEDIR)/cache
```

My guess was that `/srv` is the standard for OpenHPC packages.

components/provisioning/warewulf/SPECS/warewulf.spec
middelkoopt: The earlier code comes directly from the original Warewulf4 spec file.  To make things easier to port between versions I put the OpenHPC code changes between comment blocks.  Since it seems that OpenHPC uses `/srv` across the board I just redefined it here as per earlier versions of the OpenHPC spec file,  leaving the previous code/definitions from the original Warewulf4 .spec file in place (easier to update with fewer modifications).

The bigger question is, should we be using `/srv` regardless or go back to the distro defaults, and in the case of Rocky/Alma, this would move it from `/srv` to `/var`.  I believe this would cause issues for upgrades.  Perhaps now is the time since adoption is probably low?

components/provisioning/warewulf/SPECS/warewulf.spec
middelkoopt: This is pulled directly from the original SPEC file.  Do you think we should remove it in the OpenHPC version?

components/provisioning/warewulf/SPECS/warewulf.spec
adrianreber: I would not change it now anymore. We can think about it for 4.x.

components/provisioning/warewulf/SPECS/warewulf.spec
adrianreber: I see value in staying close to the upstream SPEC file. Makes changes in the future easier.

middelkoopt:
Rebased against 3.x

components/provisioning/warewulf/SPECS/warewulf.spec
adrianreber: I think @martin-g was talking about the comment. But WW3 is actually correct here as the spec file is using the same location as the WW3 package.

adrianreber:
@martin-g Thanks for taking a look at it. At this point I would say this is good to be merged. The duplicate entries are mainly to keep in sync with the upstream spec file. Which is acceptable from my point of view for easier spec file management. Anything still missing from your side @martin-g?

martin-g (approved):

PR #2105, opened 2025-04-11
Align MANPATH handling in lmod.spec to preserve system man pages (#2070)

Amrithasuresh:
### Summary

This PR updates the `lmod.spec` file to align MANPATH handling

### Changes

- Initialize MANPATH if unset to system defaults
- Use Lmod's 'addto' helper script to append Lmod man pages
- Applied changes in both bash (`lmod.sh`) and csh (`lmod.csh`) profile scripts
- Prevents module MANPATH overrides that hide system man pages (fixes #2070)

### Related Issue

Fixes: #2070

### Signed-off-by

Suresh Panneerselvam <sureshcbt@gmail.com>

Issue #2070, opened 2025-02-03
manpath in modules


pavalenta:
This very well could be my system configuration setup, following the install recipe though no adjustments are made to manpath so this might have cropped with others. Using Rocky Linux 9 for head node and compute nodes, /etc/man_db.conf is setup during system install and not adjusted or changed. Also manpath as a variable is not set systemwide or at the user level. 

When loading modules that have "prepend-path MANPATH" in them, mand_db.conf will not override that variable so users no longer have access to man pages except for the modules loaded that have the prepend manpath variable. A minor change that appears to work and not interfere with subsequent module dependencies is appending a colon the manpath prepend_path in the module file. 

`prepend_path("MANPATH", ":/opt/ohpc/pub/compiler/gcc/14.2.0/share/man")`

adrianreber:
Thanks for the report. Is there a specific package where you have seen this problem happening?

pavalenta:
Apologies, should have included that. Right now I see it in the following gnu modules.

- gnu14-compilers-ohpc-14.2.0-320.ohpc.2.1.x86_64
- gnu13-compilers-ohpc-13.2.0-310.ohpc.4.2.x86_64
- gnu12-compilers-ohpc-12.2.0-300.ohpc.3.6.x86_64

Looks like any module that has "prepend-path    MANPATH" this affects. hwloc-ohpc-2.11.1-320.ohpc.1.1.x86_64 is another package. autotools, pmix, valgrind....

PR #2140, opened 2025-05-08
Fix setting MANPATH in csh

### LINKED ISSUE #2139: New lmod.csh is broken
The new lmod.csh from lmod-ohpc-8.7.59-330.ohpc.7.1.x86_64 is broken and results in:
```
MANPATH: Undefined variable.
```
I'll file a PR to fix shortly.

opoplawski:
This fixes #2139 

Escaping everything needed is tricky, so switched to using `'EOF'`.   In the current version the backticks in the MANPATH command were not escaped and so evaluated to `` when building the RPM.

I don't think the `.tar.gz#$` is really needed, but I can re-add if the change breaks your workflow.  It's presence broke mine.

I still think it's not great to unconditionally append `:` to MANPATH - what if it was already present?

### DIFF
--- a/components/admin/lmod/SPECS/lmod.spec
+++ b/components/admin/lmod/SPECS/lmod.spec
@@ -21,7 +21,7 @@ Release:   %{?dist}.1
 License:   MIT
 Group:     %{PROJ_NAME}/admin
 Url:       https://github.com/TACC/Lmod
-Source0:   https://github.com/TACC/Lmod/archive/%{version}.tar.gz#$/%{pname}-%{version}.tar.gz
+Source0:   https://github.com/TACC/Lmod/archive/%{version}/%{pname}-%{version}.tar.gz
 
 BuildRequires: rsync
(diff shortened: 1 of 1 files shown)


adrianreber:
> I don't think the .tar.gz#$ is really needed, but I can re-add if the change breaks your workflow. It's presence broke mine.

Let's see if it works in OBS. Not sure why it exists.

> I still think it's not great to unconditionally append : to MANPATH - what if it was already present?

See #2070 and #2105. Initially it was checking if the colon exists at the end, but that wasn't working correctly for all use cases. Not sure, maybe the check was not working correctly. So I didn't want to completely undo the change because having the colon at the end made sense. If you have a better solution, please provide a patch. Maybe with your changes concerning escaping `$` it would now work better.

adrianreber:
The build system is happy with your changes: https://obs.openhpc.community/package/show/OpenHPC3:3.3.1:Factory/lmod

PR #2142, opened 2025-05-30
Fix OpenHPC version and other documentation fixes

### LINKED ISSUE #2144: Almalinux9/WW4/SLURM PDF template 3.3, section 3.8.2 error
In section 3.8.2 of the installation, there is a space missing before /bin/bash that causes the following command to fail:

# wwctl image exec --build=false almalinux-9/bin/bash <<- EOF
> dnf -y install ohpc-slurm-client
systemctl enable munge
systemctl enable slurmd
> dnf -y install chrony
> dnf -y install lmod-ohpc
EOF
ERROR: requires at least 2 arg(s), only received 1

-Achilles

middelkoopt:
* Bump documentation version displayed to ~3.3~ 3.4 from 3.2 
* Add missing space in rendered PDF

Fixes: #2144

### DIFF
--- a/docs/recipes/install/common/ohpc-doc.sty
+++ b/docs/recipes/install/common/ohpc-doc.sty
@@ -45,7 +45,7 @@
 % ----------- Variables to change per release -------------
 
 
-\newcommand{\OHPCVersion}{3.2}
+\newcommand{\OHPCVersion}{3.4}
 \newcommand{\OHPCVerTree}{3}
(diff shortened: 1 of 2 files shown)


adrianreber:
Do you expect a rebuild for 3.3. We usually don't do that. If not you could increase it to 3.4.

middelkoopt:
Updated to 3.4

PR #2147, opened 2025-06-12
Upgrade OpenHPC container example with…

middelkoopt:
* Contributed to OpenHPC from https://github.com/MiddelkoopT/ohpc-container
* Updated image and container names to prepend openhpc to reduce collisions.
* Remove old containers based on OpenHPC 2.x


adrianreber:
Do you have any experience automatically publishing containers in the GitHub registry. Probably as part of GitHub actions?

middelkoopt:
> Do you have any experience automatically publishing containers in the GitHub registry. Probably as part of GitHub actions?

Not really, but I've seen it been done in the GitHub actions files I've touched.  And it's not really designed for using remote containers and requires a lot of local scripts.  Part of the build path is to set the container username to match the host.   One option would be to publish a generic container that folks could just execute (I think the openhpc/openhpc3) and an Apptainer one as well.  I did not want to make any big changes because this "just works" now and wanted to get it out.

The instructions for someone to use it would be
```bash
git clone https://github.com/openhpc/ohpc.git
cd ohpc/containers
./run.sh
```

martin-g:
> Do you have any experience automatically publishing containers in the GitHub registry. Probably as part of GitHub actions?

Here is an example 
https://github.com/su2code/Docker-Builds/blob/4212d0470ba4fc203bd4bc99b801b9440ceb99e8/.github/workflows/docker-image-upload.yml#L62-L81

middelkoopt:
Thanks, I was able to adapt ones that I have been using and spent a bit more time learning.  It seems to be working now.  Feel free to review.

middelkoopt:
I've since reorganized the code to handle multiple architectures correctly. I believe it is ready to be merged.

PR #2152, opened 2025-06-25
Different cosmetic changes for containers subdirectories

adrianreber:
* This renames the ReadMe.md file to README.md as this is what we already use in the main directory.
* Apply our linting rules to shell scripts and C code
* Run README.md through the markdown checker
* cleanup temporary dnf files in the containers (decreases size by at least 120MB)
* switch to gnu14

@middelkoopt I made a couple of changes to the containers subdirectory. Please take a look if this all still makes sense to you? Let me know if you prefer some things to be changed back again.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -46,6 +46,7 @@ jobs:
         globs: |
           README.md
           CONTRIBUTING.md
+          containers/README.md
 
   lint:
     strategy:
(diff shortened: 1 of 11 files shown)


middelkoopt (approved):
Changes look good.
Briefly tested on M3/Colima/Docker and M3/podman-machine/podman

PR #2173, opened 2025-07-16
feat(warewulf): add templates and patches for warewulf

adrianreber:
This commit adds templates for chrony, networkmanager, and slurmd to the examples directory. It also adds a patch for the hosts.ww template to include the localdomain.

The templates are used to configure the nodes in the warewulf cluster. The patch is used to update the /etc/hosts file on the nodes.

All changes are trying to reduce the complexity in the recipe by creating files as needed by OpenHPC

@middelkoopt How do you feel about this approach? I tried to simplify a couple of steps in the recipes by creating the necessary files in examples package. This is based on what I saw for the InfiniBand configuration. I think the main advantage is reduced complexity in the commands provided in the recipe.

### DIFF
--- a/components/admin/examples/SOURCES/NetworkManager-wait-online.service.d.override.conf
+++ b/components/admin/examples/SOURCES/NetworkManager-wait-online.service.d.override.conf
@@ -1,0 +1,3 @@
+[Service]
+ExecStart=
+ExecStart=/usr/bin/nm-online -q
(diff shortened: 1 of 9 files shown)


middelkoopt:
I agree with this approach and the fact you used a patch for the hosts file.  Makes sense for OpenHPC.  For my own projects I've moved to `yq` and it makes it a lot more clearer what is being done (I'm a fan of executable documentation) and think we should also package up yq and use it until it hits EPEL.  It will probably be useful in other contexts and a good tool for sysadmins to use.

adrianreber:
> I agree with this approach and the fact you used a patch for the hosts file.

In RPM spec files I like to use patches and not sed/jq/yq, because those tools do not necessarily fail if it doesn't apply any more. Building a package I think it should fail if one of the steps doesn't work.

I will merge this now.

PR #2190, opened 2025-07-29
Fix warewulf documentation typo's (4.x)

middelkoopt:
Thanks to @anderbubble for the catch.


adrianreber:
So CI was not running here. Strange. Need to take a better look next time.

PR #2191, opened 2025-07-29
Package yq for manuipulation of yaml files in recipe files

middelkoopt:
Many recipes would be cleaner with `yq`, which is not en EPEL, this packages `yq` for use in OpenHPC

* Yq can be found at https://github.com/mikefarah/yq
* This packages v4.47.1
* The package name is yq-ohpc to be different from any (future?) yq packages.


adrianreber:
I think we can close this in favour of https://bodhi.fedoraproject.org/updates/FEDORA-EPEL-2025-46cf2c5319

middelkoopt:
Closing, will revisit if yq does not hit epel for some reason.

PR #2192, opened 2025-07-29
Package yq for manuipulation of yaml files in recipe files

middelkoopt:
Same as #2191 but rebased for 4.x branch.


adrianreber:
Thanks. I used claude to analyse this PR and it complains about the following things:

>   Code Quality Issues
> 
>   1. Typo in commit message: "manuipulation" should be "manipulation" in the commit message
>   2. Missing BuildRequires: The spec file only lists golang but may need more specific Go version requirements
>   3. No version constraints: No minimum Go version specified despite using Go modules
> 

Which seems to be correct. There is also problem with out buildsystem which does not have internet access. So we need a tar archive that has a complete vendor tree included.

There actually has been some movement on the EPEL 9 and 10 packages. The Fedora version of yq has gotten a PR to include the vendor tree for easier EPEL 9 and 10 inclusion: https://src.fedoraproject.org/rpms/yq/pull-requests

middelkoopt:
Updated spelling and added minimum go version (1.24.1).  Not sure how to take care of the tar archive of the complete vendor tree.

pushed on 3.x branch as well.

Good to hear about progress with yq in EPEL.

adrianreber:
Take a look at the Fedora pull request I mentioned what they do.

middelkoopt:
Fedora code requires go-vendor-tools, which is not in epel.  Trying something similar.  Triggering build.

middelkoopt:
Did not work as I expected.  Currently stuck on finding an straightforward approach.

middelkoopt:
Closing in favor of EPEL.  Will re-open if the EPEL version for some reason does not happen.

PR #2196, opened 2025-08-03
io-libs: update ADIOS2 to version 2.10.2 and use external yaml-cpp

adrianreber:
Switch to external yaml-cpp library due to internal yaml-cpp build failure with "error: use of undeclared identifier 'uint16_t'" caused by missing cstdint include.

🤖 Generated with [Claude Code](https://claude.ai/code)


adrianreber:
Also see https://github.com/ornladios/ADIOS2/issues/4534

PR #2212, opened 2025-08-27
Add dnsmasq support to Warewulf

middelkoopt:
Based on upstream changes pending for Warewulf 4.6.4.  https://github.com/warewulf/warewulf/pull/1974

* Make dnsmasq default and fix overlay bug on aarch64
* Fix rpm SPEC file dependencies. 

Will remove overlay patch once 4.6.4 goes live.


adrianreber:
The spec file still requires go >= 1.20. Warewulf, however, requires 1.22 from what I see in the buildlogs.

middelkoopt:
The warewulf spec is 1.16, I don't think this can be true.  I'll post an issue against Warewulf and fix it hopefully for v4.6.4

PR #2214, opened 2025-09-02
Fix the build of `gnu-compiler` package on openEuler 24.03 LTS SP2

zhihangdeng:
GCC version 12 is now the default.
Removed conditional checks and manual configuration for `gcc-toolset-12-gcc-c++` and `gcc-toolset-12-libstdc++-devel` in `BuildRequires`.

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -185,7 +185,6 @@ jobs:
         path: tests/**/*.log.xml
 
   build_on_openEuler:
-    if: ${{ false }}  # disable for now
     strategy:
       matrix:
         os: [ubuntu-latest, ubuntu-24.04-arm]
@@ -203,7 +202,7 @@ jobs:
       run: dnf -y install git
     - uses: actions/checkout@v4
     - name: Setup
-      run: tests/ci/prepare-ci-environment.sh
+      run: tests/ci/prepare-ci-environment.sh --pre-release ${{ matrix.compiler }}
     - id: files
       uses: Ana06/get-changed-files@v2.3.0
     - name: Validate Build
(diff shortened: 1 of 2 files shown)


adrianreber:
The commit message doesn't make much sense.

Can you also enable the GitHub Actions openeueler build step, please?

martin-g (approved):

.github/workflows/validate.yml
```diff
@@ -185,7 +185,7 @@ jobs:
         path: tests/**/*.log.xml
 
   build_on_openEuler:
-    if: ${{ false }}  # disable for now
+    # if: ${{ false }}
```
martin-g: Better remove these lines instead of commenting them out

wjunLu: I wonder why `prepare-ci-environment.sh` needs to install all `ophc-*.rpm`?

Since it doesn't build all RPMs before prepare stage.

martin-g: It needs to install just one .rpm file - ohpc-release.rpm
https://github.com/openhpc/ohpc/blob/4.x/tests/ci/prepare-ci-environment.sh#L143

It needs it to add the OpenHPC repo in /etc/yum.repos.d/...
This way whenever you run CI for a specific component it will reuse its already built dependencies.

wjunLu (approved):

.github/workflows/validate.yml
wjunLu: oh! I think I got it.

zhihangdeng:
> The commit message doesn't make much sense.
> 
> Can you also enable the GitHub Actions openeueler build step, please?

Sure.

adrianreber:
Don't enable the the test part of GitHub Actions. Only the build part. The test will not work. Yet.

wjunLu:
@adrianreber It looks good now

wjunLu:
@adrianreber Could you please merge this fix?

adrianreber:
Please squash the last two commits together and its good to go. No need to enable it first and then have another commit disabling it. Leave the changes to the spec file in a separate commit as it is. So in total two commits.

zhihangdeng:
> Please squash the last two commits together and its good to go. No need to enable it first and then have another commit disabling it. Leave the changes to the spec file in a separate commit as it is. So in total two commits.

OK!

PR #2216, opened 2025-09-02
Add EL10/Warewulf/Slurm recipe

middelkoopt:
WIP for review.  

Notes:
* Only tested EL10/Warewulf 4/Slurm/aarch64
* FIXME indicates fixes to get working in current 4.x tree 

Changes of note:
* Move to dnsmasq for DHCP and TFTP services
* Use yq for editing yaml files
* Use /etc/systemd/system-preset for enabling systemd services
* Enable (optional) debugging messages

Items for consideration:
 * Defaulting to dnsmasq in Warewulf in for 4.x for EL8 and EL9 in.
 * Consider moving tftpboot to `/srv/tftpboot` from `/var/lib/tftpboot`
   * Create `/srv/tftpboot` in `warewulf.spec`? (currently make symlink) 
   * Selinux attributes on `/var/lib/tftpboot`, keep in recipe or add in `warewulf.spec` or other location.
 * Add creation of `/etc/systemd/system-preset` to ohpc-base-compute?
   * Needed now to enable services?  Is this the proper method?
 * `$dhcp_start` and `$dhcp_end` are now needed for dnsmasq to set DHCP range
 * Is it still necessary to bring up the internal interface?
   * `ip link set dev ${sms_eth_internal} up`
 * Is `disable --now firewalld` still needed?


adrianreber:
>     * Consider moving tftpboot to `/srv/tftpboot` from `/var/lib/tftpboot`

No real opinion on that. `/srv` always feels a bit out of place, I would do what warewulf upstream does.
 
>     * Add creation of `/etc/systemd/system-preset` to ohpc-base-compute?

If we install it via RPM it should go to `/usr/lib/systemd/system-preset`.

>       ** Needed now to enable services?  Is this the proper method?

Definitely something to try out.

>     * `$dhcp_start` and `$dhcp_end` are now needed for dnsmasq to set DHCP range

Is there no static only allocation of MAC-IP possible?

>     * Is it still necessary to bring up the internal interface?    
>       ** `ip link set dev ${sms_eth_internal} up`

I am also always questioning that line :wink:. It only seems to create an error message.

>     * Is `disable --now firewalld` still needed?

On the compute nodes or on the SMS? I think makes it our life easier as we do not know how the user installed the SMS. It is also problematic to advice to turn the firewall off. I am not sure what the best approach is.

docs/recipes/install/common/install_provisioning_warewulf4_intro.tex
```diff
@@ -13,9 +13,20 @@
 % ohpc_comment_header Add baseline OpenHPC\ref{sec:add_provisioning}
 \begin{lstlisting}[language=bash,keywords={}]
 # Install base packages
-[sms](*\#*) (*\install*) ohpc-base
+## FIXME: Not yet ported (remove %)
+% [sms](*\#*) (*\install*) ohpc-base
```
adrianreber: This works for me. I can install ohpc-base. What error do you get?

middelkoopt:
> No real opinion on that. `/srv` always feels a bit out of place, I would do what warewulf upstream does.

Current proposal (new) is to place in `/var/lib/warewulf/tftpboot` for Warewulf where `/var/lib` is the "shared state" directory, for us will be `/srv` (`/srv/warewulf/tftpboot`)... We hope to take care of creating the directory and setting selinux attributes. Once this is finalized I will normalize our `.spec` file.

docs/recipes/install/common/install_provisioning_warewulf4_intro.tex
middelkoopt:
```
++ dnf config-manager --add-repo http://obs.openhpc.community:82/OpenHPC4:/4.0:/Factory/EL_10/
Adding repo from: http://obs.openhpc.community:82/OpenHPC4:/4.0:/Factory/EL_10/
++ rpm --import http://obs.openhpc.community:82/OpenHPC4:/4.0:/Factory/EL_10/repodata/repomd.xml.key
+ dnf repolist
+ grep -q OpenHPC
+ '[' 0 -ne 0 ']'
+ dnf -y install ohpc-base
created by dnf config-manager from http://obs.o 210 kB/s | 154 kB     00:00    
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
Error: 
 Problem: cannot install the best candidate for the job
  - nothing provides screen needed by ohpc-base-3.3-400.ohpc.9.1.aarch64 from obs.openhpc.community_82_OpenHPC4_4.0_Factory_EL_10_
  - nothing provides libunwind needed by ohpc-base-3.3-400.ohpc.9.1.aarch64 from obs.openhpc.community_82_OpenHPC4_4.0_Factory_EL_10_
  ```

Comment (adrianreber) [reply docs/recipes/install/common/install_provisioning_warewulf4_intro.tex]: Both packages are coming from EPEL. Probably need to enable that.

**Comment by middelkoopt** [comment]
> On the compute nodes or on the SMS? I think makes it our life easier as we do not know how the user installed the SMS. It is also problematic to advice to turn the firewall off. I am not sure what the best approach is.

SMS. For now I just took it out of what is included in recipe.sh - perhaps that is a middle ground.  Give them the advice but don't do it for them in `recipe.sh`

Jut as a background, I now run all my scripts with `-e` so I catch errors, there were a few that I missed before - this is one if the cases that I just visually ignored before.

**Comment by adrianreber** [comment]
> Jut as a background, I now run all my scripts with `-e` so I catch errors, there were a few that I missed before - this is one if the cases that I just visually ignored before.

Using `-e` is also something I was always hoping for. At least in CI.

**Comment by adrianreber** [comment]
With the recent changes, which look really good, there seems to be an ordering issue.

The block `Setting Boot Parameters` needs to happen later. After the final image has been created. Because now all the s3 commands return empty. No image has been pushed to s3.

**Comment by middelkoopt** [comment]
> Using `-e` is also something I was always hoping for. At least in CI.

It's now (yet to be pushed) also in the `wwctl image exec` lines.

File: docs/recipes/install/common/install_provisioning_warewulf4_intro.tex (hunk 1 above)
Comment (middelkoopt) [reply docs/recipes/install/common/install_provisioning_warewulf4_intro.tex]: Fixed for both the SMS and nodes, thanks.

**Comment by middelkoopt** [comment]
Working on decencies again - on the node we have the following issue
```bash
+ dnf -y install ohpc-gnu15-perf-tools
Last metadata expiration check: 0:03:21 ago on Fri 05 Sep 2025 06:31:49 PM UTC.
Error: 
 Problem: package ohpc-gnu15-perf-tools-3.3-400.ohpc.9.1.aarch64 from obs.openhpc.community_82_OpenHPC4_4.0_Factory_EL_10_ requires tau-gnu15-openmpi5-ohpc, but none of the providers can be installed
  - cannot install the best candidate for the job
  - nothing provides libjawt.so(SUNWprivate_1.1)(64bit) needed by tau-gnu15-openmpi5-ohpc-2.34.1-400.ohpc.4.1.aarch64 from obs.openhpc.community_82_OpenHPC4_4.0_Factory_EL_10_
(try to add '--skip-broken' to skip uninstallable packages or '--nobest' to use not only best candidate packages)
```

middelkoopt:
> With the recent changes, which look really good, there seems to be an ordering issue.
> 
> The block `Setting Boot Parameters` needs to happen later. After the final image has been created. Because now all the s3 commands return empty. No image has been pushed to s3.

I assume this is the wrong PR.

adrianreber:
Right. Sorry.

middelkoopt:
> > ```
> > * Add creation of `/etc/systemd/system-preset` to ohpc-base-compute?
> > ```
> 
> 
> If we install it via RPM it should go to `/usr/lib/systemd/system-preset`.
> 
> 
> >

Did more research and presets seems to be the way to go. The node images get a FirstBoot=yes triggered by removing `machine-id` and the new preset (I assume new) kicks in.  I'm still new to this preset thing.

I was thinking just creating the `/etc/systemd/system-preset` directory in `/etc`, so we don't have to explicitly do it in the image.  Since the preset files are installed in the recipe I believe it should go in /etc, not /usr/lib/.....  But I'm thinking we should just have our users create it.

The other way would be to install the presets in the ohpc-slurm-client in `/usr/lib/systemd/system-preset` (or in the ohpc-base-compute). 

So for now I left it in the recipe.  It may also help to be explicit since this might catch some folks up with other services to enable.

middelkoopt:
Ok just pushed another version.  This one is looking a lot better.  The only remaining buggy package I believe is `ohpc-gnu15-perf-tools` (details in another comment)

* I removed the `ip addr` - I can put it back if it breaks CI
* The `yq` direct URL remains until `yq` goes live in EPEL
* The `/var/lib/tftboot` will remain until Warewulf adds this to upstream.
* The node image directly enables the Factory until we go prod.

I have yet to address the dhcp range thing.  It's currently static but has a range for other devices.  I'm checking if this can be removed.  I wanted to get these small changes out first.

I think we are all good except for installing the OpenHPC 4.x repo when it goes live and removing the Factory repo from the node image.

middelkoopt:
oh, I think all the comments are addressed except the dhcp range.  Let me know if I missed something.

middelkoopt:
Looks like we can get a static dhcpd by setting the first entry to the network address and using the "static" mode.  I'll update the patch to support this.  We may want to think about getting it as an upstream options just like dhcp-server.

middelkoopt:
PR updated.  Ready for further review and testing.

docs/recipes/install/common/warewulf4_setup_centos.tex
```diff
@@ -1,20 +1,27 @@
...
+[sms](*\#*) yq -i '.netmask = "'${internal_netmask}'"' /etc/warewulf/warewulf.conf
+[sms](*\#*) yq -i '.network = "'${internal_network}'"' /etc/warewulf/warewulf.conf
+[sms](*\#*) yq -i '.dhcp["range start"] = "'${internal_network}'"' /etc/warewulf/warewulf.conf
+[sms](*\#*) yq -i '.dhcp["range end"] = "static"' /etc/warewulf/warewulf.conf
 
-# Edit the nodes.conf to mount /opt
+# Edit the nodes.conf to mount /opt on boot
 [sms](*\#*) perl -pi -e "s/defaults,noauto,nofail,ro/defaults,nofail,ro/" /etc/warewulf/nodes.conf
 
+# Configure dnsmasq to listen on the internal network iterface.
```
adrianreber: There is a typo here and the CI complains. `iterface`.

adrianreber:
Please remove all non `.tex` files from the manifest directories for now.

middelkoopt:
* Rebased against 4.x and remove WIP tag
* Corrected spelling.
* Removed all non `.tex` files

adrianreber:
Is it still WIP? If not you could also remove WIP from the title.

adrianreber:
> * Removed all non `.tex` files

I still see the files in the `x86_64` tree.

middelkoopt:
* Removed WIP in PR
* Removed non *.tex files in x86_64  manifest
* Restored Makefile and common in parent directory of manifest (aarch64)

middelkoopt:
In case you missed it - no longer WIP and ready for review/merge.  If you think this is in good shape I could also do the AlmaLinux one as well, I was going to wait until this got merged.  Also to do is the provision to disk, but I think that would be better done as another PR but could also do that.

adrianreber:
Works for x86_64 now. Thanks.

From my side ready to merge. Anything from your side missing, or ready to be merged?

middelkoopt:
Nothing from me missing or ready to be merged.  I think we are ready to merge now.  

Here are the things that I think are left to do for Warewulf before the final 4.x:

* Port AlmaLinux 9 and 10 (easy now)
* Add back in OpenHPC 4.x repo in node image when it goes live.
* Bump Warewulf version to latest.
* Merge any Warewulf upstream fixes/cleanups when they happen - all are non-blocking
  * Set  `/etc/system/system-prefix` in `wwclient` overlay
  * Set selinux attributes for `tftboot`
  * Set `interface=${sms_eth_internal}` in upstream overlay for dnsmasq (low priority) 
  * Possibly set `port=0` in upstream overlay for dnsmasq (low priority)
* Write "stateless provision to disk" docs

middelkoopt:
Rebased

PR #2240, opened 2025-09-30
docs: use almalinux and not rocky as compute node image

adrianreber:
@middelkoopt PTAL


middelkoopt:
Looks good to me.

Woops, not sure how that slipped though.  Probably happened when I had to revert some changes that I had made based on some incorrect assumptions (long story).  I believe I tested Alma images before that reversion on x86_64.

PR #2246, opened 2025-10-08
Add enable_dracut for Warewulf one-stage testing

middelkoopt:
Add `enable_dracut` to Warewulf two-stage (Dracut) to completely disable two-stage and use single stage provisioning.  Templates default to two-stage provision to tmpfs.

### DIFF
--- a/docs/recipes/install/common/finalize_warewulf4_provisioning.tex
+++ b/docs/recipes/install/common/finalize_warewulf4_provisioning.tex
@@ -21,6 +21,7 @@ Section~\ref{sec:optional_kargs} to aid in debugging.
 
 
 % begin_ohpc_run
+% ohpc_command if [[ ${enable_dracut} -eq 1 ]];then
 % ohpc_comment_header Two-Stage provision to ramfs
 \begin{lstlisting}[language=bash,literate={-}{-}1,keywords={},upquote=true,literate={BOSVER}{\baseos{}}1]
 ## Install Dracut in the image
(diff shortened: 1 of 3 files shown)


docs/recipes/install/common/finalize_warewulf4_provisioning.tex
```diff
@@ -36,14 +37,15 @@ \subsubsection{Two-Stage Provisioning (optional)}
...
+% ohpc_command fi
 % end_ohpc_run
 
 Optionally, configure the compute node to provision to disk.  Two stage
 provisioning (previous step) must also be enabled.  The compute node will use,
 \textbf{and erase}, the disk specified by \texttt{node\_disk}.
 
 % begin_ohpc_run
-% ohpc_command if [[ ${enable_todisk} -eq 1 ]];then
+% ohpc_command if [[ ${enable_dracut} -eq 1 && ${enable_todisk} -eq 1 ]];then
```
adrianreber, line 48: This is probably the first time we have an `if` with two conditions in the `recipe.sh`. Just something to watch out. We do not run ShellCheck on the recipe (probably a good idea at some point), but I know that ShellCheck would complain about this: https://www.shellcheck.net/wiki/SC2107

The recommendation is to change this to `[[ ${enable_dracut} -eq 1 ]] && [[ ${enable_todisk} -eq 1 ]]`

middelkoopt: The check and solution is for `[` (single square bracket), not the `[[` used int the code, the following code passes in the test
```bash
#!/bin/bash

enable_dracut=1
enable_todisk=1

if [[ ${enable_dracut} -eq 1 && ${enable_todisk} -eq 1 ]];then
  echo pass
fi
```
Cool tool!

PR #2304, opened 2025-11-24
Add proof-of-concept markdown-based documentation system

adrianreber:
This commit introduces a new documentation system that replaces LaTeX with Jinja2-templated Markdown as the primary source format. The system can generate PDF (via Pandoc/XeLaTeX), HTML, and recipe.sh installation scripts from a single markdown source.

Key components:
- Jinja2 templates (.md.j2) with YAML configuration for variable substitution
- Modular common templates in poc-markdown/common/ for reusable content
- parse_doc_md.py: Python parser to extract recipe scripts from HTML comments
- Makefile-based build system with dependencies tracking
- Custom Pandoc Lua filter for format-specific rendering
- LaTeX header templates for PDF customization
- Example implementation for Rocky 10/x86_64/Warewulf4/Slurm

Benefits over LaTeX:
- Simpler markdown syntax easier to read and maintain
- Single source generates multiple output formats (PDF, HTML, recipe.sh)
- Better separation of content and presentation
- Jinja2 variables reduce duplication across OS/arch variants
- HTML comments preserve recipe extraction while keeping content visible

Documentation structure:
- poc-markdown/common/: Reusable template components and utilities
- poc-markdown/rocky10/x86_64/warewulf4/slurm/: Example configuration
- poc-markdown/common/figures/: Shared PDF graphics
- poc-markdown/*.md: Documentation about the system itself

The ohpc_indent directive system has been removed in favor of simpler hardcoded indentation for do-loops only, with all other code at column 0.


middelkoopt:
Having issues with mdtoc - the python one does not support the args used but the npm one is named `markdown-toc`.  It seems to work (it does not quit/crash) when calling `makrdown-toc` directly

adrianreber:
Using the Go one. Will provide an RPM.

middelkoopt:
Going through the code to understand it and think big picture.  About to leave some comments on some of the things I saw, mostly confirming my understanding and possible improvements - realizing that this is a POC with generated code.

docs/recipes/install/poc-markdown/common/parse_doc_md.py
```diff
@@ -0,0 +1,314 @@
...
+        # Regular SMS prompt
+        elif match := self._PATTERNS["sms_prompt"].search(line):
+            cmd = self._substitute_variables(match.group(1))
+            output.write(f"{cmd}\n")
+        # Plain comments
+        elif line.startswith("#"):
+            comment = self._substitute_variables(line)
+            output.write(f"{comment}\n")
+
+    def _substitute_variables(self, cmd: str) -> str:
```
middelkoopt: Looks like this is unused and that the variable substitution is done by Jinja2?  Is this understanding correct?

docs/recipes/install/poc-markdown/common/parse_doc_md.py
```diff
@@ -0,0 +1,314 @@
...
+                    output.write(f"   {line}\n")
+                else:
+                    output.write(f"{line}\n")
+
+                if self._PATTERNS["done_line"].search(line):
+                    break
+        except StopIteration:
+            pass
+
+    def _process_refs(self, comment: str) -> str:
```
middelkoopt, line 250: Does not look like this is not used as well - I assume it is meant to flatten markdown URL's in comments.

docs/recipes/install/poc-markdown/common/parse_doc_md.py
```diff
@@ -0,0 +1,314 @@
...
+
+                if line.endswith("\\"):
+                    output.write(f"\n\t{line}")
+                else:
+                    output.write(f"\n\t{line}\n")
+                    break
+        except StopIteration:
+            output.write("\n")
+
+    def _process_do_loop(self, match: re.Match, output: StringIO) -> None:
```
middelkoopt, line 220: I wonder if re-processing loops adds unneeded complexity, I had a lot of issues in the LaTeX version.  Perhaps the code should be formatted properly in the original markdown?

adrianreber: There shouldn't be a need for loops anymore. This is just what Claude converted from the original script. The chapter numbers are not in the resulting recipe and I removed the indent support also. We will just indent with a default value.

adrianreber: There shouldn't be a need for loops anymore. This is just what Claude converted from the original script. The chapter numbers are not in the resulting recipe and I removed the indent support also. We will just indent with a default value.

adrianreber: I meant no need for reprocessing. We should just have the source material in an acceptable form.

docs/recipes/install/poc-markdown/README.md
```diff
@@ -0,0 +1,123 @@
...
+- `\newcommand` macros for variables (OS name, package manager commands, etc.)
+- `\input{}` directives to include common content
+- `\iftoggleverb{}` for conditional compilation
+- Special comment markers for script generation
+
+This POC replaces that with:
+- Markdown for document formatting
+- YAML configuration files for variables
+- Jinja2 templating for includes and conditionals
+- Jinja2 comments (`{# #}`) for script generation markers
```
middelkoopt: Is this still true?  It seems that you now use HTML comments `<!-- -->`, which I can understand why since the Jinja2 comments are consumed when producing the final `.md` file.

Perhaps another method would be to create a manifest of files (list) to "cat" together replacing the `steps.md.j2` file - I can see a few downsides to this method also.  This process could add the file markers.

middelkoopt: After further investigation html comments seem like the way to go. Claud recommends using the `@` as a delimiter for directives as in `<!-- @ohpc_command -->`

middelkoopt:
Overall it looks very cool and I believe this is the right direction.   think we should invest in this more and take the opportunity to rename the include files to something a bit more standard so it is easier to find things, and other general cleanups. 

One thing I would like to remove is the sms prefix `[sms]#` before commands because some renders (GitHub, mkdocs, and even pandoc with some javascript) can include a copy button and this gets copied (or even a simple triple click for select entire line), it is very helpful to simply click to copy. I have seen some sites that show the prompt but don't copy it, but don't recall where to see what they where using.  The prefix may also cause issues for syntax color hilighting (vsc md preview for example)

As for using pandoc, we may want to look bit harder for the specific renderer we use (mkdocs and others).  There also seems to be some interesting tools for extracting and executing code from markdown.

adrianreber:
> Overall it looks very cool and I believe this is the right direction. think we should invest in this more and take the opportunity to rename the include files to something a bit more standard so it is easier to find things, and other general cleanups.

Yes. I started to replace all calls to `perl` with `sed`. Changing multiple things at the same time makes it harder to test.

You seem to have a stronger opinion related to the file naming. Happy to follow your suggestions. Should we just merge something in a parallel documentation directory and then we can continue to work both on it? It might pollute the git history a bit, but would maybe be easier. Or we could also use a new branch and do the initial work there.

> One thing I would like to remove is the sms prefix `[sms]#` before commands because some renders (GitHub, mkdocs, and even pandoc with some javascript) can include a copy button and this gets copied (or even a simple triple click for select entire line), it is very helpful to simply click to copy. I have seen some sites that show the prompt but don't copy it, but don't recall where to see what they where using. The prefix may also cause issues for syntax color hilighting (vsc md preview for example)

I was also unsure about the `[sms]#` prefix. In the end all commands are running on the SMS. It is not really necessary to have different prefixes for different machines.

> As for using pandoc, we may want to look bit harder for the specific renderer we use (mkdocs and others). There also seems to be some interesting tools for extracting and executing code from markdown.

Sure. As we do not need to post-process the code blocks now as Jinja2 makes all the substitutions early. Definitely could be simplified.

middelkoopt:
> > Overall it looks very cool and I believe this is the right direction. think we should invest in this more and take the opportunity to rename the include files to something a bit more standard so it is easier to find things, and other general cleanups.
>
> Yes. I started to replace all calls to `perl` with `sed`. Changing multiple things at the same time makes it harder to test.

+1 on `sed`

Agreed.  I think your incremental approach to doing one thing at a time is the way to go, keeps the git changes easy to see/audit.  Right now it seems like mostly a "bulk" copy of the docs (I typically do that) then updates from there.

> You seem to have a stronger opinion related to the file naming. Happy to follow your suggestions. Should we just merge something in a parallel documentation directory and then we can continue to work both on it? It might pollute the git history a bit, but would maybe be easier. Or we could also use a new branch and do the initial work there.

Perhaps a new branch, we could squash commits if we wanted it to make it cleaner we it is merged into 4.x.  Although I think one person should do the renaming work.  We would also probably need to hand things off between work/commits because the probability of a conflict would be high (coordinate over slack) unless we coordinate.

The only strong feeling I have is that the current naming is very bad, so something a bit more deliberate would be good, other than that I don't have strong feelings.  Not sure how much near-term time I have to do it given the holidays here.

> > One thing I would like to remove is the sms prefix `[sms]#` before commands because some renders (GitHub, mkdocs, and even pandoc with some javascript) can include a copy button and this gets copied (or even a simple triple click for select entire line), it is very helpful to simply click to copy. I have seen some sites that show the prompt but don't copy it, but don't recall where to see what they where using. The prefix may also cause issues for syntax color hilighting (vsc md preview for example)
> 
> I was also unsure about the `[sms]#` prefix. In the end all commands are running on the SMS. It is not really necessary to have different prefixes for different machines.

Agreed.

> > As for using pandoc, we may want to look bit harder for the specific renderer we use (mkdocs and others). There also seems to be some interesting tools for extracting and executing code from markdown.
> 
> Sure. As we do not need to post-process the code blocks now as Jinja2 makes all the substitutions early. Definitely could be simplified.

I spent a bit more time on this, and I think python is the way to go, it would allow some subtle customizations (say our @ohpc directives) that would get bad real quick if we used a standard tools.  The code for finding it is simple and we just pull out all the extra/unused stuff from the current code.

adrianreber:
@middelkoopt I created a branch for this work at https://github.com/openhpc/ohpc/tree/2025-12-01-markdown

If you have any ideas or improvements please use that branch as basis.

PR #2306, opened 2025-11-28
Make custom documentclass and factor out formatting

mikerenfro:
In preparation for converting docs from pure LaTeX to Markdown, I've made a custom documentclass for install documentation, moving all of the page formatting commands out of steps.tex and other content files.

The new class is based from [memoir](https://ctan.org/pkg/memoir), which should make it easier to improve formatting in the future. For now, memoir is emulating the original article class.

I've built and skimmed over all existing 4.x versions of the documentation. Though the resulting steps.pdf files aren't a line-by-line replica of the originals, they're pretty close, and arguably improved in terms of avoiding widow/orphan lines and breaks in listings.


adrianreber:
Thanks for taking a look at this. I created a branch with my markdown changes here: https://github.com/openhpc/ohpc/tree/2025-12-01-markdown

Please use that as basis to open pull requests against it for any improvements.

middelkoopt:
The document structure has been refactored, reorganized, and ported to Markdown+jenga2+pandoc.  I also noted the inconsistency in heading levels and addressed it in the new documentation.  I may have missed some.

Thanks for this work, closing this PR in favor of the new documentation.

PR #2309, opened 2025-12-18
v4.x: Bump Score-P to v9.4

Thyre:


adrianreber:
Thanks. Our update checker also detected it: https://github.com/openhpc/ohpc/issues/2266

I had to run the update checker manually because it only runs once a week and you were much faster. Thanks!

PR #2381, opened 2026-03-03
docs: add missing '"'

adrianreber:
@middelkoopt


middelkoopt:
Woops, I don't source this file for my testing but set them explicitly so I missed this one.

PR #2416, opened 2026-03-19
Complete Confluent aarch64 support

middelkoopt:
Complete aarch64 port of Confluent.

Tested on Rocky 10.1 and AlmaLinux 10.1 on qemu running on MacOS (M4).

### DIFF
--- a/docs/install/templates/provisioner/confluent/confluent-install.md.j2
+++ b/docs/install/templates/provisioner/confluent/confluent-install.md.j2
@@ -21,7 +21,7 @@ provisioning service on the *head node*:
 # Install Confluent and required packages
 {{ pkg_install }} lenovo-confluent
 {{ pkg_install }} tftp-server nfs-utils policycoreutils-python-utils yq jq
-{{ pkg_install }} confluent_ipxe-aarch64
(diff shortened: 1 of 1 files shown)


docs/install/templates/provisioner/confluent/confluent-install.md.j2
```diff
@@ -21,7 +21,7 @@ provisioning service on the *head node*:
 # Install Confluent and required packages
 {{ pkg_install }} lenovo-confluent
 {{ pkg_install }} tftp-server nfs-utils policycoreutils-python-utils yq jq
-{{ pkg_install }} confluent_ipxe-aarch64
+{{ pkg_install }} confluent_ipxe-aarch64 confluent_osdeploy-aarch64
```
adrianreber, line 24: This works on x86_64 and aarch64? No ifs or something like that?

middelkoopt: > This works on x86_64 and aarch64? No ifs or something like that?

Works on both. This way way it should support provisioning both x86_64 and and aarch64 on both platforms (untested - yet another combo to test).

Just double checked x86_64 on rocky and almalinux.

PR #2420, opened 2026-03-22
docs: escape "$" to not resolve variable too early

adrianreber:
@middelkoopt Does that make sense? Without a change like this variable is not correctly resolved. Do you have a preferred way how to do this?

### DIFF
--- a/docs/install/templates/provisioner/warewulf/compute-install-ohpc.md.j2
+++ b/docs/install/templates/provisioner/warewulf/compute-install-ohpc.md.j2
@@ -24,7 +24,7 @@ wwctl image exec --build=false {{ baseos }} -- /bin/bash -ex <<- EOF
   ohpc_repo="{{ openeuler_ohpc_repo_server }}"
   ohpc_repo+="/{{ openeuler_ohpc_repo_dir }}"
   ohpc_repo+="/{{ openeuler_ohpc_repo_file }}"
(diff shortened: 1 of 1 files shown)


middelkoopt:
Tricky case, how about something like the following:

```
dnf -y config-manager --add-repo {{ openeuler_ohpc_repo_server }}\ 
/{{ openeuler_ohpc_repo_dir }}\
{{ openeuler_ohpc_repo_file }}
```
Note no indentation is possible, bonus it will copy/paste from recipe into a `wwctl image shell` and run... 

You could also do something like

```
  ohpc_repo="{{ openeuler_ohpc_repo_server }}"
  ohpc_repo+="/{{ openeuler_ohpc_repo_dir }}"
  ohpc_repo+="/{{ openeuler_ohpc_repo_file }}"
{{ compute_run("dnf -y config-manager --add-repo ${ohpc_repo}") }}
```

Both untested - still working on my test rig/other issues.  Will try to test a bit later.

I'm thinking I prefer the first.

adrianreber:
@middelkoopt Can you open a PR for the first approach?

middelkoopt:
Sure - I'll need to work through the other issues first in order to test.

middelkoopt:
Replaced by https://github.com/openhpc/ohpc/pull/2433

PR #2433, opened 2026-03-25
warewulf: openEuler support fixes

middelkoopt:
- openEuler Warewulf recipe work — Fixed the openEuler repo URL construction (dropped intermediate shell variable, used line-continuation syntax instead) in both the head node and compute image install paths. Added ipxe-bootimgs RPM dep for openEuler in the warewulf spec.
- Dracut/ignition decoupling — Separated ignition and gdisk from the base dracut install; they're now only installed and configured when enable_todisk=1. This avoids pulling in disk tools on nodes that don't use provision-to-disk.
- Conditional SELinux setup — Wrapped the semanage/restorecon block in warewulf-install.md.j2 with a runtime getenforce check so it's a no-op on systems with SELinux disabled (relevant for openEuler, which may not have SELinux enforcing).
- Recipe script section echo — Added runtime echo '=== section: ... ===' output to mkdoc.py so the generated recipe.sh prints progress as each section begins executing to aid in finding the template.

Replaces https://github.com/openhpc/ohpc/pull/2420

OpenEuler 24.03-SP2 is EOL so that will need to be updated.  

Two stage to ram works.  Provision to disk is missing ignition.  Hard to fully test, mirrors are flakey and US mirror does not fully mirror SP2.

### DIFF
--- a/components/provisioning/warewulf/SPECS/warewulf.spec
+++ b/components/provisioning/warewulf/SPECS/warewulf.spec
@@ -90,7 +90,7 @@ BuildRequires: systemd
 BuildRequires: golang >= 1.22
 BuildRequires: firewalld-filesystem
 Requires: nfs-utils
-%if 0%{?rhel} < 8
+%if 0%{?rhel} < 8 || 0%{?openEuler}
 Requires: ipxe-bootimgs
 %else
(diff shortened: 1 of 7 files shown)


adrianreber:
Thanks. Works for me.

PR #2488, opened 2026-04-25
fix invalid slurm_node_config values in 3.x

mikerenfro:
Might as well apply these fixes for 4.x back into 3.x and 2.x


adrianreber:
Thanks.

PR #2513, opened 2026-04-30
Fixes for OpenCHAMI on 3.x

middelkoopt:
Minor updates to get OpenCHAMI to work on the 3.x branch.  Some of this will need to be back-ported to 4.x (I'm accumulating these).
Tested on almalinux/x86_64 and rocky/aarch64

### DIFF
--- a/docs/install/templates/provisioner/openchami/compute-image-build.md.j2
+++ b/docs/install/templates/provisioner/openchami/compute-image-build.md.j2
@@ -160,7 +160,7 @@ repos:
     url: '{{ distro_base_url }}{{ distro_version }}/CRB/{{ arch }}/os/'
     gpg: '{{ distro_base_url }}RPM-GPG-KEY-{{ distro_name }}-{{ distro_version }}'
   - alias: 'Epel{{ distro_version }}'
-    url: 'https://dl.fedoraproject.org/pub/epel/{{ distro_version }}z/Everything/{{ arch }}/'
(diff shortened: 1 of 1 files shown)


docs/install/templates/provisioner/openchami/compute-image-build.md.j2
```diff
@@ -160,7 +160,7 @@ repos:
     url: '{{ distro_base_url }}{{ distro_version }}/CRB/{{ arch }}/os/'
     gpg: '{{ distro_base_url }}RPM-GPG-KEY-{{ distro_name }}-{{ distro_version }}'
   - alias: 'Epel{{ distro_version }}'
-    url: 'https://dl.fedoraproject.org/pub/epel/{{ distro_version }}z/Everything/{{ arch }}/'
+    url: 'https://dl.fedoraproject.org/pub/epel/{{ distro_version }}/Everything/{{ arch }}/'
```
adrianreber, line 163: Are you sure about this one? The `z` is necessary to point to the latest release of RHEL. Without `z` it points to CentOS Stream 10.

middelkoopt: I am not, it was a quick fix for a 404.  Does the fact that this is EL9 matter?  Do you know what it should be?

adrianreber: You are right, that is a epel 10 only thing. Sorry.

PR #2557, opened 2026-05-07
fix(openeuler): correct Warewulf3 recipes and update SP4/COPR repos

middelkoopt:
- Rename openEuler 22.03 recipes from warewulf -> warewulf3 provisioner
- Update VNFS chroot template to SP4 repository URLs (was SP3)
- Update OpenHPC COPR repo target to openeuler-24.03_LTS_SP2
- Add virtio_pci before virtio_net in Warewulf3 bootstrap for VM support
- Abstract MariaDB/MySQL service name and config path per distro
- Abstract extra-repos package per distro (epel-release vs openEuler-release)
- Remove stale BOS_MIRROR comment and openeuler-extra-repos install step
- Add ChangeLog entries for 3.x

Tested on aarch64 via qemu on a M3 mac.

### DIFF
--- a/components/admin/docs/SPECS/docs.spec
+++ b/components/admin/docs/SPECS/docs.spec
@@ -61,8 +61,8 @@ for recipe in \
 	"almalinux9/x86_64/warewulf3/openpbs" \
 	"almalinux9/x86_64/confluent/slurm" \
 	"almalinux9/x86_64/openchami/slurm" \
-	"openeuler22.03/x86_64/warewulf/slurm" \
-	"openeuler22.03/x86_64/warewulf/openpbs" \
+	"openeuler22.03/x86_64/warewulf3/slurm" \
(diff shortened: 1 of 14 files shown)


adrianreber:
Is this intended? "Update OpenHPC COPR repo target to openeuler-24.03_LTS_SP2" to use the 24.03 copr with 22.03 for OpenHPC 3.x?

You do not need to update the ChangeLog file. It is not being used. But if you want we can leave it in.

middelkoopt:
Good catch.  No and yes-ish.  I missed the fact that the refactor had ported Warewulf to openeuler, not Warewulf3, so early in the debugging I had thought yq was required and did not realize/forgot that the Warewulf3 recipes don't use yq.  So no, not required for Warewulf3 but should work if we need it (yq is a static binary).  Updated and ChangeLog entry removed (perhaps we should remove the Changelog?). 

Tested again and ready to merge.

PR #2572, opened 2026-05-14
fix(warewulf3): assemble VNFS after optional customization steps

middelkoopt:
Move finalize-provisioning (wwbootstrap + wwvnfs) from the end of the "Install Warewulf Provisioner" chapter to the start of the "Deploy Cluster" chapter.

Previously, wwvnfs snapshotted the chroot before the "Additional Customization" chapter ran, so optional components installed into the compute image (InfiniBand, OmniPath, GPU drivers, NHC, syslog, etc.) were silently missing from the image served to compute nodes.

With this change, all chroot modifications — including optional ones — are complete before wwvnfs assembles the VNFS capsule.

### DIFF
--- a/docs/install/templates/chapters/deploy-warewulf3.md.j2
+++ b/docs/install/templates/chapters/deploy-warewulf3.md.j2
@@ -1,6 +1,7 @@
 {# Warewulf v3 Deploy Aggregator #}
 # Deploy Cluster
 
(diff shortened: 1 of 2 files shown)


middelkoopt:
Tested on rocky-9/aarch64/warewulf3 on 3.x branch with slurm and openpbs using qemu on a M3 mac

PR #2586, opened 2026-05-19
Update to Warewulf 4.7.0

middelkoopt:
Upgrade to Warewulf 4.7, removing patches that are upstream.

Tested on aarch64 for rocky-9 and almalinux-9 via qemu on a M3 mac. 

Unable to test x86_64, something in the repo/build is broken for me.


adrianreber:
It doesn't build on leap 15.5 because golang is too old. We are not shipping warewulf4 recipes for leap 15.5. So we can ignore it. Will provide a fix.

adrianreber:
The CI run with two VMs was successful now. This is good to be merged. Thanks. Tomorrow more testing on our test cluster.

PR #2599, opened 2026-06-02
improve time sync for non-UTC systems

mikerenfro:
Out of the box, compute node clocks are synced to the SMS, but always with a UTC timezone. This breaks munge in non-UTC settings. We should also take steps to ensure that chrony jumps if time is too far out of sync.

Whether or not we should abandon the local chrony.conf.ww and use [the upstream one](https://github.com/warewulf/warewulf/blob/main/overlays/chrony/rootfs/etc/chrony.conf.ww) can be discussed.

### DIFF
--- a/components/admin/examples/SOURCES/chrony.conf.ww
+++ b/components/admin/examples/SOURCES/chrony.conf.ww
@@ -1,1 +1,2 @@
 server {{.Tags.ntpserver}} iburst
+makestep 1.0 3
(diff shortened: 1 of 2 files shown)


middelkoopt:
Tested end-to-end on aarch64 qemu (AlmaLinux 9 for 3.x, AlmaLinux 10 for 4.x) using locally-built examples-ohpc RPMs from each branch.

Verified on compute node:

localtime overlay: timezone propagated correctly from SMS (America/Chicago → CDT on compute node) ✓
makestep 1.0 3 present in /etc/chrony.conf ✓
chronyc tracking: synchronized to head, sub-millisecond offset ✓

PR #2614, opened 2026-06-11
nhc: cherry-pick 5e18b7f from dev branch to fix #2404

### LINKED ISSUE #2404: nhc-ohpc hangs on high core count nodes
nhc-ohpc is packaging nhc 1.4.3 which doesn't include https://github.com/mej/nhc/pull/172, hanging on nodes with high core counts.

elelaysh:
as stated in issue, bash read loop from /proc/cpuinfo forces the regeneration of the whole file on each line. Given >10k lines with higher 300 CPUs, it becomes too slow and causes timeouts and Slurm node put in down state.

This upstream fix first reads /proc/cpuinfo into a bash variable, before working on it in a read loop.
It also introduces cache files, which I've not tested.

Fixes #2404

### DIFF
--- a/components/admin/nhc/SOURCES/5e18b7f-speedup-read-cpuinfo.patch
+++ b/components/admin/nhc/SOURCES/5e18b7f-speedup-read-cpuinfo.patch
@@ -1,0 +1,779 @@
+From 5e18b7f5f803765a24cb281f8c4b452281482c4c Mon Sep 17 00:00:00 2001
+From: Michael Jennings <mej@eterm.org>
+Date: Fri, 14 May 2021 02:27:53 -0600
+Subject: [PATCH] WIP:  lbnl_hw:  Fixes/speedups for procfs file reads
+
+This branch/changeset extensively refactors `lbnl_hw.nhc` as a
+possible solution/improvement for the `/proc` file I/O problem (e.g.,
+
+ - Split `nhc_hw_gather_data()` into distinct functions per section,
+   allowing more fine-grained control over which `procfs` and/or
+   `sysfs` files actually need to be read and parsed in the first
+   place;
+ - Convert the path+filename for each file being parsed into a
+   configuration variable that users can customize as needed;
+ - Alter the way NHC is pulling data from each file, by using a single
+   `read` invocation, to avoid the
+   [`lseek()`/rebuild problems described](https://github.com/mej/nhc/issues/43#issuecomment-324690311)
+   by @mattmix in #43.
+ - Add support for using a cached copy of the original file (instead
+   of reading directly from `/proc` or `/sys`) to avoid the resulting
+   poor performance; and
+ - Allow for the aforementioned configuration variables to specify a
+   process substitution expression (see `bash(1)` for details) rather
+   than a path+filename.  This, in turn, permits the user to, for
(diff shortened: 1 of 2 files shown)


elelaysh:
I've downloaded nhc sources and manually applied the patch, then tested  that the fix was effective.

nhc is faster, but in fact the function to gather hardware info has been split up and parts are only called on demand.
Contrary to previous version, where cpuinfo gathering was done as part of ethernet or infiniband checks.

I had to add this check to nhc.conf to trigger the cpu part:

> myhost.internal || check_hw_cpuinfo 1 96 96

Is it an acceptable change in behaviour?

adrianreber:
> I've downloaded nhc sources and manually applied the patch, then tested that the fix was effective.
> 
> nhc is faster, but in fact the function to gather hardware info has been split up and parts are only called on demand. Contrary to previous version, where cpuinfo gathering was done as part of ethernet or infiniband checks.
> 
> I had to add this check to nhc.conf to trigger the cpu part:
> 
> > myhost.internal || check_hw_cpuinfo 1 96 96
> 
> Is it an acceptable change in behaviour?

I don't know. Not sure we have much experience with nhc on the OpenHPC side. We ship it, but the people that introduced are not active any more. I see in our recipe a line like:

```
$ pdsh -w c1 "/usr/sbin/nhc-genconf -H '*' -c -" | dshbak -c
```

Which generates the following on our test cluster:
```
#######################################################################
###
### Hardware checks
###
 * || check_hw_cpuinfo 1 16 32
 * || check_hw_physmem 197704192kB 197704192kB 3%
 * || check_hw_swap 0kB 0kB 3%
 * || check_hw_ib 10 mlx5_0:1
 * || check_hw_ib 10 mlx5_1:1
 * || check_hw_eth lo
 * || check_hw_eth eth0
 * || check_hw_eth eth1
```

It seems you have much more experience with nhc, so I leave the question if this is acceptable to you :wink: 

Thanks for the PR. Let us know if you think this is ready to merged. I will merge it and can then provide a link to the build system where to download the package for the upcoming release.

elelaysh:
Thank you @adrianreber for your kind and toughtful reply!

We [only monitor](https://github.com/stackhpc/ansible-slurm-appliance/blob/main/ansible/roles/nhc/templates/nhc.conf.j2) mounted (shared) filesystems and network interfaces by default, to be honest! I guess it depends on what problems tend to creep up in your clusters.

To my mind the PR is now ready to merge.

adrianreber:
Package built for upcoming 4.2 release is here: https://obs.openhpc.community/project/show/OpenHPC4:4.2:Factory

adrianreber:
Ah, you opened it against 3.x. Missed that. Disregard the comment above.

elelaysh:
Thank you @adrianreber , would you also like a PR agains 4.x?

adrianreber:
Yes, please. A cherry pick could work.

adrianreber:
The builds for 3.x are available here: https://obs.openhpc.community/project/show/OpenHPC3:3.6:Factory

elelaysh:
Thank you. I think we will patch the file instead: https://github.com/stackhpc/ansible-slurm-appliance/pull/989

PR #2637, opened 2026-06-29
Add ppc64le (IBM POWER) architecture support

viniciusferrao:
## Summary

ppc64le (IBM POWER) architecture enablement for OpenHPC 4.x / EL10. @adrianreber suggested bringing this upstream. Sponsored by VersatusHPC; validated on POWER9 hardware (AlmaLinux 10 and openEuler 24.03).

Restructured per review: **one commit per spec file**, single-line subjects. The gotcha ccache handling is brought in by cherry-picking the three relevant 2.x commits (`git cherry-pick -x`), with the ppc64le-specific gotcha change applied on top of them — not folded into a single commit.

## Scope

Component specs that needed architecture handling: gotcha, gnu-compilers, boost, fftw, openblas, likwid (+ a perf_event uncore patch), pdtoolkit, numpy, mpich, and the meta-package closure. **11 files.**

The GNU compiler POWER9 baseline already landed in `OHPC_setup_compiler` upstream, so it is no longer part of this PR.

## Design

- **All ppc64le-specific changes are gated by `%ifarch ppc64le`** (or `%ifnarch … ppc64le` in the meta-package), so the existing **x86_64 and aarch64 builds are unaffected**. For gotcha, `rpmspec --parse` differs by exactly one line between targets (the ppc64le `CMAKE_INSTALL_LIBDIR=lib`).
- The baseline is **`-mcpu=power9`**, matching the EL9+ ppc64le distro minimum (`redhat-rpm-config`'s `%__cflags_arch_ppc64le`) and forward-compatible with POWER10/POWER11.
- Per-package highlights: openblas `TARGET=POWER9`, fftw `--enable-vsx`, boost `architecture=power`, likwid `GCCPOWER` + a perf_event uncore patch, pdtoolkit `ibm64linux`, a numpy VSX3 build fix, gotcha `CMAKE_INSTALL_LIBDIR=lib` (cmake otherwise defaults to `lib64`, which breaks the module's `lib` path — matching the existing `scalapack` idiom), and the meta-package groups scoped to the MPI families built on POWER (openmpi5).

## Verification

Built **every touched package from source on both x86_64 and ppc64le** (clean containers; AlmaLinux 10 and openEuler 24.03). All compile, with **no regression on the existing architectures** — the gated hunks are inert off-POWER, confirmed by both `rpmspec --parse` diffs and real RPM builds. gotcha 1.0.10 (the version this PR cherry-picks from 2.x) builds clean on ppc64le with all 18 ctest cases passing.

## Notes

- Independent of the ucx/hwloc spec fixes at the merge level. A from-source build of the full ppc64le stack also wants the ucx `Source0` fix to fetch ucx, but that's a build-time convenience, not a merge dependency.

Sponsored by VersatusHPC.

### DIFF
--- a/components/admin/meta-packages/SPECS/meta-packages.spec
+++ b/components/admin/meta-packages/SPECS/meta-packages.spec
@@ -121,23 +121,25 @@ Collection of compute node base packages
 
 %package -n %{PROJ_NAME}-%{compiler_family}-io-libs
 Summary:   OpenHPC IO libraries for GNU
+%ifnarch ppc64le
 Requires:  adios2-%{compiler_family}-mpich%{PROJ_DELIM}
-Requires:  adios2-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
-Requires:  hdf5-%{compiler_family}%{PROJ_DELIM}
 Requires:  netcdf-cxx-%{compiler_family}%{PROJ_DELIM}
 Requires:  netcdf-cxx-%{compiler_family}-mpich%{PROJ_DELIM}
-Requires:  netcdf-cxx-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 Requires:  netcdf-fortran-%{compiler_family}%{PROJ_DELIM}
 Requires:  netcdf-fortran-%{compiler_family}-mpich%{PROJ_DELIM}
-Requires:  netcdf-fortran-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 Requires:  netcdf-%{compiler_family}%{PROJ_DELIM}
 Requires:  netcdf-%{compiler_family}-mpich%{PROJ_DELIM}
-Requires:  netcdf-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 Requires:  pnetcdf-%{compiler_family}-mpich%{PROJ_DELIM}
-Requires:  pnetcdf-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
 Requires:  phdf5-%{compiler_family}-mpich%{PROJ_DELIM}
+%endif
+Requires:  adios2-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
+Requires:  hdf5-%{compiler_family}%{PROJ_DELIM}
+Requires:  netcdf-cxx-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
+Requires:  netcdf-fortran-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
+Requires:  netcdf-%{compiler_family}-%{mpi_family}%{PROJ_DELIM}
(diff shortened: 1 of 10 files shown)


adrianreber:
For easier re-use on our different branches it would be good to have one commit for each spec file. Please change that, so that each commit changes only one spec file. The ccache changes in gotcha.spec should also not be part of the same commit. There are three commits on the 2.x branch for gotcha which can be brought to 4.x with `git cherry-pick -x`. Then apply your changes on top of it.

viniciusferrao:
> For easier re-use on our different branches it would be good to have one commit for each spec file. Please change that, so that each commit changes only one spec file. The ccache changes in gotcha.spec should also not be part of the same commit. There are three commits on the 2.x branch for gotcha which can be brought to 4.x with `git cherry-pick -x`. Then apply your changes on top of it.

Alright. I'll look at it.

viniciusferrao:
Hi @adrianreber, PR was amended to have the separate commits per file.

The other issues were also addressed.

Thanks.

viniciusferrao:
Those `aarch64` failures were unexpected. I will look at it.

adrianreber:
> Those `aarch64` failures were unexpected. I will look at it.

Don't worry, this is because of the meta packages for the arm HPC compiler. I know how to fix it.

adrianreber:
Please rebase. That should solve the problems.

adrianreber:
I see you make mpich on ppc64le depend on hwloc from OpenHPC. Any reason for that?

viniciusferrao:
> I see you make mpich on ppc64le depend on hwloc from OpenHPC. Any reason for that?

I tough that was the correct thing to do since OpenHPC ships hwloc. Should it depends on the distro hwloc?

adrianreber:
> > I see you make mpich on ppc64le depend on hwloc from OpenHPC. Any reason for that?
> 
> I tough that was the correct thing to do since OpenHPC ships hwloc. Should it depends on the distro hwloc?

Probably correct. But then we should do it for all architectures. Let's leave this PR for now. It is already large enough. But maybe something for a future change.

viniciusferrao:
> > > I see you make mpich on ppc64le depend on hwloc from OpenHPC. Any reason for that?
> 
> > 
> 
> > I tough that was the correct thing to do since OpenHPC ships hwloc. Should it depends on the distro hwloc?
> 
> 
> 
> Probably correct. But then we should do it for all architectures. Let's leave this PR for now. It is already large enough. But maybe something for a future change.

Ok. I can remove that and we do afterwards.

viniciusferrao:
Done @adrianreber 

I'm still running tests here. Recompiling everything takes a while.

adrianreber:
Sorry, but this needs another rebase. There were still things missing from other branches.

viniciusferrao:
> Sorry, but this needs another rebase. There were still things missing from other branches.

Yeah, I was looking at it. There was permission errors in the CI. Let me fix.

adrianreber:
I merged this for now. For further changes please try to do smaller PRs for easier testing of the changes.

PR #2640, opened 2026-07-01
numpy: ignore distro installed pkgs in pip install

viniciusferrao:
Hi @adrianreber this is PR to fix the CI/CD failure in the initial `ppc64le` support PR.

Failed here: https://github.com/openhpc/ohpc/actions/runs/28473398738/job/84424354415?pr=2637

The issue here is that `pip` checks whether numpy is already installed in the build chroot, not in --prefix. If `python3-numpy` already exists, from the distribution, `pip` prints "Requirement already satisfied". 

That simples does not install anything into the buildroot.

So the proposed solution is to add `--ignore-installed`.

Regards,


adrianreber:
Thanks.

PR #2644, opened 2026-07-02
scorep: gate fdupes BuildRequires to SUSE

viniciusferrao:
Same issue as described in the `boost` PR. Gate to SUSE since we don’t use it in OpenEuler and it’s necessary for ppc64le build.


adrianreber:
Currently all your PRs are failing because your commits are missing Signed-off-by

adrianreber:
We should probably drop fdupes everywhere. If we ever need we can handle it differently.

PR #2646, opened 2026-07-04
gnu-compilers: build texinfo inline

viniciusferrao:
Hi @adrianreber, this one is a little bit different. And I think this is the last one in the track of `ppc64le`.

openEuler's does not ship texinfo in `ppc64le` So the idea was to gate the BuildRequires to openEuler `ppc64le` only; it stays on EL, SUSE, and openEuler x86_64/aarch64, etc.

I didn't removed it because we would lose the docs for every system. The info manuals (gcc.info, gfortran.info, etc) are skipped there.

There is a decision here: should OpenHPC drops the info manuals, or ships texinfo for OpenEuler `ppc64le`.

### DIFF
--- a/components/compiler-families/gnu-compilers/SPECS/gnu-compilers.spec
+++ b/components/compiler-families/gnu-compilers/SPECS/gnu-compilers.spec
@@ -34,6 +34,10 @@
 %global gnu15_mpc_version 1.4.1
 %global gnu15_mpfr_version 4.2.2
 
+# openEuler has no dependable texinfo (none at all on its ppc64le port); build
+# this version inline on openEuler so makeinfo is available for GCC's info manuals.
+%global texinfo_ver 7.0.3
+
 %if "%{compiler_family}" == "gnu12"
 %global gnu_major_ver gnu12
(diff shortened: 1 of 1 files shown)


adrianreber:
What about building texinfo before building the compiler like here:

https://github.com/openhpc/ohpc/blob/4.x/components/mpi-families/mvapich2/SPECS/mvapich2.spec#L35

Would that be doable?

viniciusferrao:
Is it a build cache issue that causes aarch64 build to fail? I could not make sense of it.

adrianreber:
> Is it a build cache issue that causes aarch64 build to fail? I could not make sense of it.

No idea what is going trying a rebuild works: https://github.com/openhpc/ohpc/pull/2647

adrianreber:
Maybe try the texinfo change for openeuler on all arches.

adrianreber:
Now it fails because the upstream tar archive locations are not reachable. I will trigger a rebuild soon.

adrianreber:
Now it works. This is a bit strange, but let's merge it.

PR #2647, opened 2026-07-06
gnu-compilers: test build with tmate debug on failure

adrianreber:
Add tmate SSH debug session to the openEuler build job to investigate brp-strip failure on aarch64. Includes a dummy spec change to trigger the gnu-compilers build.


viniciusferrao:
@adrianreber I was looking at this and it feels like a race condition in the parallel `%__os_install_post`. The binaries that fail are all hardlinks, so I think find-debuginfo / brp-strip ends up stripping both names of the same inode at once and corrupts it, which is why it's unreliable and only happens on the aarch64 runner. Have you tried it with `-j1`?

adrianreber:
Closing this as it was just as test.

> Have you tried it with -j1?

No, I haven't.

PR #2652, opened 2026-07-11
docs: style fix

viniciusferrao:
This PR normalizes `systemctl` calls in the install guides to always spell out `.service`, and adds line spacing to some code snippets.

The guides were inconsistent: most call sites used the bare unit name while a handful spelled out `.service`, sometimes both ways within the same file (`nfs.md.j2`, `time.md.j2`). Per the discussion below, this now normalizes **towards** `.service` rather than away from it, which also matches the other unit types the recipes already reference (`.socket`, `.target`).

Non-service units keep their own suffix. `tests/ci/integration-test.sh` matches `systemctl restart rsyslog` with `sed` to append `|| true`, so that pattern is updated to keep matching.

Spacing change keeps the pattern of:

```
# comment
code
<empty line>
# comment
code
```

### DIFF
--- a/docs/install/templates/base-os/disable-firewall.md.j2
+++ b/docs/install/templates/base-os/disable-firewall.md.j2
@@ -11,6 +11,6 @@ follows:
 <!-- ohpc_begin -->
 <!-- ohpc_comment Disable firewall -->
 ```bash
-systemctl disable --now firewalld || true
+systemctl disable --now firewalld.service || true
 ```
 <!-- ohpc_end -->
--- a/docs/install/templates/base-os/nfs.md.j2
+++ b/docs/install/templates/base-os/nfs.md.j2
@@ -21,7 +21,7 @@ echo "/opt/ohpc/pub *(ro,no_subtree_check,fsid=11)" >> /etc/exports
 exportfs -a
 
 # Restart and enable nfs server
-systemctl restart nfs-server
-systemctl enable nfs-server
+systemctl restart nfs-server.service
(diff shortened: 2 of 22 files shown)


middelkoopt:
I've actually gotten in the habit of typing out these in full when doing this on the CLI (auto-completion).  However, are you just normalizing them all to not have .service? (if so, regardless of the style, normalizing them is the way to go.  @adrianreber any thoughts on going with .service over removing?

viniciusferrao:
To be honest it doesn't matter if it have `.service` or not. I usually type without it. But the idea here is consistency, the vast majority have `systemctl` without the `.service` part, so it was to keep everything similar.

adrianreber:
> I've actually gotten in the habit of typing out these in full when doing this on the CLI (auto-completion). However, are you just normalizing them all to not have .service? (if so, regardless of the style, normalizing them is the way to go. @adrianreber any thoughts on going with .service over removing?

Consistency is nice. Not sure what is better. Having `.service` sounds good as there are also things like `.timer`, `.socket`, `.mount`. Which means for things like those you have to use it. So having `.service` everywhere sound also right.

viniciusferrao:
Fixes implemented, `.service` is added not only on `systemctl` but also where it applies when running `journalctl`.

viniciusferrao:
@middelkoopt this is an easy one. It's just the style convetion we agreed upon.

middelkoopt (approved):
Approved - sorry about the delay, I've been busy and this got lost.

PR #2653, opened 2026-07-11
docs/install: include the scheduler chapter dynamically

viniciusferrao:
Use the same dynamic include idiom as the provisioner chapters so new schedulers only need config + templates.

Sponsored-by: VersatusHPC

### DIFF
--- a/docs/install/templates/chapters/main.md.j2
+++ b/docs/install/templates/chapters/main.md.j2
@@ -2,7 +2,7 @@
 {% include "chapters/introduction.md.j2" %}
 {% include "chapters/base-os.md.j2" %}
 {% include "chapters/ohpc.md.j2" %}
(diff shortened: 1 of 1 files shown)


middelkoopt (changes requested):
Please integrate this into a single PR (https://github.com/openhpc/ohpc/pull/2654).

viniciusferrao:
The idea was to keep it easy to be cherrypicked and @adrianreber usually asks. But that's fine. I'll add to the one you mentioned.

Are you sure @middelkoopt because this change isn't only affecting xCAT. It will affect the other provisioners.

middelkoopt:
I will need to test all the provisioners and would rather do it all at once, but please keep it as a single commit.

viniciusferrao:
Superseeded by #2654

PR #2654, opened 2026-07-11
docs/install: add xCAT provisioner with rocky10/almalinux10 recipes

viniciusferrao:
'An corp' xCAT stateless guides for OpenHPC 4.x with EL10 systems.

Sponsored-by: VersatusHPC


middelkoopt:
I most likely will not be able to review to this until after PEARC and PTO (Mid Aug), but will try.  Thoughts on integrating stateless and stateful into one doc?  We do something similar in Warewulf for provision-to-disk.

viniciusferrao:
> I most likely will not be able to review to this until after PEARC and PTO (Mid Aug), but will try. Thoughts on integrating stateless and stateful into one doc? We do something similar in Warewulf for provision-to-disk.

Stateful is not ready yet. I wanted to finish that for the next TSC meeting be not sure yet. Still testing the steps.

But the idea was to file another PR. If we will merge in a single one, I would need to redo it.

If that's the approach let me know that I will start the rework instead of finishing the stateful part.

middelkoopt:
The general idea is to merge the two recipes, not have two separate ones - and that would be best as a single PR.  I don't  quite understand why there would be two.  If they really need to be two recipes (very different workflow) then that's fine - it just seems odd (I've not looked at the previous xcat or used xcat in the past).  Also note that there will be considerable testing to get it to actually get it working on our testing environment so doing both at once would be good.

viniciusferrao:
I need to get the stateful back on track. I procrastinated due to the cancellation of next meetings.

viniciusferrao:
@middelkoopt At first I wasn't though it was a good idea, however I was able to merge both stateless and stateful in a single recipe. It required some replacement of information but it looked fine in my opinion, in chapters 5 and 7.

In fact it's way better now.

middelkoopt:
Good to hear - I hope to get back to looking at this in detail soon.

middelkoopt:
Working on it now.  Would it be possible to publish aarch64 packages at xcat.org?  This is my dev/build platform and may folks have aarch64 systems now (NVIDIA)

viniciusferrao:
Unfortunately aarch64 isn't supported in xCAT, yet.

The xCAT Consortium will deliver it probably in 2.20, before SC26, but we need to get some hands on aarch64 hardware.

I can arrange x86_86 and ppc64le servers of you need @middelkoopt

middelkoopt:
I have access to x86_64 hardware, thanks.

PR #2677, opened 2026-09-03
genders: require perl-devel on EL

viniciusferrao:
The Perl bindings need /usr/share/perl5/ExtUtils/typemap, which perl-devel
provides. The BuildRequires applied only to openEuler.

The guard stays because SUSE has no perl-devel package.


adrianreber:
Please drop the "ci: retrigger" commit

viniciusferrao:
> Please drop the "ci: retrigger" commit

Done, it was a hack to... you guessed: retrigger CI.

PR #2678, opened 2026-09-03
charliecloud: point Source0 at the published release asset

viniciusferrao:
The package_files URL returns 404. Use the release asset URL for 0.44 and
drop the now unused package_file_id macro.

### DIFF
--- a/components/runtimes/charliecloud/SPECS/charliecloud.spec
+++ b/components/runtimes/charliecloud/SPECS/charliecloud.spec
@@ -20,18 +20,17 @@
 %define versionize_script() (sed -i 's,/env python,/env %1,g' %2)
 
 # Unfortunately this changes with every release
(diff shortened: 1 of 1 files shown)


adrianreber:
Could you update it to 0.45.1 and also take a look at this: https://gitlab.com/charliecloud/charliecloud/-/releases#release-tarballs-now-have-usable-urls

I mentioned it to the upstream developers that they do not have stable tar archive URLs and this seems to have changed with newer releases.

viniciusferrao:
> Could you update it to 0.45.1 and also take a look at this: https://gitlab.com/charliecloud/charliecloud/-/releases#release-tarballs-now-have-usable-urls
> 
> I mentioned it to the upstream developers that they do not have stable tar archive URLs and this seems to have changed with newer releases.

It is done.

Issue #2470, opened 2026-04-17
Component submission: environment-modules


xdelaruelle:
---
#### Software Name
Environment Modules

---
#### Public URL
https://envmodules.io/

---
#### Technical Overview

Environment Modules project provides an alternative `module` implementation to Lmod with many new features to smoothly manage user environment and a lot more to come in the future. Modules is now a Linux Foundation project part of [HPSF](https://hpsf.io/)

As of version 5.6, Modules supports:
* Hierarchical module organization
* All Lmod Tcl modulefile commands (copes with the `family`, `depends-on`, `module try-add` commands used in OpenHPC modulefiles.

More information:
* List of major new features with examples: https://modules.readthedocs.io/en/latest/MIGRATING.html
* A detailed comparison with Lmod: https://modules.readthedocs.io/en/latest/other-implementations.html

The aim of this submission is to allow OpenHPC users to use Environment Modules on their OpenHPC installation instead of Lmod if they want to. This is not to replace what is currently set as default but to provide an alternative solution.

This integration will require to have all modulefiles and modulercs of ohpc repository written in Tcl, or at least a Tcl modulefile for each Lua modulefile you would like to keep (Lmod copes very well this that). There are a few exceptions but the ohpc repository already mostly contains Tcl modulefiles.

We will provide all the required support for this integration and on the `module` topic at large, as we always do with projects integrating ours (RHEL, Spack, EasyBuild).

---
#### Latest stable version number
5.6.1

---
#### Open-source license type
GPLv2+

---
#### Relationship to component?
- [x] contributing developer
- [ ] user
- [ ] other

If other, please describe:

---
#### Build system
- [x] autotools-based
- [ ] CMake
- [x] other

If other, please describe:
`configure` script is a bash script, but it acts like an autotools-generated `configure` script.

A RPM spec file is available at https://github.com/cea-hpc/modules/blob/master/contrib/rpm/environment-modules.spec.in

Does the current build system support staged path installations?
For example: ```make install DESTIR=/tmp/foo``` (or equivalent)

- [x] yes
- [ ] no

---
#### Does component run in user space or are administrative credentials required?
- [x] user space
- [ ] admin

---
#### Does component require post-installation configuration.

- [x] yes
- [ ] no

If yes, please describe briefly:
[...]

PR #2680, opened 2026-09-05
Translate Lua modulefiles to Tcl and define default version with `.version` file

xdelaruelle:
Translate the 5 existing Lua modulefiles into Tcl to harmonize all OpenHPC modulefiles in this language.

Change the 6 existing `default` symlink into Tcl `.version` modulerc files to harmonize the way module default version are defined in all OpenHPC packages.

This change is part of the preparatory work for the addition of Environment Modules (#2470)


adrianreber:
Thanks. I think this might be one of the first PR I will be looking at and testing.

adrianreber:
This probably needs to be split in smaller PRs. The GCC rebuild usually takes almost as long as GitHub Actions allows it. So it probably will not finish.

xdelaruelle:
> This probably needs to be split in smaller PRs. The GCC rebuild usually takes almost as long as GitHub Actions allows it. So it probably will not finish.

If I do a specific PR for GCC and another for the other packages, do you think it will be ok?

adrianreber:
> > This probably needs to be split in smaller PRs. The GCC rebuild usually takes almost as long as GitHub Actions allows it. So it probably will not finish.
> 
> If I do a specific PR for GCC and another for the other packages, do you think it will be ok?

Yes, that should work. Only trilinos would be a problem as it takes forever to build.

xdelaruelle:
> > > This probably needs to be split in smaller PRs. The GCC rebuild usually takes almost as long as GitHub Actions allows it. So it probably will not finish.
> > 
> > 
> > If I do a specific PR for GCC and another for the other packages, do you think it will be ok?
> 
> Yes, that should work. Only trilinos would be a problem as it takes forever to build.

It seems it was able to finish. So is it still worth to split in 2 PRs?

openEuler job seems to fail on something else. Should I patch something?

xdelaruelle:
> openEuler job seems to fail on something else. Should I patch something?

I propose #2692 to fix the openEuler build issue with plasma. Rebased this PR on the other

adrianreber:
So, CI is happy, good. I would like to have one more discussion in the TSC before starting to merge your PRs. The next meeting is probably on 2026-09-16. From my side it looks ready.

We already agreed in the TSC that it would be a good addition for OpenHPC but let's have one more round in the TSC next week.

PR #2682, opened 2026-09-07
ci: select ruff rules explicitly

viniciusferrao:
`ruff check` started failing on pull requests that do not touch Python code. The failures comes from a change in Ruff's default rule set, combined with the lint container installing from `fedora:latest`.

To temporaly address the issues, we explicitly select the rules that Ruff previously enabled by default. This keeps lint behavior stable for now until we adapt everything for the new rules.


adrianreber:
I fixed all the ruff complaints. I prefer that instead of disabling checks. Thanks for the PR but I will close it.

PR #2694, opened 2026-09-12
misc: respect prerelease opt-in for GitHub release lists

Hanabi9248:
With a version pin, the update checker skips `/releases/latest` and selects from the release list. That list includes prereleases, so a pin such as `4.2` can select `v4.2.10rc1` over `v4.2.9` even without `--prereleases`. The fallback after a failed `/releases/latest` request has the same problem.

Filter prereleases before selecting from the list unless the option is enabled. Keep the existing release order and version-pin matching.

Validation: `python -m unittest discover -s misc -p test_check_for_package_updates.py -v` passes all 5 tests; 4 fail on the baseline. The tests use fixed GitHub API responses and cover pinned releases, the fallback, opt-in, and no matching stable release. Ruff check, Ruff format check, and `git diff --check` pass. The full RPM/package scan was not run.

### DIFF
--- a/misc/check_for_package_updates.py
+++ b/misc/check_for_package_updates.py
@@ -389,6 +389,12 @@ def get_latest_github_release(
         debug_info(f"No releases found for {repo}", verbose)
         return None
 
+    if not check_prereleases:
+        releases = [rel for rel in releases if not rel.get("prerelease", False)]
+        if not releases:
+            debug_info(f"No stable releases found for {repo}", verbose)
+            return None
+
     if version_pin:
         for rel in releases:
             tag = rel.get("tag_name", "")
--- a/misc/test_check_for_package_updates.py
+++ b/misc/test_check_for_package_updates.py
@@ -1,0 +1,57 @@
+"""Regression tests for GitHub release selection."""
+
+import unittest
+from unittest.mock import Mock, patch
+
+import check_for_package_updates as checker
+
+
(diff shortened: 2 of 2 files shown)


adrianreber:
What problem are you trying to fix? When do you see the problem?

Hanabi9248:
I found this while testing the version-pin path, not from a failed package build. The `v4.2.10rc1` example is a synthetic API fixture; I have not observed an incorrect alert for the currently packaged PMIx 4.2.9. I should have made that distinction clearer.

The trigger is a matching prerelease appearing before the stable release in `/releases`. With `version_pin="4.2"` and `check_prereleases=False`, the current function returns that prerelease without checking its flag. The new `test_pinned_release_skips_prerelease` reproduces this on d9662185; it returns `v4.2.10rc1` where the test expects `v4.2.9`. The fix makes this path honor the documented opt-in, as `/releases/latest` already does for unpinned checks.

adrianreber:
Not sure we need a fix for something that is not a problem. Not sure I understand the reason here. Can you give an example.

Hanabi9248:
I checked the current PMIx release data and cannot point to an affected package in the present OpenHPC configuration. The test demonstrates a conditional mismatch with the option, but I do not have a concrete failing update report to justify this change. I will close this for now. Thanks for asking for that evidence.

PR #2696, opened 2026-09-13
containers: pass ARCH build-arg to all image builds

xdelaruelle:
The `container` image Containerfile declares `ARG ARCH=aarch64`, but `build.sh` only passed `ARCH` to the `openhpc` image. On an x86_64 host the `container` image build fetched the aarch64 `ohpc-release` RPM and failed.

Pass the host architecture (from `uname -m`, with the existing `arm64` to `aarch64` mapping) to every image build. The `head` and `node` images do not declare the argument and ignore it.


adrianreber:
Rebased to pick up the CI fixes.

@middelkoopt Can you take a look at the container changes?

PR #2697, opened 2026-09-13
Fix scalasca build with ccache, MinIO image pull and skip known-broken packages in CI

xdelaruelle:
Three packages fail to build in CI as soon as a pull request touches their spec file, and none of them is present in the OBS 4.2 Factory repositories. This PR fixes the one that is CI-specific and skips the two that need real work. It also repairs the OpenCHAMI integration test, which broke independently.

- **Handle ccache in the scalasca spec.** Scalasca's configure passes CC/CXX to the MPI compiler wrappers through `-cc=`, which cannot parse the `ccache gcc` value used by CI, so the mpich variant fails with "C compiler cannot create executables". The same PATH masquerade workaround already applied to scorep in #2634 is used.
- **Skip opencoarrays and python-scipy in CI.** They are added to `SKIP_CI_SPECS` next to lustre and trilinos. The openmpi5 variant of opencoarrays is refused by upstream CMake as OpenMPI is incompatible with the coarray implementation of gfortran 15. python-scipy 1.5.4 has no `python3-pybind11-devel` build dependency available on EL10 and its Cython sources are incompatible with numpy 2.x, so it needs a version update.
- **Pull the MinIO image from quay.io in the OpenCHAMI recipe.** The `minio/minio` repository is gone from Docker Hub, so the minio quadlet fails to pull its image and the recipe aborts before any node is provisioned. The image is still published on quay.io.

The build failures were exposed by the CI runs of #2679, which rebuilds 32 spec files. The OpenCHAMI failure affects every run of the AlmaLinux 10 openchami Slurm Integration job since 10 September.


adrianreber:
Everything looks correct. Thanks. opencoarrays and scipy is not actively used any more and maybe we should just delete it at some point. Not entirely sure. Let's merge this.

PR #2698, opened 2026-09-14
Make the CI build driver and integration tests more robust

xdelaruelle:
Two CI fixes exposed by the runs of #2679 and of this PR.

- **Tolerate invalid UTF-8 in the build output.** The CI build driver reads the rpmbuild output in text mode with strict UTF-8 decoding. Parallel builds interleave the output of several compiler processes on the same pipe, and a write boundary may fall inside a multibyte character such as the curly quotes of gcc diagnostics. The resulting invalid byte sequence raises `UnicodeDecodeError` and aborts the whole job, as happened on the RHEL aarch64 build of #2679 in the middle of the tau build, which emits thousands of such warnings. Decode with `errors="replace"` so such bytes turn into a replacement character instead of an exception. The failure is intermittent, the same build passed on the previous run of the PR.
- **Compute the build order on openEuler too.** `misc/build_order.py` drops the kernel module packages from the dependency graph, but only matches the `kmod-*` names of the RHEL branch of lustre.spec, not the `*-kmp` names of its other branch. On openEuler the script dies with `KeyError: 'lustre-client-ohpc-kmp'` and the build driver silently falls back to the spec file order, which builds perf-tools before serial-libs. scorep then gets built against the opari2 package published on OBS instead of the one built by the job and records that install path in `scorep-config`. Since the opari2 bump to 2.0.10, the openEuler tests of #2679 install the new opari2 and every scorep instrumented compilation fails on the missing 2.0.9 `opari2-config` path. The `*-kmp` names are now skipped as well.


adrianreber:
Please drop the always rebuild the docs package from the PR. I triggered a manual rebuild in the build system. Seems like the webhook for the notification didn't work.

xdelaruelle:
> Please drop the always rebuild the docs package from the PR. I triggered a manual rebuild in the build system. Seems like the webhook for the notification didn't work.

done

PR #2699, opened 2026-09-15
Fix setting of LMOD_SETTARG_CMD for csh

opoplawski:


adrianreber:
For all your commits we need a Signed-off-by to pass CI.

opoplawski:
Thanks for the info.  Done.

PR #2704, opened 2026-09-18
tests/ci: move openEuler validate image to 24.03 SP3

middelkoopt:
## Move the openEuler validate CI image to 24.03 SP3

openEuler 24.03 SP2 is heading out of support and does not ship the `golang-1.24`
package that current builds need (e.g. to bootstrap a newer Go). SP3 provides it.

Also switch the validate image's mirror to a US mirror
(`mirrors.ocf.berkeley.edu`), since CI runs on US-based runners and the
default/CN mirrors are slow.

Split out from #2703 so the CI-image move can land first.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


middelkoopt:
After this merges, the openhpc org validate image needs a rebuild/republish to pick up SP3 — either the weekly `build-ohpc-validate-container.yml` cron or a manual `workflow_dispatch`. Until that image is republished, openEuler PR CI keeps pulling the current SP2 image.

